PR #1, opened 2025-12-15
Add readme

apurvam:


README.md
```diff
@@ -1 +1,23 @@
...
\ No newline at end of file
+# Why OpenData?
+
+## Our Beliefs 
+
+1. We believe that object storage is a fundamentally new ingredient in data systems: it provides highly durable, highly available, infinite storage with unique performance and cost structures. On the one hand, it solves one of the hardest problems in distributed data systems: consistent replication. On the other hand, care must be taken to make it work corretly, performantly, and cost effectively. It turns out that the techniques to achieve the latter are pretty similar across systems: batching on writes, caching on reads, snapshot isolation. When done right, systems built natively on object storage are far simpler and cheaper to operate in modern clouds than those that haven't used object storage as a starting point.
+2. Inspired by the UNIX philosophy, we believe single purpose systems that compose well are superior to systems that try to solve many problems under one umbrella. Today's business landscape incentivizes the vendors who are custodians of our beloved open source systems to bloat their offerings in order to justify higher prices that can deliver required margins. This hurts the developer base at large.
```
apurvam: I think it's better to probably drop references to vendors here. Instead, we can just say:

1. We believe in the UNIX philosophy of composable single purpose systems that share a common design language, rather than multi-purpose monoliths. 
3. We believe that there is a general lack of objectstore native open source data systems, to the detriment of the developer community at large.

README.md
```diff
@@ -1 +1,23 @@
-# opendata
\ No newline at end of file
+# Why OpenData?
```
agavra, line 125: I feel like the `README.md` should start by answering the question "what is this" and the entire document should be phrased in that way. It feels a little off to me to phrase it as Beliefs/Vision (I don't think I've seen that in any OSS project). 

The points are all valid and I think we can tweak it to be more of a "what" than a "why" and then save some of the "why" for a different document.

For example, we can say that this is a "collection of data systems that leverage object storage and therefore are highly durable, highly available, have infinite storage and provide unique performance and cost structures." When talking about the techniques those can be a list of "features common throughout all of the implementations"

README.md
```diff
@@ -1 +1,81 @@
...
+## Metadata 
+
+TODO, but here's a sketch:
+
+1. The metadata layer leverages SlateDB's [manifest](https://github.com/slatedb/slatedb/blob/main/rfcs/0001-manifest.md). The manifest contains the locations of all the files that comprise the data in the system. 
+2. The general principle is that data is only read if the files are accessible via the manifest. 
+3. SlateDB leverages compare-and-set operations that are available in every object store to atomically updatet the manifest.
+4. Ingestors and Compactors flush data to objectstore, and then atomically update the locations in the manifest.
+5. SlateDB further versions the metadata using epochs, which ensures that zombie writers can't roll versions back.
+6. Readers of data, including Compactors and Query Executors, thus always get a consistent view of the metadata at any point in time.
```
apurvam, line 98: Basic question: If the manifest is large, can it be read atomically from objectstorage? I assume there must be some guarantee at that level.

README.md
```diff
@@ -1 +1,81 @@
-# opendata
\ No newline at end of file
+# Overview
+
+OpenData is a collection of open source data infrastructure projects designed from ground up for objectstoraege. We aim to deliver highly focused, objectstore-native versions of online databases that can power your application stack. As of today, we have a competitive timeseries database and a data streaming backend. We plan to ship vector search, text search, and other database types over time. Contributions are welcome!
```
cadonna:
```suggestion
OpenData is a collection of open source data infrastructure projects designed from ground up for object storage. We aim to deliver highly focused, objectstore-native versions of online databases that can power your application stack. As of today, we have a competitive timeseries database and a data streaming backend. We plan to ship vector search, text search, and other database types over time. Contributions are welcome!
```

README.md
```diff
@@ -1 +1,81 @@
...
+5. SlateDB further versions the metadata using epochs, which ensures that zombie writers can't roll versions back.
+6. Readers of data, including Compactors and Query Executors, thus always get a consistent view of the metadata at any point in time. 
+7. This allows basic snapshot isolation across reads and writes. 
+
+# Our systems.
+
+OpenData ships two systems today:
+
+* TSDB: An objectstore native timeseries database that can serve as a backend for Prometheus. Its a great option for a low cost, easy to operate, grafana backend. Learn more about it here : <link to RFC>
+* EventStore: Think of it as Kafka 2.0. An objecstore native event streaming backend that supports millions of logs, so you can finally get a replayable log per key. Learn more about it here: <link to rfc>
```
cadonna: Where does the name `EventStore` come from?

README.md
cadonna: `objectstorage` or `object storage`?

README.md
```diff
@@ -1 +1,81 @@
...
+To address these structural problems with opensource data infrastructure today, OpenData systems adopt the following principles: 
+
+1. Each OpenData product will target one core use case and aim to be best-in-class at it. 
+2. Each OpenData product will be designed from ground up with object storage as a starting point.
+3. Each OpenData product will be built on a common set of primitives. One set of primitives is [SlateDB](https://www.slatedb.io), but there may be others. 
+4. Each product and every common component will be licensed under the MIT license. 
+5. All OpenData products will be built with a uniform operational experience, which includes common CLI tooling, common versioning schemes, common configuration styles, and more.
+6. Every OpenData product will be designed to compose with the others. 
+
+OpenData aims to put control back in the hands of developers. Your data is always in your custody, and OpenData makes it practical to serve that data to any app without needing multiple specialist engineers on staff.
```
cadonna: Maybe the BYOC aspect should be emphasized more. WDYT?

README.md
```diff
@@ -1 +1,87 @@
-# opendata
\ No newline at end of file
+# Overview
+
+OpenData is a collection of open source databases designed from ground up for object storage. We aim to deliver highly focused, objectstore-native versions of online databases that can power your application stack. As of today, we have a competitive timeseries database and a data streaming backend. We plan to ship vector search, text search, and other database types over time. Contributions are welcome!
+
+Building performant, cost effective, and correct online systems on objectstorage takes special care. Successful designs all have to solve the problem of write batching, multiple levels of caching, and snapshot isolation for correctness. OpenData systems build on a common foundation to solve these problems. This common foundation gives OpenData systems a common set of operational tools, configuration systems, etc., that make our systems easier to operate in aggregate.
```
hachikuji: nit: cost-effective

README.md
apurvam: This is addressed in the latest version.

README.md
```diff
@@ -1 +1,87 @@
-# opendata
\ No newline at end of file
+# Overview
+
+OpenData is a collection of open source databases designed from ground up for object storage. We aim to deliver highly focused, objectstore-native versions of online databases that can power your application stack. As of today, we have a competitive timeseries database and a data streaming backend. We plan to ship vector search, text search, and other database types over time. Contributions are welcome!
```
hachikuji, line 3: nit: we're a little inconsistent on the use of "object storage" vs "objectstorage" in this doc. I slightly prefer treating the words separately, but I'm ok either way.

README.md
```diff
@@ -1 +1,87 @@
...
+Each OpenData product has a high level architecture that looks like this:
+
+![OpenData high level architectural components](public/img/opendata-arch.png)
+
+## Data Flow
+
+The core concepts are best understood by following the typical data flow.
+
+1. The API layer is specific to each database, and implements the appropriate read and write APIs for the database. When the API layer receives write requests, they are forwarded to the Ingestor services.
+2. The Ingestor services are stateless. They batch writes coming from the API layer, flush the data in a Write-Ahead-Log (WAL) format which is optimized for high throughput writes to object storage, and update the Metadata. We describe the metadata layer in detail later, but for now it suffices to know that the matadata contains locations and versions of all files in the system. The metadata lives in objectstorage, and all OpenData systems use the same components to enable writing metadata updates atomically. Atomic updates to metadata ensure that readers and writers in the system operate on consistent views of the data in the system. The Ingestors are fairly similar across database types.
```
hachikuji: I'm trying to wrap my head around this. For the tsdb implemetnation, we have OTEL ingest as an option. We may also implement stateless ingest through object storage as a common ingest framework. I don't think the data system itself specifies anything about the statefulness of ingest. It is possible that we have stateful kafka ingest on top of our log implementation for example. These are the key points in my mind:
1) Ingest is logically decoupled from query (object storage allows both logical and physical separation)
2) Ingest is pluggable (just like query). For a tsdb, a user could rely on OTEL ingest exclusively. Another user might rely on the stateless ingest layer we write. Another might ingest from our log

agavra: nit: let's use "component" rather than service since these can and often are co-located for simple deployments

README.md
```diff
@@ -1 +1,87 @@
...
+
+The flip side of this decoupled architecture is that you have higher end-to-end latency between when data is inserted into the system and when it is returned in a query. This means truly interactive use cases where users must read their writes as soon as possible are not good fits for OpenData databases. However, when some end-to-end latency is acceptable, the flexiblity of the OpenData architecture makes OpenData databases the superior option in a cloud-native world. 
+
+
+# Databases
+
+OpenData ships two databases today:
+
+* TSDB: An objectstore native timeseries database that can serve as a backend for Prometheus. Its a great option for a low cost, easy to operate, grafana backend. Learn more about it here : <link to RFC>
+* EventStore: Think of it as Kafka 2.0. An objecstore native event streaming backend that supports millions of logs, so you can finally get a replayable log per key. Learn more about it here: <link to rfc>
```
agavra: nit: looks like we're calling it `log`

README.md
```diff
@@ -1 +1,87 @@
...
+
+
+# Why OpenData?
+
+1. We believe that object storage is a fundamentally new ingredient in data systems: it provides highly durable, highly available, infinite storage with unique performance and cost structures. It solves one of the hardest problems in distributed data systems: consistent replication. At the same time, tremendous care must be taken to make object storage work correctly, performantly, and cost effectively. When done right, systems built natively on object storage are far simpler and cheaper to operate in modern clouds than the alternatives. We want to bring the benefits of object storage to every database.
+2. Inspired by the UNIX philosophy, we believe single purpose systems that compose well are superior to systems that try to solve many problems under one umbrella, which is what has tended to happen with existing open surce projects since each is developed in a silo. Buy building multiple focused systems on a common core, Open Data will not have to trade off power for simplicity. 
+4. Object store native designs have tended to be proprietary, to the detriment of the developer community at large. Open Data remedies that by being MIT licensed from day one.
```
agavra:
```suggestion
3. Object store native designs have tended to be proprietary, to the detriment of the developer community at large. Open Data remedies that by being MIT licensed from day one. 
```

README.md
```diff
@@ -1 +1,87 @@
...
+2. Inspired by the UNIX philosophy, we believe single purpose systems that compose well are superior to systems that try to solve many problems under one umbrella, which is what has tended to happen with existing open surce projects since each is developed in a silo. Buy building multiple focused systems on a common core, Open Data will not have to trade off power for simplicity. 
+4. Object store native designs have tended to be proprietary, to the detriment of the developer community at large. Open Data remedies that by being MIT licensed from day one. 
+
+
+# License
+
+MIT License
```
agavra: we can drop this, we have a `LICENSE.md`

README.md
agavra: what do you mean by "read atomically"? manifests are immutable so they're never updated

agavra (approved):
This looks really good to me! I'm happy with this as the first cut for the README. 

I do think we should take a pass through and reduce word count since it's a wordy at times. Left some nits inline.

README.md
apurvam: Ah of course, that would be the case on S3. In my head I imagined them more akin to a database, with rows being updated.

README.md
```diff
@@ -1 +1,87 @@
...
+2. The Ingestor services are stateless. They batch writes coming from the API layer, flush the data in a Write-Ahead-Log (WAL) format which is optimized for high throughput writes to object storage, and update the Metadata. We describe the metadata layer in detail later, but for now it suffices to know that the matadata contains locations and versions of all files in the system. The metadata lives in objectstorage, and all OpenData systems use the same components to enable writing metadata updates atomically. Atomic updates to metadata ensure that readers and writers in the system operate on consistent views of the data in the system. The Ingestors are fairly similar across database types.
+3. There are two types of Compactors in the OpenData system. One type is called the WAL compactor. It takes new WAL files written by ingestors and write the WAL entries into a format that's optimized to serve the queries of each database. For example, a compactor for a TSDB will take a WAL file and write inverted indexes, dictionaries, etc. These read-optimized files are called Collections.
```
hachikuji, line 65: This feels a little deep for the README. Perhaps a separate architecture document could get into the details of systems on top of SlateDb. I wonder if this doc could focus on the core abstractions: storage as key-values, readers (ingest), writers (query). My mental model of opendata is structured like this:

1) Root: The root of the system is SlateDb which implements the LSM abstraction and provides the hooks to control its behavior.
2) Core Data Structures: on top of SlateDb, we have the core storage modules. This basically defines the SlateDb record mapping and the core API to consume the data structure. We treat tsdb and log as two core data structures. The API of these data structures is modeled after SlateDb.
3) Ingest: On top of core data structures, we define ingest APIs which are basically providing the server abstraction to take client requests and invoke the write APIs on the underlying data structure. OTEL is an ingest layer for tsdb. For the log, we could build a Kafka adapter which accepts `Produce`requests.
4) Query: On top of core data structures, we also expose higher-level query functionality. While the core of tsdb data access may be primitive, we would define a prometheus query layer which works with PromQL. Or for the log, we could define a Kafka layer which works like a consumer or accepts `Fetch` requests.

README.md
apurvam: I like this breakdown. For the TSDB, does the OTEL ingest basically implements the OTEL protocol and writes WAL files direct to S3? So in some sense the API and Ingestor is the same module?

hachikuji: Each slate exposes a writer which is fenced through the manifest. The writer is the one that writes to the WAL. I think the main thing ingest governs is the path from the user to the writer. For the OTEL implementation, it is an OTEL server which embeds the writer. So an OTEL request gets translated into a write on the tsdb data structure. The writer then writes to the WAL.

In the case of stateless ingest, here is how I was thinking about it (which may not be the final implementation). We don't want to have every writer in the system fighting for the manifest. Instead, any producer can dump raw log files directly into object storage without any fencing protection. In some other process, we could embed the actual slate writer which monitors the directory for raw log files and then writes them into the data structure. We may opt to skip the WAL in this case since the data is already durable in the raw log files. So the path from the user to the writer is through object storage.

apurvam: Ah. that makes sense.

>  In some other process, we could embed the actual slate writer which monitors the directory for raw log files and then writes them into the data structure.

This 'other process' is essentially the WAL compactor in Rohan's terminology.

README.md
apurvam: This makes sense. I added a 10,000ft view of the architecture that is along these lines. I didn't get into the specifics of building custom data structures on top of slateDB, aligning APIs, etc, since that will get too detailed.

README.md
apurvam: I updated the text to reflect this nuance.

hachikuji (approved):
LGTM. Just one nit. I wonder if we can use the ascii diagrams in the README. They are easier to evolve than pngs and easier to view in editors.

PR #2, opened 2025-12-15
Open-log RFC 1

hachikuji:
This patch contains an initial sketch of OpenLog storage and APIs. I've focused on the core structure and APIs and punted more complicated work such as range scanning and push protocols to future work.

### DIFF
--- a/open-log/rfcs/0001-storage.md
+++ b/open-log/rfcs/0001-storage.md
@@ -97,13 +97,14 @@ struct Record {
     value: Bytes,
 }
 
+#[derive(Default)]
 struct WriteOptions {
     await_durable: bool,
 }
 
 impl Log {
-    async fn append(&self, record: Record, options: WriteOptions) -> Result<(), Error>;
-    async fn append_batch(&self, records: Vec<Record>, options: WriteOptions) -> Result<(), Error>;
+    async fn append(&self, records: Vec<Record>) -> Result<(), Error>;
+    async fn append_with_options(&self, records: Vec<Record>, options: WriteOptions) -> Result<(), Error>;
 }
 ```
 
@@ -119,6 +120,7 @@ struct LogEntry {
 }
 
 // TODO: decide which SlateDB ScanOptions parameters to pass through
+#[derive(Default)]
 struct ScanOptions {
 }
 
@@ -129,26 +131,36 @@ impl ScanIterator {
 }
 
 impl Log {
-    fn scan(&self, key: Bytes, seq_range: impl RangeBounds<u64>, options: ScanOptions) -> ScanIterator;
+    fn scan(&self, key: Bytes, seq_range: impl RangeBounds<u64>) -> ScanIterator;
+    fn scan_with_options(&self, key: Bytes, seq_range: impl RangeBounds<u64>, options: ScanOptions) -> ScanIterator;
 }
 ```
 
 ### Count API (under consideration)
 
-Lag is a critical metric for tracking progress reading from a log. Without contiguous sequence numbers, computing lag requires the SST enhancements described in [SST Representation](#sst-representation).
+Lag is a critical metric for tracking progress reading from a log. Without contiguous sequence numbers, computing lag requires the SST enhancements described in [SST Representation](#sst-representation). This proposal adds an explicit API to count the number of records that are present within any range of the log for a key. 
 
 ```rust
 // TODO: decide which SlateDB ScanOptions parameters to pass through
+#[derive(Default)]
 struct CountOptions {
-    approximate: bool,
+    approximate: bool,  // default: false (precise counts)
 }
 
 impl Log {
-    async fn count(&self, key: Bytes, seq_range: impl RangeBounds<u64>, options: CountOptions) -> Result<u64, Error>;
+    async fn count(&self, key: Bytes, seq_range: impl RangeBounds<u64>) -> Result<u64, Error>;
+    async fn count_with_options(&self, key: Bytes, seq_range: impl RangeBounds<u64>, options: CountOptions) -> Result<u64, Error>;
 }
 ```
 
 This mirrors the scan API but returns a count rather than entries. The `approximate` option allows counting from the index alone without reading boundary blocks.
+
+For example, to compute the current lag for a key from a given sequence number:
+
+```rust
+let current_seq: u64 = 1000;
+let lag = log.count(key, current_seq..).await?;
+```
 
 ## Alternatives
 


open-log/rfcs/0001-storage.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Design
+
+### Key Encoding
+
+SlateDB keys are a composite of the user key and a `u64` sequence number. A leading byte discriminator distinguishes entry types.
+
+```
+Log Entry:
+  SlateDB Key:   | 0x01 | key (bytes) | sequence (u64) |
```
agavra: do we want to add a versioning byte like we have for the open-tsdb?

open-log/rfcs/0001-storage.md
```diff
@@ -0,0 +1,154 @@
...
+
+#### Variable-Length Key Considerations
+
+With variable-length keys, the boundary between key and sequence number is ambiguous during comparison. Two approaches to handle this:
+
+**Custom comparator** — SlateDB could support a custom comparator that treats the key and sequence number as separate components. This would enable correct ordering without encoding changes.
+
+**Scan filtering** — Without a custom comparator, scans may return entries from adjacent keys when key bytes happen to interleave with sequence number bytes. Results can be filtered by exact key match.
+
+In practice, users are likely to use fixed-length keys, which avoids this issue entirely.
```
agavra, line 54: I wonder if we should have different key types (0x01/0x02) for fixed and not-fixed key length. The benefit of this is that the encoding for fixed doesn't need to have a `length` encoded with it. (On that note, would be good to have an encoding section)

open-log/rfcs/0001-storage.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Sequence Numbers
+
+Sequence numbers are assigned from a global counter that increments on every append. Each key's log is monotonically ordered by sequence number, but the sequence numbers are not contiguous—other keys' appends are interleaved in the global sequence.
```
agavra: What do you mean by "global"? I assume this is scoped to a "SlateDB" instance? It might be worth a quick note on distributing this workload or the deployment model / extensions for distributing it

open-log/rfcs/0001-storage.md
```diff
@@ -0,0 +1,154 @@
...
+    fn scan(&self, key: Bytes, seq_range: impl RangeBounds<u64>, options: ScanOptions) -> ScanIterator;
+}
+```
+
+### Lag and Count (under consideration)
+
+Without contiguous sequence numbers, computing lag requires additional bookkeeping. The approach under consideration augments SlateDB data structures:
+
+- Each SST tracks the number of records it contains
+- Each SST provides the relative index of a given entry within it
```
agavra: I'm not certain I understood what you meant by this? do we need to track this for every key+seq?

open-log/rfcs/0001-storage.md
```diff
@@ -0,0 +1,154 @@
...
+## Design
+
+### Key Encoding
+
+SlateDB keys are a composite of the user key and a `u64` sequence number. A leading byte discriminator distinguishes entry types.
+
+```
+Log Entry:
+  SlateDB Key:   | 0x01 | key (bytes) | sequence (u64) |
+  SlateDB Value: | record value (bytes) |
```
agavra, line 39: do we want to consider headers from day 0? I feel like this caused a lot of pain in Kafka but I don't have the context as to why. I'm not totally sure I see why headers should be something special.

open-log/rfcs/0001-storage.md
apurvam: I'd assume so. Kafka by analog stores things like log end offset per partition. This is similar in spirit as I understand it. 

However, we probably need to optimize it if we expect the number of keys to be orders of magnitudes more than kafka partitions.

open-log/rfcs/0001-storage.md
hachikuji: Do we need to encode the length for a variable length key? If the prefix and suffix have a fixed size, the key is whatever is left.

agavra: 💡 oh that's an interesting idea, very tricky! I guess we don't.

open-log/rfcs/0001-storage.md
hachikuji: Good question. I would tend to view headers as not part of the core data structure. A messaging abstraction built on top of the log could add constructs for headers built into either the key or the value. What do you think?

open-log/rfcs/0001-storage.md
hachikuji: I added some further notes about how this might be implemented. Let me know if it is still not clear.

open-log/rfcs/0001-storage.md
hachikuji: I clarified that this is a per-writer sequence number.

hachikuji:
> 1. Scan N records from for key `K` with sequence > `S`: what's the execution path. how many index lookups for the first time. Is it the same operations for each of the `N` messages in the scan, or are the subsequent messages in the scan cheaper to retrieve?

Good question. I will add some notes about this, but it may be incomplete. We will need a separate RFC which describes compaction knobs that we will expose. At a high level, the read characteristics will depend on the density of the data. If the data is not dense in L0, then there will be some read amplification because blocks may span multiple keys. The more you compact, the denser you can pack the data at the cost of more writes. In regard to the cost, I had thought that we would probably pass through SlateDb's underlying caching knobs. Since reads are probably mostly at the tip of the log and since recent blocks will be in memory, read amplification would be less of an issue.

> 2. Compute 'lag', or 'number of messages to end of the log' for key `K` at current position `S`.

I added some details about this. I see it as conceptually similar to a range scan, but mostly limited to the indexes. To be precise, we need to read blocks from the range bounds at each relevant level in our search. If you wanted to find the total count of records for a given key, we would compute the number of records in each L0 and each sorted run, and then sum the total. Worst case, that is two blocks read from each L0 and each sorted run. However, I figure that most of the time, lag will be computed at the tip of the log, so indexes will all be in-memory (probably blocks too). The bloom filter might actually help us with this search, but we need the granularity to be at the level of a log key, and not a SlateDb key. Let me add some discussion about this.

> 3. Publish `N` messages across keys A-C: how many writes across data and indexes does that take? How many hops before being readable. What's the lower bound on end-to-end latency given that dataflow?

Data is durable in SlateDb once written to the WAL (or once the memtable has been flushed if the WAL is disabled). I was thinking we would expose Slate's durability options. User would be able to read data before it is durable for example. This is useful when the source of the data is already durable. 

We have discussed a stateless ingest layer, which would add an additional persistence hop. I see this as a logically separate piece which could be used for ingest across any of the opendata systems. Stateless ingest would take control of the writer and probably expose data before it is durable in the log as discussed above.

apurvam:
Thanks for the updates and your detailed explanations. I think I finally got the basic model that you are going for. I'll try to explain in my own words to make sure I got it.

Essentially, the `(key, global sequence)` tuple totally orders the log. Each SlateDB Block will include the start `key, sequence` and the count of records in the block. 

When we want to read N messages from of key `K` at sequence `>S`, we have to find the first block which includes `K` with Sequence `>=S`. This can be done using bloom filters to filter out SSTs without `K`. For the remaining, we have to read the block indexes to find the right starting block. We start from the top of the tree and move down. 

Once we find the right starting block, we scan it to find the message for `K` and `S' > S`. From there, we walk up the tree, skipping blocks and SSTs until we find the next block with `K` and sequence `S'`.

Isn't a precise count operation exactly the same? We find the block containing the starting key, and then find every newer block which contains K and count the number of instances of `K` in each of those blocks until the end. 

An approximate count could get away with just using the 'count' field from each block. Older blocks will have much more accurate counts for a given key since keys will be compacted to be close to each other. So the approximate count would be sufficient. 

--

Do I have it right in my head?

hachikuji:
@apurvam Yes, I think your understanding is correct. The `range` and `count` operations both walk the tree. The only difference is the state that is accessed. For the count, we only need to access blocks if the range bounds do not cleanly align on a block boundary. I thought an approximate count which relies only on indexes might be good enough for lag tracking.

open-log/rfcs/0001-storage.md
agavra: that always made sense to me, which is why I'm not sure I understand why headers were such a big deal in Kafka... maybe it's one of those ecosystem things where people designed schemas without headers, then wanted headers and forced it into the data model so they could avoid schema evolution or something like that...

open-log/rfcs/0001-storage.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Design
+
+### Key Encoding
+
+SlateDB keys are a composite of the user key and a `u64` sequence number. A version prefix and record type discriminator provide forward compatibility.
+
+```
+Log Entry:
+  SlateDB Key:   | version (u8) | type (u8) | key (bytes) | sequence (u64) |
```
agavra, line 38: wdyt about having version|type just be a single u8? I imagine 128 values for each is plenty (though prefix encoding in slate would make it not such a big deal anyway...) just thinking that if this is a common pattern for all opendata systems it might be good to standardize on that

open-log/rfcs/0001-storage.md
```diff
@@ -0,0 +1,182 @@
...
+
+```
+Block Entry: | block_offset | cumulative_record_count | first_key |
+```
+
+This enables counting records in a range by scanning the LSM at the index level rather than reading all entries. Block boundaries may not align with the query range, so the first and last blocks in each overlapping level may need to be read for exact counts. An approximate count can be computed from the index alone.
+
+#### Bloom Filter Granularity
+
+SlateDB SSTs include bloom filters to accelerate point lookups. For open-log, the bloom filter should be keyed on the log key alone, not the composite SlateDB key which includes the sequence number. This allows the bloom filter to indicate whether a given log is present in an SST, reducing the blocks read during `scan` or `count` queries.
```
agavra: I mentioned this offline, but we should implement [Prefix Bloom Filters](github.com/facebook/rocksdb/wiki/Prefix-Seek#configure-prefix-bloom-filter) in Slate to support this, though as you discussed it would be extra nice if slate was aware of compound key structures for things like this and the comparators

open-log/rfcs/0001-storage.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Append-Only Scan Optimization
```
agavra, line 82: we also need an "inverse" ordered scan (we should scan from the last SST upwards instead of from the first matching SST downwards)

open-log/rfcs/0001-storage.md
```diff
@@ -0,0 +1,182 @@
+# RFC 0001: Open-Log Storage
```
cadonna: What about calling it simply `log`? It is already in the `opendata-oss` organization in the `opendata` repo. IMO, there are enough `open`s. 🙂

agavra: I like it, I can also rename `open-tsdb` -> `tsdb`

open-log/rfcs/0001-storage.md
hachikuji: Yes, I agree. I had forgotten that our tsdb prototype packed them into a "record tag. "

open-log/rfcs/0001-storage.md
hachikuji: Agreed. Nod to the Department of Redundancy Department.

open-log/rfcs/0001-storage.md
hachikuji: With 4 bits reserved each for the version and the type, that is 16 values for each, not 128. That ought to be good for starters, but I don't really know if one byte is worth the savings. In the prototype, we used one byte for the version and then we used one byte as the "record tag" which coupled together a type (4 bits) and  record attributes (4 bits). I'm sort of inclined to keep version as one byte rather than trying to pack it, but I don't feel too strongly about it.

open-log/rfcs/0001-storage.md
hachikuji: I implemented this suggestion. In the API, I use "Log" instead of "OpenLog." When referring to our implementation of log in this repository within the documentation, I used the term "OpenData-Log." How does that sound?

open-log/rfcs/0001-storage.md
hachikuji: Yeah, I remember Jay was really opposed to having headers in Kafka. It is tricky. The benefit of having headers in the Kafka protocol is that middleware could be designed to work at the level of Kafka's API. Otherwise, middleware needs to define its own schema, which makes adoption difficult. I see the API we're building here as more of a low-level primitive. Much like SlateDb itself, just with log backing instead of KV. In general, the less we can say about the structure of the data at this level, the better. Some use cases will not need headers and any cost associated with header is pure waste. On the other hand, if we built a Kafka abstraction on top of this log, we could fit headers into it. 

I think the interesting fork in the road here is in regard to how we expose batches and metadata around batches. In the initial draft, I treat batching as just a transport/storage optimization (similar to SlateDb itself). This means that an implementation which used headers would have to attach them at the record level (or design their own batching on top). This potentially amplifies the cost of headers which might be more efficiently stored at the batch level. But I figure it is a reasonable tradeoff and fundamentally necessary for the LSM data structure to control batching in storage. Compression should offset the cost of redundant header information anyway.

hachikuji: I added headers in the list of alternatives. I think it's a good example that demonstrates the kind of simplicity tradeoff that we expect in opendata structures.

open-log/rfcs/0001-storage.md
```diff
@@ -0,0 +1,199 @@
...
+### SST Representation
+
+OpenData-Log proposes two enhancements to SlateDB's SST structure to support efficient `scan` and `count` operations.
+
+#### Block Record Counts
+
+Each block entry in the SST index would include a cumulative record count:
+
+```
+Block Entry: | block_offset | cumulative_record_count | first_key |
```
apurvam, line 73: Is the cumulative record count the number of records in the log upto and including this block? Or the number of records cumulatively in the SST including this block? A short definition would be helpful.

open-log/rfcs/0001-storage.md
apurvam: I think 1 byte each is fine.

open-log/rfcs/0001-storage.md
apurvam: Referring to the log as "OpenData-Log" in documentation makes sense to me. It's a good convention IMO.

open-log/rfcs/0001-storage.md
agavra: Oooooof that's embarrassing ... I was like "8 bits = 256 values" just split that in half! Definitely 1 byte makes sense.

agavra (approved):
LGTM as well! Thanks @hachikuji excited to ship this and see the log live on ✊

PR #8, opened 2025-12-18
add rfc-0001 for tsdb storage

agavra:
This PR removes the storage design from the README and puts it into a dedicated RFC, adding a bit more clarity particularly in rejected alternatives.

### DIFF
--- a/timeseries/README.md
+++ b/timeseries/README.md
@@ -1,18 +1,18 @@
-# OpenTSDB
+# OpenData Timeseries
 
-OpenTSDB is a time series database with prometheus-like APIs which uses SlateDB
-as the underlying storage engine.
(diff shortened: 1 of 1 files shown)


apurvam (dismissed):
LGTM, assuming that the readme was just a port of the internal doc to begin with.

agavra:
@apurvam it was not, it makes significant simplifications (see rejected alternatives, all which were original parts of the design)

agavra:
@apurvam I agree, I'll add a quick section on top about the ingest/query path but the details will be left to implementation and/or another RFC. In particular, ingest is totally flexible (we could scrape, we could remote write, we could OLTP) and it's totally decoupled from storage.

apurvam (approved):
Thanks for adding the query flow section. Super helpful!

PR #13, opened 2025-12-19
use inverted index to serve labels/label_values instead of forward

agavra:
## Summary

The original approach was embarrassingly bad... I can't believe I merged it that way! I had to change it from the Aion implementation because we don't have explicit dictionaries. 
                                                                                                                                 
  - Optimizes `labels()` and `label_values()` endpoints to use inverted index keys instead of iterating through forward index
  - Implements hybrid approach: forward index for filtered queries (match[]), inverted index for unfiltered queries
  - Adds `InvertedIndexKey::attribute_range()` for efficient storage scans by label name
  - Adds `QueryReader::label_values()` method that scans only keys matching the requested label

## Test plan

  - [x] Existing tests pass (322 tests)
  - [x] Added unit test for `attribute_range()` key matching
  - [x] Existing integration tests for `labels()` and `label_values()` endpoints cover both filtered and unfiltered cases

## Checklist

- [x] Tests added/updated

### DIFF
--- a/open-tsdb/src/index.rs
+++ b/open-tsdb/src/index.rs
@@ -32,11 +32,18 @@ pub(crate) trait InvertedIndexLookup {
     /// Intersect posting lists for the given terms.
     /// Returns series IDs that match ALL terms.
     fn intersect(&self, terms: Vec<Attribute>) -> RoaringBitmap;
+
+    /// Get all attribute keys in the inverted index.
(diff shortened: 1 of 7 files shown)


agavra:
I'm going to optimistically merge this to unblock some perf testing, but it's definitely wroth specifying all the query paths in an RFC

PR #15, opened 2025-12-19
Rename open-log to log and open-tsdb to timeseries

hachikuji:
Renames modules as discussed on discord. Module naming is intended to be minimalistic, but descriptive.


agavra (approved):
I'll fix up documentation for the timeseries one after we merge (there's some conflicts with #8)

criccomini:
WDYT about opendata-common -> common?

PR #16, opened 2025-12-19
Add RFC for low-level timseries write API

hachikuji:
This RFC proposes a low-level `TimeSeries` API which is analogous to SlateDb's `Db` and OpenData's `Log`. The idea is to have a simple, permissive low-level API with all the complexity from systems like OTEL and Prometheus layered on top. I started with the write API because that is relatively simple compared to the read side.

I am hoping that each of our systems can expose similar low-level API with conventions and semantics which are roughly inline with SlateDb itself. Basically it should look and feel like SlateDb, but with a custom data model. We can then add APIs which cross-cut each system in higher-level traits. For example, we would probably introduce some trait to expose SlateDb's underlying checkpoint system.

### DIFF
--- a/timeseries/rfcs/0002-write-api.md
+++ b/timeseries/rfcs/0002-write-api.md
@@ -11,7 +11,7 @@ This RFC proposes a public write API for OpenData-TimeSeries, which is modeled a
 
 ## Motivation
 
-It would be useful to expose a low-level API to write time series data and enable ingestion flexibility. Although OTEL is likely to be the primary data model for ingestion, it has a more complicated data model with support for histograms and summary types. We can isolate representation of these higher-level metric types in a separate OTEL module without affecting the core time series simplicity (similar to Prometheus). In principle, OpenData-TimeSeries can accommodate other representations of time series data including those closer to user applications.  
+It would be useful to expose a low-level API to write time series data and enable ingest flexibility. Although OTEL is likely to be the primary data model for ingestion, it has a fairly complicated representation with support for higher-level metric types such as histograms and summaries. OpenData-TimeSeries aims to be simple in its core. We can isolate higher-level types in a separate OTEL module without compromising a simpler internal representation (similar to Prometheus). In principle, this allows OpenData-TimeSeries to accommodate other representations of time series data including those closer to user applications.    
 
 ## Goals
 
@@ -156,7 +156,7 @@ OpenData emphasizes minimalism and composition. The core data model is the small
 
 Richer abstractions are composed on top. For example, an OTEL mapping module would decompose histograms into multiple simple series (`_bucket`, `_sum`, `_count`) at ingestion time. Prometheus takes a similar approach for the same reasons.
 
-In a similar vein, Prometheus adds alphanumeric validation on label names in order to keep PromQL simple. However, nothing in our underlying timeseries data system actually relies on this assumption, and we can therefore layer it on top.  
+In a similar vein, Prometheus adds alphanumeric validation on label names in order to keep PromQL simple. However, nothing in our underlying timeseries data system actually relies on this assumption, and we can therefore layer it on top. Future systems may allow more localization even in the core APIs, so we aim to keep the internal representation as permissive as possible.  
 
 This design has several benefits:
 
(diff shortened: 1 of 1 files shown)


timeseries/rfcs/0002-write-api.md
```diff
@@ -0,0 +1,475 @@
...
+pub struct TimeSeries {
+    // Internal Tsdb - not exposed
+}
+
+impl TimeSeries {
+    /// Open or create a time series database with the given configuration.
+    pub async fn open(config: TimeSeriesConfig) -> Result<Self>;
+
+    /// Create an in-memory time series database for testing.
+    pub fn in_memory() -> Self;
```
agavra: curious if we need this if we expose `InMemoryStorage` (available in the `common` workspace)

timeseries/rfcs/0002-write-api.md
```diff
@@ -0,0 +1,475 @@
...
+    /// method can be used to ensure durability immediately.
+    pub async fn flush(&self) -> Result<()>;
+}
+
+/// Options for the write operation.
+#[derive(Debug, Clone, Default)]
+pub struct WriteOptions {
+    /// Wait for data to be flushed to durable storage before returning.
+    /// Default: false (return immediately after buffering)
+    pub await_durable: bool,
```
agavra, line 307: I'm curious if we need this for timeseries data - right now it's not super straightforward to get a handler to await durable with the batching mechanism I implemented. It would complicate the code with some annoying async book-keeping.

timeseries/rfcs/0002-write-api.md
```diff
@@ -0,0 +1,475 @@
...
+
+A separate module handles conversion from OpenTelemetry data models to the TimeSeries data model, following the approach used by Prometheus. This module is responsible for:
+
+1. **Type decomposition** - Converting complex OTEL types (histograms, summaries) into multiple simple series
+2. **Label sanitization** - Ensuring label names conform to Prometheus conventions
+3. **Attribute mapping** - Converting OTEL resource/scope/metric attributes to labels
+
+```rust
+/// Module for converting OpenTelemetry metrics to TimeSeries format.
+pub mod otel {
```
agavra, line 341: when we do eventually introduce this, I think it should be behind a `cfg` rust flag so that we don't need to include protobuf dependencies unless required

timeseries/rfcs/0002-write-api.md
```diff
@@ -0,0 +1,475 @@
...
+- **Composable** - Higher layers can implement any histogram scheme without core changes
+- **Permissive** - No opinions about bucket boundaries, quantiles, or naming conventions
+
+Prometheus takes a similar approach for the same reasons.
+
+**Trade-offs**: Users must use the OTEL module (or implement their own decomposition) to ingest histogram/summary data.
+
+## Open Questions
+
+1. **Metric metadata storage**: Where should metadata be stored? With the series data, or in a separate metadata store?
```
agavra: I think it's fine to have it in the forward index for now and we can figure out if we need to split it out if need be assuming that's what you meant by "metric metadata". Eventually we can split it out if it ends up causing a significant overhead to store the same metadata over and over again with each series for the same metric. Every additional layer of indirection is more bookkeeping :/

agavra (approved):
This looks great, thanks @hachikuji! No serious comments or concerns. Only question around `await_durable` and whether that's necessary for timeseries ingestion.

timeseries/rfcs/0002-write-api.md
hachikuji: Yeah, let's drop it

timeseries/rfcs/0002-write-api.md
hachikuji: I didn't see the harm exposing it at the lower level. I think our default would just be false.

timeseries/rfcs/0002-write-api.md
hachikuji: Yes, agreed.

timeseries/rfcs/0002-write-api.md
hachikuji: Yes, I agree.

timeseries/rfcs/0002-write-api.md
hachikuji: Perhaps some users would want to write billing telemetry at this level and might like a strong guarantee.

agavra: That's fair - I think we can eventually support it. Maybe we'll wait to implement it until we have slatedb-wal factored out or a pluggable slate memtable so that we don't need to handle the synchronization between the buffer and slate's `await_durable` mechanism ourselves (it was a bit of a pain in slatedb to handle that and i really don't want to custom implement it if we can avoid it).

PR #17, opened 2025-12-29
Timeseries stubs

hachikuji:
Adds stubbing for timeseries RFC 2. I decided to flatten SeriesMetadata to simplify the API. I also made some other minor changes to the RFC. I imagine the following evolution of the implementation internally:

1. Implement basic API on top of existing internal data structures.
2. Migrate OTEL implementation on top of new API.
3. Consolidate internal/public data structures.

### DIFF
--- a/timeseries/rfcs/0001-tsdb-storage.md
+++ b/timeseries/rfcs/0001-tsdb-storage.md
@@ -57,9 +57,9 @@ the query path helps explain the storage model. A timeseries database has three
 storage components:
 
 - **Raw series storage**: Stores `(timestamp, value)` pairs keyed by series ID
-- **Inverted index**: Maps attribute/value pairs to series IDs, enabling efficient
+- **Inverted index**: Maps label/value pairs to series IDs, enabling efficient
   label selector matching (e.g., `{job="api", status="500"}`)
-- **Forward index**: Maps series IDs to their canonical attribute specifications
+- **Forward index**: Maps series IDs to their canonical label specifications
   (all labels as UTF-8 strings), used to resolve series discovered via the
   inverted index
 
@@ -68,7 +68,7 @@ method="GET", status="200"})` is served:
 
 1. Parse the query into a query plan
 2. Use the inverted index to find series matching `{__name__="metric", path="/query", method="GET", status="200"}`
-3. Use the forward index to resolve each series ID to its full attribute set (including `instance` for grouping)
+3. Use the forward index to resolve each series ID to its full label set (including `instance` for grouping)
 4. Sum the values for matching series and return the result
 
 ### Record Layout
@@ -158,8 +158,8 @@ architectures (x86, ARM).
 | `0x00`  | *(reserved)*        | Reserved for future use                                   |
 | `0x01`  | `BucketList`        | Lists the available time buckets                          |
 | `0x02`  | `SeriesDictionary`  | Stores the mapping of series to series IDs                |
-| `0x03`  | `ForwardIndex`      | Stores the canonical attribute set for each series        |
-| `0x04`  | `InvertedIndex`     | Maps attribute/value pairs to posting lists of series IDs |
+| `0x03`  | `ForwardIndex`      | Stores the canonical label set for each series            |
+| `0x04`  | `InvertedIndex`     | Maps label/value pairs to posting lists of series IDs     |
 | `0x05`  | `TimeSeries`        | Holds the raw time-series payloads                        |
 
 ## Record Definitions & Schemas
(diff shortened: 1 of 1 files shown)


timeseries/src/error.rs
```diff
@@ -0,0 +1,82 @@
...
+/// - [`Storage`](Error::Storage): Errors from the underlying storage layer,
+///   such as I/O failures or corruption.
+/// - [`Encoding`](Error::Encoding): Errors during serialization or deserialization
+///   of time series data.
+/// - [`InvalidInput`](Error::InvalidInput): Errors caused by invalid parameters or
+///   arguments provided by the caller.
+/// - [`Internal`](Error::Internal): Unexpected internal errors that indicate bugs
+///   or invariant violations.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Error {
```
agavra, line 25: we should consider naming it `TimeseriesError` so every db doesn't have an error class named `Error`. We also already have OpenTsdbError in util.rs, we should consolidate the two

timeseries/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+//! - **Series**: A time series identified by labels, containing timestamped samples.
+//! - **Label**: A key-value pair that identifies a time series.
+//! - **Sample**: A single data point with a timestamp and value.
+//!
+//! # Example
+//!
+//! ```ignore
+//! use timeseries::{TimeSeries, Config, Series};
+//!
+//! #[tokio::main]
```
agavra, line 25: should we update `main.rs` to match or just remove it altogether?

timeseries/src/series.rs
```diff
@@ -0,0 +1,319 @@
...
+            labels: Vec::new(),
+            metric_type: None,
+            unit: None,
+            description: None,
+            samples: Vec::new(),
+        }
+    }
+
+    /// Adds a label to the series.
+    pub fn label(mut self, name: impl Into<String>, value: impl Into<String>) -> Self {
```
agavra, line 214: consider `add_label` and `add_sample` to make it clear that you can add multiple (as opposed to something like `unit` which overwrites)

timeseries/src/timeseries.rs
```diff
@@ -0,0 +1,143 @@
+//! Core TimeSeries implementation with write API.
```
agavra, line 1: should this just replace `tsdb.rs` or are you planning on using that internally here?

timeseries/src/series.rs
```diff
@@ -0,0 +1,319 @@
+//! Core data types for OpenData TimeSeries.
```
agavra, line 1: we should consolidate this and what's already in `model.rs` -- also note that in your comment "Migrate OTEL implementation on top of new API" the current implementation in this repo is already very close to this, it doesn't use OTEl at all (see `tsdb.rs` which has a main ingest API of `pub(crate) async fn ingest_samples(&self, samples: Vec<SampleWithAttributes>)`)

I think it'd probably be easier to rename the existing `model.rs` structs (e.g. `SampleWithAttributes` to `Series`, and `Attribute` to `Label`) to match your proposal so the rest of the code just comes along for the ride.

timeseries/src/config.rs
```diff
@@ -0,0 +1,76 @@
+//! Configuration options for OpenData TimeSeries operations.
```
agavra, line 1: Note that we already have promql/config.rs, which is the Server configuration

agavra (commented):
The actual proposal LGTM but this PR creates a large fork from the code that's already in opendata/timeseries. I've commented in-line about where the equivalent items exist in the current implementation.

hachikuji:
> The actual proposal LGTM but this PR creates a large fork from the code that's already in opendata/timeseries. I've commented in-line about where the equivalent items exist in the current implementation.

@agavra Yep, I understood there was some duplication. The rough plan I had is in the description. We could cut straight through some of it if you prefer. Usually I find it a bit easier to start with the API that you want and consolidate internally rather than trying to factor out the API from the internal implementation, but I don't think it makes a dramatic difference.

agavra:
> @agavra Yep, I understood there was some duplication. The rough plan I had is in the description. We could cut straight through some of it if you prefer. Usually I find it a bit easier to start with the API that you want and consolidate internally rather than trying to factor out the API from the internal implementation, but I don't think it makes a dramatic difference.

got it, that makes sense to me. I wanted to make sure you had seen those other components and to point out which ones refer to what so it's easier when we do consolidate them. I have no preference on the order of which we do it so we can merge this PR with the duplication for now.

hachikuji:
Makes sense. Let me take a quick look and see if there are any obvious candidates that can be consolidated now with minimal refactoring.

hachikuji:
@agavra The label/attribute consolidation was straightforward (just renaming), so I pushed that into this patch. I suggest we do the rest in subsequent patches. How does that sound?

PR #18, opened 2026-01-05
Log sequence tracking RFC

hachikuji:
The initial log RFC is vague about sequence number tracking. This patch suggests a block-based approach implemented on top of SlateDb.

### DIFF
--- a/log/rfcs/0001-storage.md
+++ b/log/rfcs/0001-storage.md
@@ -247,7 +247,7 @@ The simpler key+sequence encoding preserves key ordering and avoids the collisio
 
 ### SlateDB Sequence Numbers
 
-SlateDB maintains its own internal sequence number for MVCC versioning. Ideally, OpenData-Log would reuse this counter rather than implementing separate tracking with `LastBlock` records. However, SlateDB's sequence number is not currently exposed in its public API. If SlateDB adds support for accessing or influencing sequence assignment in the future, this design could be simplified.
+SlateDB maintains its own internal sequence number for MVCC versioning. Ideally, OpenData-Log could reuse this counter rather than implementing separate tracking with `LastBlock` records. However, SlateDB's sequence number is not currently exposed in its public API. Furthermore, since the sequence number is embedded into the key, we would need to align the sequence number prior to writing. Current proposals to expose SlateDb sequence numbers do not offer such a mechanism, but we can reevaluate once the effort is complete. Finally, we should reserve the ability to offer an `edit` API so that a corrupt/incorrect record may be overwritten. This may be more difficult if we are too coupled with the SlateDb sequence number.  
 
 ### Headers
 


log/rfcs/0001-storage.md
```diff
@@ -55,12 +55,41 @@ In practice, users are likely to use fixed-length keys, which avoids this issue
...
+Sequence numbers are assigned from a single counter that is maintained by the SlateDB writer and is incremented after every append. Each key's log is monotonically ordered by sequence number, but the sequence numbers are not contiguous—other keys' appends are interleaved in the global sequence. Additionally, sequence numbers may have gaps due to crash recovery (see below). The only guarantee is monotonicity: within a key's log, sequence numbers are strictly increasing.
 
 This approach simplifies ingestion by avoiding per-key sequence tracking. The trade-off is that sequence numbers do not reflect the count of entries within a key's log.
 
 If SlateDB supports multi-writer in the future, each writer would maintain its own sequence counter. This design assumes each key would still have a single writer—interleaving appends from multiple writers to the same key would break monotonic ordering within that key's log.
 
+#### Block-Based Sequence Allocation
```
agavra, line 64: I'm wondering why we don't use the sequence number maintained by Slate itself? I believe we want to expose that anyway (see https://github.com/slatedb/slatedb/issues/1138#issuecomment-3700157484 - it's pretty deep in there but the proposal is good)

agavra: ah! just saw your comment at the bottom. yeah that makes sense to me as an interim thing, we might want to just prioritize the slate ticket thought

agavra (approved):
LGTM, though I think we should consider prioritizing the SlateDB change to expose the internal seqnum

hachikuji:
@agavra To be honest, I'm on the fence about exposing the sequence number in SlateDb. I guess my concern is how it ties down the implementation for the future. Perhaps as long as it works within a multi-writer context, it's fine. And maybe we have to do it anyway for some use cases. I'll read the thread and try to get a better read on it.

apurvam (approved):
LGTM! I agree with almog's comment about prioritizing generalizing and reusing the slateDB sequence number if it is already playing a similar role. That would be in the spirit of the substrate.

criccomini:
> LGTM, though I think we should consider prioritizing the SlateDB change to expose the internal seqnum

I have been wanting to do this for a long time but never got around to it. If you guys want to PR it, that'd be 🔥 .

hachikuji:
I looked at the suggestion in the comment here: https://github.com/slatedb/slatedb/issues/1138#issuecomment-3700157484. The API looks reasonable to me (I'll see if I can pick it up since I've been looking for some slatedb work), but it wouldn't solve our problem here. We need the sequence number to be embedded in the key, which means we need to know what the sequence number will be before writing. Getting keys to align with writes seems a little slippery to me. There is also a benefit to decoupling the two. The nice thing about the LSM representation is that it would be possible to overwrite a bad record. That wouldn't be possible anymore if we were tied to the internal sequence number.

agavra:
> We need the sequence number to be embedded in the key, which means we need to know what the sequence number will be before writing.

🤦 yes... that makes sense to me. Though FWIW this could be a use case for decoupling the WAL and WAL compactors in Rohan's terminology. You know the seqnum when its inserted into the WAL, we could then transform that into a full key when it's written to L0 using the sequence number.

Anyway, I'm OK decoupling the two. I imagine it will make things easier in the long run.

PR #21, opened 2026-01-06
remove OpenTsdbError in favor of TimeseriesError

agavra:
## Summary

After #17 we now have two errors types for `timeseries` this consolidates the two and renames `Error` -> `TimseriesError` so it makes codesearch easier going forward

## Related Issues

N/A

## Test Plan

Ran existing tests

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/timeseries/src/delta.rs
+++ b/timeseries/src/delta.rs
@@ -6,12 +6,13 @@ use std::sync::atomic::{AtomicU32, Ordering};
 use dashmap::DashMap;
 
 use crate::{
+    error::TimeseriesError,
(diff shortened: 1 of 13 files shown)


agavra:
@hachikuji I don't feel too strongly, I can make that change back. Just seems like we'll have a lot of `Error` classes in one code base 😅

hachikuji:
Yeah, that is true. But probably unlikely we would have  more than one or two in scope at any time. I feel like rust offers enough alias flexibility like using`timeseries::Error` directly or `timeseries::Error as TimeSeriesError` that the simpler naming is preferable.

PR #28, opened 2026-01-07
Add basic sequence number bookkeeping

hachikuji:
This patch contains basic sequence number initialization and allocation logic. For now, the logic is synchronous. Eventually we can move it to a separate thread so that we are unlikely to block on sequence number allocation.

### DIFF
--- a/log/src/lib.rs
+++ b/log/src/lib.rs
@@ -41,13 +41,13 @@
 //! }
 //! ```
 
-mod codec;
 mod config;
 mod error;
 mod log;
 mod model;
 mod reader;
 mod sequence;
+mod serde;
 
 pub use config::{Config, CountOptions, ScanOptions, WriteOptions};
 pub use error::{Error, Result};
--- a/log/src/sequence.rs
+++ b/log/src/sequence.rs
@@ -28,8 +28,8 @@ use bytes::Bytes;
 use opendata_common::{Record, Storage};
 use tokio::sync::Mutex;
 
-use crate::codec::{LastSeqBlockKey, SeqBlock};
 use crate::error::Result;
+use crate::serde::{LastSeqBlockKey, SeqBlock};
 
 /// Default block size for sequence allocation.
 pub(crate) const DEFAULT_BLOCK_SIZE: u64 = 4096;
--- a/log/src/serde.rs
+++ b/log/src/serde.rs
@@ -1,6 +1,6 @@
 #![allow(dead_code)]
 
-//! Codec for log storage encoding and decoding.
+//! Serde for log storage
 //!
 //! This module provides encoding and decoding for log records stored in SlateDB.
 //! The encoding scheme is designed to preserve lexicographic ordering of keys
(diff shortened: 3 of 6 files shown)


log/rfcs/0001-storage.md
```diff
@@ -45,28 +45,29 @@ This encoding preserves lexicographic key ordering, enabling key-range scans. En
 
 #### TerminatedBytes
 
-A `TerminatedBytes` is a variable-length byte sequence that terminates with a `0xFF` delimiter. This delimiter provides an unambiguous boundary between the user key and the sequence number, which is necessary for correct lexicographic ordering when keys have variable length.
+A `TerminatedBytes` is a variable-length byte sequence that terminates with a `0x00` delimiter. This delimiter provides an unambiguous boundary between the user key and the sequence number, which is necessary for correct lexicographic ordering when keys have variable length.
```
agavra, line 48: after we merge this I'll update the vector RFC to match, but we should put all utilities related to this in `common` since we'll definitely need it in both timeseries and vector

log/src/codec.rs
```diff
@@ -0,0 +1,810 @@
+#![allow(dead_code)]
+
+//! Codec for log storage encoding and decoding.
```
agavra: nit: if we want to make this consistent with `timeseries` we should put this all in a `serde` module with individual files for each more complex type

I'm OK either way, we can also rename serde -> codec (I like that better anyway) 🤷

timeseries/src/promql/evaluator.rs
```diff
@@ -11,10 +11,10 @@ use promql_parser::parser::{
 };
 
 use crate::index::ForwardIndexLookup;
+use crate::model::Label;
```
agavra, line 14: something is weird with this PR, looks like it's pulling in my other commits? I wonder what's up maybe needs some rebase?

agavra: oh weird this is my fault - I think somehow my branch was accidentally pushed into yours? I'm so confused... it says I force pushed? I don't think I did... figuring this out.

hachikuji: Hmm, weird. I'll fix.

agavra: this is on me, I figurd out what happened. just fixed it by I think you need to rebase and cherry-pick your commit on the new main

agavra: I think I accidentally force pushed my local main to this repo instead of for the opendata-oss/site repo I had meant to push to, reverting the commit I had for #25 so that corrupted the tree. I just re-introduced it with #30 

(FYI I disabled force push bypass for us, I thought it was already disabled but my branch protection ruleset was off)

hachikuji: Gotcha. No worries.

log/src/codec.rs
hachikuji: Makes sense. Let's go with `serde`.

PR #29, opened 2026-01-07
initial commit for vector storage rfc

agavra:
this is an initial pass at a vector database system storage design - it's heavily influenced by Turbopuffer (SPFresh design) and our existing timeseries/log storage design. the main goal is to unblock initial development so we can get to a working prototype and learn a lot from it. in other words, it's OK if our design is suboptimal in v0 we can go back and iterate once we've gotten some more hands on experience.

### DIFF
--- a/vector/rfcs/0001-storage.md
+++ b/vector/rfcs/0001-storage.md
@@ -63,9 +63,10 @@ index structures are stored as key-value pairs in the LSM tree.
 │   │                  Write Path                         │   │
 │   │                                                     │   │
 │   │   1. Vector written to WAL (durability)             │   │
-│   │   2. Assigned to nearest centroid(s)                │   │
-│   │   3. Posting list updated via merge operator        │   │
-│   │   4. Vector data + metadata written                 │   │
+│   │   2. External ID looked up / allocated internal ID  │   │
+│   │   3. Assigned to nearest centroid(s)                │   │
+│   │   4. Posting list updated via merge operator        │   │
+│   │   5. Vector data + metadata + dictionary written    │   │
 │   └─────────────────────────────────────────────────────┘   │
 │                                                             │
 │   ┌─────────────────────────────────────────────────────┐   │
@@ -74,7 +75,7 @@ index structures are stored as key-value pairs in the LSM tree.
 │   │   Centroids:     [chunk 0] [chunk 1] ... [chunk N]  │   │
 │   │                  (loaded into HNSW for navigation)  │   │
 │   │                                                     │   │
-│   │   Posting Lists: centroid_id → RoaringBitmap        │   │
+│   │   Posting Lists: centroid_id → RoaringTreemap       │   │
 │   │                  (vector IDs per cluster)           │   │
 │   │                                                     │   │
 │   │   Deleted:       centroid_id=0 → deleted vector IDs │   │
@@ -83,8 +84,9 @@ index structures are stored as key-value pairs in the LSM tree.
 │   ┌─────────────────────────────────────────────────────┐   │
 │   │                  Vector Storage                     │   │
 │   │                                                     │   │
+│   │   IdDictionary:  external_id → internal vector_id   │   │
 │   │   VectorData:    vector_id → f32[dimensions]        │   │
-│   │   VectorMeta:    vector_id → metadata fields        │   │
+│   │   VectorMeta:    vector_id → external_id + metadata │   │
 │   │   MetadataIndex: (field, value) → vector IDs        │   │
 │   └─────────────────────────────────────────────────────┘   │
 │                                                             │
@@ -135,21 +137,59 @@ matches numeric ordering of encoded values.
 
 Key identifiers used in this RFC:
 
-- `vector_id` (u32): Internal vector identifier, unique within the namespace
+- `vector_id` (u64): Internal vector identifier, system-assigned, unique within the namespace
+- `external_id` (string): User-provided identifier, max 64 bytes, maps to internal vector_id
 - `centroid_id` (u32): Cluster centroid identifier
 - `chunk_id` (u32): Centroid chunk identifier
 
-### Vector ID Generation
+### External and Internal IDs
 
-Vector IDs are user-provided u32 integers. Clients must ensure uniqueness per namespace; in
-practice, many workloads keep a dictionary that maps their external UUIDs or strings onto a compact
-range of integers. Maintaining roughly monotonically increasing IDs improves Roaring bitmap
-compression because sequential IDs cluster well, but the storage engine does not enforce any
-allocation policy.
+Users provide **external IDs**—arbitrary strings up to 64 bytes—to identify their vectors. The
+system maintains an `IdDictionary` that maps each external ID to a system-assigned **internal ID**
+(u64). Internal IDs are used in all index structures (posting lists, metadata indexes) because:
 
-**Collision / upsert behavior:** TBD. A follow-up RFC will define whether writers may overwrite an
-existing `vector_id` or if a prior delete is required. Until then, clients should treat duplicate
-IDs as undefined behavior.
+1. Fixed-width u64 keys enable efficient bitmap operations (RoaringTreemap)
+2. Monotonically increasing IDs improve bitmap compression (sequential IDs cluster well)
+3. Decoupling external from internal IDs allows the system to manage ID lifecycle
+
+**Upsert Behavior:**
+
+When inserting a vector with an external ID that already exists:
+
+1. Look up the existing internal ID from `IdDictionary`
+2. Delete the old vector: add internal ID to deleted bitmap, tombstone `VectorData` and `VectorMeta`
+3. Allocate a new internal ID
+4. Write new vector data with the new internal ID
+5. Update `IdDictionary` to point to the new internal ID
+
+This "delete old + insert new" approach ensures that posting lists and metadata indexes are updated
+correctly via merge operators, without requiring expensive read-modify-write cycles to remove the
+old internal ID from every index entry.
+
+### Block-Based ID Allocation
+
+Internal vector IDs are allocated from a monotonically increasing counter using block-based
+allocation (similar to the Log RFC's sequence allocation). Rather than persisting the counter after
+every insert, the writer pre-allocates a block of IDs and records the allocation in a `SeqBlock`
+record.
+
+**Allocation procedure:**
+
+1. On initialization, read the `SeqBlock` record to get the last allocated range `[base, base+size)`
+2. Allocate a new block starting at `base + size` and write a new `SeqBlock` before processing
+3. During normal operation, assign IDs from the current block, incrementing after each insert
+4. When the current block is exhausted, allocate a new block and write an updated `SeqBlock`
+
+**Recovery:**
+
+On crash recovery, read the `SeqBlock` and allocate a fresh block starting after the previous range.
+Any IDs allocated but not used before the crash are skipped. This may create gaps in the ID space,
+but monotonicity is preserved.
+
+**Block sizing:**
+
+Block size is an implementation detail balancing write amplification (larger blocks reduce
+`SeqBlock` write frequency) against ID space efficiency (smaller blocks waste fewer IDs on crash).
 
 ### Standard Key Prefix
 
@@ -178,7 +218,8 @@ with future schema changes that may require sub-type discrimination.
 - `Array<T>`: `count: u16` followed by `count` serialized elements of type `T`.
 - `FixedElementArray<T>`: Serialized elements back-to-back with no count prefix; length derived from
   buffer size divided by element size.
-- `RoaringBitmap`: Standard Roaring bitmap serialization format for compressed u32 integer sets.
+- `RoaringTreemap`: Roaring treemap serialization format for compressed u64 integer sets (64-bit
+  extension of Roaring bitmap).
 
 **Key Encodings** (big-endian for lexicographic ordering):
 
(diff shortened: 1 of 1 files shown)


vector/rfcs/0001-storage.md
```diff
@@ -0,0 +1,626 @@
...
+- Query execution and search algorithm details
+- Quantization strategies (PQ, SQ)
+- Distributed/sharded deployment
+
+## Design
+
+### Architecture Overview
+
+Each namespace corresponds to a single SlateDB instance with a global SPANN index. Vector data and
+index structures are stored as key-value pairs in the LSM tree.
```
agavra, line 56: this is one thing I was debating, whether to make namespaces part of the data model or to just require multiple slateDB instances. the overhead of one is pretty small (active used memory of the L0) especially if we share a WAL (WAL-compactor model). one less level of hierarchy makes it easy to reason about and easy to load individual namespaces.

for reference, in turbopuffer a namespace would be something like a single Cursor Codebase or a single Notion Teamspace

vector/rfcs/0001-storage.md
```diff
@@ -0,0 +1,626 @@
...
+  buffer size divided by element size.
+- `RoaringBitmap`: Standard Roaring bitmap serialization format for compressed u32 integer sets.
+
+**Key Encodings** (big-endian for lexicographic ordering):
+
+- `NullTerminatedUtf8`: UTF-8 bytes followed by a `0x00` terminator, using FoundationDB tuple
+  encoding for escaping:
+  - `0x00` in payload → `0x00 0xFF` (escaped null)
+  - `0xFF` in payload → `0xFF 0x00` (escaped 0xFF to avoid ambiguity)
+  - Terminator is an unescaped `0x00` (i.e., `0x00` not followed by `0xFF`)
```
agavra: I'll update this to match what @hachikuji is doing for log when that's finalized and just reference that

vector/rfcs/0001-storage.md
```diff
@@ -0,0 +1,626 @@
...
+
+**Schema Evolution:**
+
+The `schema_version` field tracks metadata schema changes only (not structural changes to the
+collection). Supported evolutions:
+
+- **Adding metadata fields**: New fields can be appended to `metadata_fields`. Existing vectors
+  without the field return null/missing for that field.
+- **Enabling indexing**: A field's `indexed` flag can be changed from false to true. A background
+  job must rebuild the `MetadataIndex` entries for existing vectors.
```
agavra, line 308: still TBD on how we do this with single-writers

vector/rfcs/0001-storage.md
```diff
@@ -0,0 +1,626 @@
...
+- Posting lists are balanced by the hierarchical clustering algorithm (target size configurable,
+  e.g., 1000-5000 vectors)
+- LIRE maintenance splits oversized postings and merges undersized ones
+- The deleted posting list (centroid_id = 0) is used to filter search results without per-vector
+  tombstone records
+
+**Merge Operators:**
+
+Posting lists are updated via SlateDB merge operators rather than read-modify-write. This enables
+concurrent updates without locking:
```
agavra: we should definitely batch these together in memory to avoid the problems we saw with high write amplification and serde overhead with `timeseries`

vector/rfcs/0001-storage.md
```diff
@@ -0,0 +1,626 @@
...
+Record keys are built by concatenating big-endian binary tokens. Lexicographical ordering of keys
+matches numeric ordering of encoded values.
+
+Key identifiers used in this RFC:
+
+- `vector_id` (u32): Internal vector identifier, unique within the namespace
+- `centroid_id` (u32): Cluster centroid identifier
+- `chunk_id` (u32): Centroid chunk identifier
+
+### Vector ID Generation
```
agavra: we should consider bringing this into the storage format so we can support upserts by deleting old ones and inserting new IDs - this will just require that we maintain the dictionary from external id -> id internally for retrieval/deletion of vectors by id

agavra: After prototyping the insert API I'm a bit more confident we should bring this in house. The problem is that if implement upserts as internally a delete + a new record then u32 may not be sufficient (which sucks because roaring bitmaps are best implemented with u32s). The max u32 is ~4.2B which would be the number of insertions/updates possible before bricking the database.

agavra: see discussion here: https://github.com/opendata-oss/opendata/pull/34/changes#diff-a81c71541f1a68fa9abb14ede2715ee442ab7b79c94af96e5f9f0f5595ff00a9R439-R467

vector/rfcs/0001-storage.md
apurvam: This is a good discussion to have. Including a namespace part of the datamodel gives more flexibility. The system could have exactly one namespace per Slate, or multiple in that event. 

Also, to make sure I understand, when you say "Each namespace corresponds to a single SlateDB instance with a global SPANN index" , 'global SPANN index' means that it covers every vector in the namespace?

vector/rfcs/0001-storage.md
```diff
@@ -0,0 +1,626 @@
...
+- **Metadata index**: Inverted index mapping metadata field/value pairs to vector IDs
+
+To illustrate, here's how the query
+`find top-10 similar to query_vec where category="shoes" and price < 50` is served:
+
+1. Search the centroid index to find the k nearest centroids to `query_vec`
+2. Load posting lists for those centroids from disk
+3. Intersect candidate vector IDs with the metadata filter `{category="shoes", price < 50}` using
+   the inverted index
+4. Compute exact distances for remaining candidates
```
apurvam, line 130: This exact distance computation will mean that you need to load the vector data for each ID and then run the distance function right? So the actual data access is: find centroids, load posting lists, filter, load N vectors from the forward index, run distance function. Further, these N vectors are unlikely to have any data locality right? The random access may be problematic for latency and costs (if they need to be fetched from S3).

This can likely be optimized. Either way, its definitely worth detailing that this exact distance calculation will inolve random access across the LSM tree.

vector/rfcs/0001-storage.md
```diff
@@ -0,0 +1,626 @@
...
+**Key Layout:**
+
+```
+┌─────────┬─────────────┬──────────┐
+│ version │ record_tag  │ chunk_id │
+│ 1 byte  │   1 byte    │ 4 bytes  │
+└─────────┴─────────────┴──────────┘
+```
+
+- `chunk_id` (u32): Identifies the chunk within the namespace, starting from 0
```
apurvam, line 332: I may have missed it but why do chunks need ids? Is it just nice to have? I didn't see use case for looking up a chunk id during a query.

vector/rfcs/0001-storage.md
apurvam: I probably have not understood it properly, but merge is a read time operation right? Is the implication here that during compaction, the merge operator will be used to combine posting lists while reading to write out updated posting lists post-compaction?

apurvam (commented):
This looks pretty solid to me. I spent time understanding the underlying concepts and I must say that it fits really well with the LSM tree stucture.

I think the only main concern is the random access when you go from vector ids to loading vectors to do the exact distance calculations. It feels like that can get pretty slow the way the data is laid out in this RFC.

vector/rfcs/0001-storage.md
agavra: > This is a good discussion to have. Including a namespace part of the datamodel gives more flexibility. The system could have exactly one namespace per Slate, or multiple in that event.

Yeah, I'm the most torn about this. If we expect to have millions of namespaces with small(er) vectors per namespace (the cursor/notion use cases) then having namespace be a first class Slate thing is probably necessary. Though I'm wondering if this is something we can push into the substrate where you can "share" a WAL + Memtable but treat them like different slates...

> 'global SPANN index' means that it covers every vector in the namespace?

it covers all the centroids for a namespace and that's what you hold in memory. The alternative is splitting up the centroids into multiple segments and then scatter-gather.

vector/rfcs/0001-storage.md
agavra: yeah I didn't phrase this super clearly

the problem we saw with timeseries is that if we use merge liberally our read performance sucks because we're spending a bunch of time deserializing and merging entries on every read. OTOH if we read-modify-write we have signfiicant write amplification that requires deserializing and reserializing on each insert which tanked write throughput. 

both sucked -- former for reads, latter for writes. so the solution is to batch writes together in memory and then use merges (basically it's an in-between where we don't read-modify-write, but we also don't merge-on-every-insert). The downside of this is that we need to maintain our own memtable basically and separate out the WAL in order to get low latency durable writes. 

It's a bit of an unsovled (though not unsolvable) problem with our architecture. I'm hoping maybe @rodesai has some fun ideas that come out of his perf work with tsdb.

------

as an aside -- the "without locking" bit is misleading, that's just some AI slop leftover I didn't catch. We wouldn't need locking either way b/c we have atomic write batches.

vector/rfcs/0001-storage.md
agavra: yeah this is a very good observation, I don't know if there's any good way to have data locality for those vectors without storing them directly in the centroid posting list, which is prohibitively expensive (can't use roaring bitmap and docid delta strategies and they're copied multiple times to neighboring centroids). 

I feel like this is just a fundamental problem to any search system. It was like fetching the forward index for documents in search to compute ranking features. 

I wouldn't be totally surprised if caching is very effective here. If it's anything like search use cases there's going to be a zipf distribution of how frequently specific vectors are retrieved. 

as with many slate-backed systems, I think the recommendation is to fit the entire working set on disk so you avoid S3 lookups altogether and only deal with that on recovery/restore scenarios

vector/rfcs/0001-storage.md
agavra: just so they don't share the same record key - we always scan them fully (`scan(version|0x02|*)`) on startup to load up the HNSW graph. In the future we could consider adding an extra layer of hierarchy and page centroid in/out based on chunk id but that's out of scope for this. I did debate whether we need chunking at all but the problem with putting each centroid in its own key is that we're exploding the number of keys we need to fetch on startup == higher startup latency (just more raw data fetched)

vector/rfcs/0001-storage.md
apurvam: Makes sense. Doesn't that suggest that putting the names pace in the data model is beneficial, or even making it part of the key in some way? In the cursor analogy, then you would find all vectors for the project contiguously, making caching and querying far more efficient.

vector/rfcs/0001-storage.md
apurvam: Makes sense!

vector/rfcs/0001-storage.md
agavra: yeah that's the suggestion in this RFC. the namespace is either explicitly in the model (put it in the key) or implicitly (1:1 relationship between a SlateDB instance and a namespace). The problem is within a namespace you can still have millions of vectors, where your problem rings true -- how do you get data locality for vectors within a namespace?

vector/rfcs/0001-storage.md
apurvam: > both sucked -- former for reads, latter for writes. so the solution is to batch writes together in memory and then use merges (basically it's an in-between where we don't read-modify-write, but we also don't merge-on-every-insert). 

Understood the problem with merge only on read, and the read-modify-write inefficiencies. I'm still not clear what the proposal is. Do you mean that the flush to L0 will merge buffered entries? Or that the merge operation will be while reading from the memtable? or something else?

vector/rfcs/0001-storage.md
apurvam: Yea, that is indeed a difficult problem. It may be a non issue in practice for the relevant use cases though, because the access pattern may dictate that you have to page all vectors in a logical unit (ie namespace) into memory regardless of how big the namespace is.

vector/rfcs/0001-storage.md
agavra: I mean we buffer up writes before we even put them into slate. An example here is if we were to get writes to vectors with ids 100-199 within the buffer interval and we have 10 centroids, we'll collect the updates to the posting lists in memory and then just write 10 merges instead of 100 (i.e. we'll apply ongoing batch merges in memory instead of apply them at the slateDB level). 

if we want to make the in-memory buffer queryable we need to:
1. separate WAL from Slate ingest so we get durable writes (i.e. we write to WAL, then buffer in memory, then flush to slate on a timer)
2. make the flushing operations atomic (i.e. we need to make sure that we don't pause writes when we flush data from the buffer into slatedb and that we don't read duplicates while flushing, so we need a consistent snapshot of slatedb and the buffers and them lock + update them when the flush is done)
3. maybe have something like github.com/slatedb/slatedb/issues/1105 to avoid relying on the SlateDB memtable in addition to our own memory buffer (which is basically a memtable)

I had prototyped some of this stuff with timeseries but didn't go through committing it because it was quite error prone. Instead i just added a delay before things were visible (so buffer it up, then write to slate -- read went directly to slate so you didn't see recent writes during the buffer interval)

apurvam: Ah. That makes a lot of sense. So to be totally clear, the 'merges' you are talking about here are not using slateDBs merge operator. they are a custom operation in the app layer (ie outside of slate db).

agavra: 😵‍💫 almost but it's a little more complicated - we use in-memory merges for the buffered entries (not slateDB merge) but once we merge the in-memory buffer we write it with a `merge` into slateDB (using slateDB merges) to avoid any read-modify-write loop. Both merge operators have the same logic, but one works on `RoaringBitmap` directly and the other works on `Bytes`

vector/rfcs/0001-storage.md
agavra: Updated the RFC.

vector/rfcs/0001-storage.md
```diff
@@ -0,0 +1,764 @@
...
+│  │  name:       Utf8                                        │  │
+│  │  field_type: u8  (0=string, 1=int64, 2=float64, 3=bool)  │  │
+│  │  indexed:    u8  (0=false, 1=true)                       │  │
+│  └──────────────────────────────────────────────────────────┘  │
+└────────────────────────────────────────────────────────────────┘
+```
+
+**Structure:**
+
+- `schema_version`: Monotonically increasing version number, incremented on metadata field changes
```
rodesai: this bumps based on user input right? its not the db's internal schema version?

PR #32, opened 2026-01-07
RFC for log segmentation

hachikuji:
This patch adds a logical segmentation notion which is similar to the way that time buckets work in timeseries. Segments can be created based on time-based boundaries, such as hourly like with timeseries. The user can also seal a segment based on their own application logic. This provides a useful foundation for cross-key operations. For example, when reading from a range of keys, we can use the segment to position reads across all keys in the range. It could also be used as a basis for performing cross-key retention semantics (e.g. drop all segments older than one day).

### DIFF
--- a/log/rfcs/0002-logical-segmentation.md
+++ b/log/rfcs/0002-logical-segmentation.md
@@ -186,7 +186,7 @@ impl Log {
 ```
 
 This would enable use cases such as:
-- Creating checkpoint boundaries aligned with application logic
+- Creating segment boundaries aligned with application logic
 - Attaching metadata accumulated during writes (checksums, record counts, correlation IDs)
 - Aligning segments with external events (e.g., end of business day)
 


log/rfcs/0002-logical-segmentation.md
```diff
@@ -0,0 +1,207 @@
...
+| version | type | key (TerminatedBytes) | sequence |
+```
+
+While this supports efficient scans for a specific key's entire log or a sequence number range, it lacks the ability to efficiently seek to a particular point based on criteria other than sequence number. Consider these use cases:
+
+1. **Tail queries**: "Read the last day of data" requires scanning from the beginning to find relevant entries, as there's no mapping from time to sequence number.
+
+2. **Time-bounded queries**: "Read entries from 2pm to 4pm yesterday" similarly requires a full scan.
+
+3. **Checkpoint recovery**: Restarting a consumer from a time-based checkpoint requires knowing which sequence numbers correspond to that time.
```
agavra: FWIW we have this index already in SlateDB: https://github.com/slatedb/slatedb/blob/main/rfcs/0012-sequence-tracker.md

log/rfcs/0002-logical-segmentation.md
```diff
@@ -0,0 +1,207 @@
...
+4. **Sparse in sequence space**: Segment boundaries don't need to align with sequence allocation blocks
+5. **Atomic batch placement**: Each write batch lands entirely within a single segment; batches never span segment boundaries
+
+### Key Encoding with Segments
+
+The log entry key format is extended to include the segment ID:
+
+```
+Log Entry:
+  | version (u8) | type (u8) | segment_id (u64 BE) | key (TerminatedBytes) | sequence (u64 BE) |
```
agavra: do we need a full u64 for segment id? u32 would allow 4B segments. adding this to every key seems pretty significant for a log structure. There are even varint encodings that maintain lexicographical ordering (see https://github.com/khonsulabs/ordered-varint for example, which we could easily vendor)

log/rfcs/0002-logical-segmentation.md
```diff
@@ -0,0 +1,207 @@
...
+
+Segments are "bumped" (a new segment is started) either automatically or manually.
+
+#### Time-Based Trigger
+
+The built-in trigger starts a new segment after a configurable wall-clock duration. Example: With a 1-hour interval, a new segment starts every hour. This provides predictable time-based partitioning similar to timeseries buckets.
+
+Time-based triggering is simple to implement because it only requires comparing the current wall-clock time against the segment's `start_time_ms`. No internal state tracking is needed beyond what's already stored in the segment metadata.
+
+#### Manual Trigger
```
agavra: I'm wondering if we should wait for a concrete use case for this before introducing it. The benefit of something predictable like time/size is that we can tune compaction to match. If segments can only be sealed manually it's a bit of a free for all for compaction.

log/rfcs/0002-logical-segmentation.md
```diff
@@ -0,0 +1,207 @@
...
+
+End boundaries (end sequence, end time) are derived from the next segment's start values, or from the current log state for the active segment.
+
+#### Metadata Lifecycle
+
+Segment metadata is written in two phases:
+
+1. **On open**: When a new segment is created, a `SegmentMeta` record is written with `start_seq`, `start_time_ms`, and empty `user_meta`. This ensures the segment is immediately discoverable.
+
+2. **On seal**: When `seal_segment()` is called with user metadata, the `SegmentMeta` record is overwritten to include the user-provided bytes. If no user metadata is provided, the record is left unchanged.
```
agavra: we probably need to track the current segment id and increment it somewhere, right?

log/rfcs/0002-logical-segmentation.md
```diff
@@ -0,0 +1,207 @@
...
+    /// via manual `seal_segment()` calls.
+    ///
+    /// Default: `None` (disabled)
+    auto_seal_interval: Option<Duration>,
+}
+```
+
+With the default configuration (`auto_seal_interval: None`), the log writes to segment 0 indefinitely. Users who want time-based partitioning can enable it by setting an interval. This keeps the default behavior simple while allowing opt-in to automatic segment management.
+
+## Alternatives
```
agavra, line 119: I'm curious if you've considered making the segment concept just an index instead of adding it to the data model. I haven't given it much thought, but because segment id and sequence numbers are co-monotonic I don't know if we actually need to encode them in each key, right?

log/rfcs/0002-logical-segmentation.md
agavra: alternatively, we can make sequence a u32 and make the total ordering segment + sequence so that we have u64 in total (sequences are unique within a segment)

hachikuji: I agree 8 bytes is a lot. Another idea I was considering is to add an attribute flag packed into the record type to specify the segment_id size. We could use 4 bytes by default, but switch to 8 bytes if needed. We probably wouldn't need to implement it until we need it. Just need to know it is possible.

log/rfcs/0002-logical-segmentation.md
hachikuji: I'm trying to picture what a global secondary index would look like in our system. All of the indexing in timeseries is oriented around time buckets (which are basically the same as segments as I've defined here). Even Kafka's indexing is at the segment level. I guess my intuition was that the segment would be the natural place to stick any secondary indexes. If we wanted a time index like Kafka, for example, we could stick it at the segment level. Otherwise, we'd have to figure out global index maintenance. Honestly not sure what problems lurk there. It probably complicates range query support? Rather than just reading the entire index for a given segment, we'd need to read thruogh each key individually. Sort of the same problem you have if you try to the logs corresponding to a prefix. The segment gives you a way to step forward across all keys at the same time. I guess my gut is that this is going to be a useful foundation for anything we want to build since it gives us an approach to handle ranges of keys rather than individual keys. We can drop segments for example rather than enforcing retention for each key separately. Similar to what we planned for timeseries.

log/rfcs/0002-logical-segmentation.md
hachikuji: Yeah, I'm good with that. It is more optimistic than foundational.

log/rfcs/0002-logical-segmentation.md
hachikuji: I do like the idea to consider segment + sequence as a composite. One thing that bothers me a little bit is that we would have to bump when the u32 sequence is exhausted. Perhaps that's not a big deal, but I did kind of like the idea of representing "semantic segments" as we have with timeseries (each bucket represents an hour of time). It would be annoying to have to handle overflow segments. Perhaps we could consider using varlength u64 for the sequence number which always resets to 0? So u32 segment_id and varlength u64 sequence. This option wasn't possible until we had the fancypants delimiter for the key.

agavra: image

> We could use 4 bytes by default, but switch to 8 bytes if needed. 
> So u32 segment_id and varlength u64 sequence

I like both of these ideas! I agree that dealing with overflow-triggering segment bumps is not fun.

log/rfcs/0002-logical-segmentation.md
agavra: yeah, the scoped secondary index to a segment makes a lot of sense to me. I think that's a good reason to keep this design pattern.

log/rfcs/0002-logical-segmentation.md
rodesai: Is the concern the storage overhead? I would imagine that gets optimized away pretty well by the prefix encoding in the SSTs

log/rfcs/0002-logical-segmentation.md
```diff
@@ -0,0 +1,207 @@
...
+- Within a key, entries are ordered by sequence number
+- Prefix scans can be constrained to specific segments
+
+### Segment Metadata
+
+Each segment has associated metadata stored in a separate record:
+
+```
+SegmentMeta Record:
+  Key:   | version (u8) | type (u8=0x03) | segment_id (u64 BE) |
```
rodesai: one thought: does it make sense to include the notion of a segment key _and_ a segment id, and use the id as the key for internally generated segments? The idea being that when we support user-generated segments they can select their own keys that can be looked up later

log/rfcs/0002-logical-segmentation.md
rodesai: I think we can just read the metadata in reverse order to find the last segment.

log/rfcs/0002-logical-segmentation.md
agavra: that + memory overhead, I guess prefix compression would basically eliminate storage overhead.

log/rfcs/0002-logical-segmentation.md
hachikuji: I kind of expected we would keep the segment metadata in memory. It's tiny.

log/rfcs/0002-logical-segmentation.md
hachikuji: I removed this point. It didn't make a lot of sense to me. I think it was saying mostly the same thing as the other points. Even at the level of individual key, there should be a way to read only the portion of the log that matches a given time range. In the current implementation, you just have to scan everything. Probably we'll need a time index, but I think segments are the foundation for that.

log/rfcs/0002-logical-segmentation.md
hachikuji: That's a nice idea. Do you think this could be added through a version bump later? Another interesting idea I was considering is to allow segments to be defined at prefix/range granularity. That might open the door to enforcing different retention semantics for different keys. I do think there might an interesting application design space here.

hachikuji:
> One high level thought is whether the notion of a segment is something we should just bake into slatedb directly. It feels very similar to what we'd want with twcs.

@rodesai  I agree that is the interesting question. I was wondering if the API in slatedb could work with minimal assumptions about key structure. For example, maybe we could have an API to mark a range of keys as "sealed" which makes those keys immutable and stops compaction? I think we need a design in slate sooner than later. Not sure this work needs to block behind it though?

log/rfcs/0002-logical-segmentation.md
hachikuji: I implemented something close to what we discussed. First, I switched to a u32 segment encoding, Second, I changed the sequence number to be a varlength u64 which is defined relative to the base sequence from the SegmentMetadata.

agavra: that's perfect since we'll get great prefix compression from the segment in slate but actually pretty bad compression from the sequence number so this may turn out to be better overall than the original design 👍

log/rfcs/0002-logical-segmentation.md
agavra: makes sense, I had a brain fart that we could just compute it from the latest SST last key on restart

log/rfcs/0002-logical-segmentation.md
agavra: I think these are neat ideas but I'm +1 to keeping things as simple as possible and only tinroducing those concepts if necessary later to avoid complicating things

rodesai:
> Not sure this work needs to block behind it though?

Totally agree. We can move forward with the approach here and revisit as we understand the problem better.

log/rfcs/0002-logical-segmentation.md
rodesai: Yep, sgtm

hachikuji:
If there are no objections, I'll merge this.

PR #33, opened 2026-01-07
Log read write impl

hachikuji:
This patch has implementations for the append/scan apis exposed by `Log`.

### DIFF
--- a/log/src/log.rs
+++ b/log/src/log.rs
@@ -7,6 +7,7 @@
 use std::ops::RangeBounds;
 use std::sync::Arc;
 
+use async_trait::async_trait;
 use bytes::Bytes;
 use opendata_common::storage::factory::create_storage;
 use opendata_common::{
@@ -301,6 +302,7 @@ impl Log {
     }
 }
 
+#[async_trait]
 impl LogRead for Log {
     async fn scan_with_options(
         &self,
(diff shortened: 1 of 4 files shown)


log/src/log.rs
```diff
@@ -46,7 +67,35 @@ impl LogIterator {
     ///
     /// Returns an error if there is a storage failure while reading entries.
     pub async fn next(&mut self) -> Result<Option<LogEntry>> {
-        todo!()
+        // Lazily initialize the storage iterator on first call
```
agavra: curious why not rename `new` -> `async open` and just load it there? (no preference, just curiosity)

opendata-common/src/storage/slate.rs
```diff
@@ -175,12 +175,24 @@ impl Storage for SlateDbStorage {
...
+
+    async fn put_with_options(
+        &self,
+        records: Vec<Record>,
+        options: WriteOptions,
+    ) -> StorageResult<()> {
         let mut batch = WriteBatch::new();
         for record in records {
             batch.put(record.key, record.value);
         }
```
agavra, line 190: fwiw, this is another reason why we couldn't use the sequence number from slate. a single AtomicWrite gets only a single seqnum but we need each individual record here to have a distinct monotonically increasing one

log/src/log.rs
hachikuji: Sounds good to me.

PR #34, opened 2026-01-08
[vector] rfc for the write API

agavra:
see second commit only as this is stacked behind #29

### DIFF
--- a/vector/rfcs/0002-write-api.md
+++ b/vector/rfcs/0002-write-api.md
@@ -52,9 +52,16 @@ A vector is the primary unit of ingestion, combining an ID, embedding values, an
 /// # Identity
 ///
 /// A vector is uniquely identified by its `id` within a namespace. The ID is
-/// a user-provided u32 that must be unique. Writing to an existing ID or a
-/// previously deleted ID is undefined behavior; callers must ensure IDs are
-/// never reused.
+/// a user-provided string (max 64 bytes UTF-8) that serves as an external
+/// identifier. The system internally maps external IDs to compact u64 internal
+/// IDs for efficient storage and indexing.
+///
+/// # Upsert Semantics
+///
+/// Writing a vector with an existing ID replaces the previous vector. The old
+/// vector is marked as deleted and a new internal ID is allocated. This ensures
+/// posting lists and metadata indexes are updated correctly without expensive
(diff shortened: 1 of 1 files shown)


vector/rfcs/0002-write-api.md
```diff
@@ -0,0 +1,513 @@
...
+///
+/// A vector is uniquely identified by its `id` within a namespace. The ID is
+/// a user-provided u32 that must be unique. Writing to an existing ID or a
+/// previously deleted ID is undefined behavior; callers must ensure IDs are
+/// never reused.
+///
+/// # Embedding Values
+///
+/// The `values` field contains the embedding vector as f32 values. The length
+/// must match the `dimensions` specified in the `Config` when the database
```
apurvam, line 69: Not sure what the convention is for Rust, but is it typical not to specific the return value for failure? In this case, if the input dimensions don't match the configured dimensions, Is it assumed to throw a standard error or something?

agavra: The failure will happen when you call `pub async fn write(&self, vectors: Vec<Vector>) -> Result<()>;` on `VectorDb`. The `Result` will be an error if it doesn't match the DB's configured dimensions.

vector/rfcs/0002-write-api.md
```diff
@@ -0,0 +1,513 @@
...
+///
+///     db.write(vectors).await?;
+///     Ok(())
+/// }
+/// ```
+pub struct VectorDb {
+    // Internal storage - not exposed
+}
+
+impl VectorDb {
```
hachikuji, line 181: Should we use the "Db" suffix for TimeSeries and Log as well?

agavra: ah, um I can go either way on this. i'm good removing `Db` here to call it just `Vector`

agavra: ah actually it's nice having the `Db` suffix for this one because there will be an API component called `Vector` to represent just a single vector.

hachikuji: Yeah, I think I prefer the Db suffix. Log could mean a lot of things. LogDb is clearer.

agavra: #57

PR #44, opened 2026-01-11
some perf fixes for ingest

rodesai:
## Summary

Adds some perf fixes for ingestion
- apply all samples to a given bucket in-batch within a single request. Amortizes costs from applying individual deltas to the current bucket (previously we were creating a new delta, and entering the critical section on every sample)
- configure some read-ahead for scans to improve bucket loading times. Currently we just use a fixed 1MB read-ahead. This is perhaps not appropriate for some scan scenarios. We can do something smarter later. Takes load time from minutes to millis
- adds a metric for samples ingested over remote-write

## Test Plan

Tested with victoriametrics prometheus benchmark. After this patch (and an additional patch to add backpressure: #45), I was able to ingest ~150K samples/second on an r5.xlarge (4 cores)

## Checklist

- [x] Tests added/updated

### DIFF
--- a/common/src/storage/slate.rs
+++ b/common/src/storage/slate.rs
@@ -1,15 +1,15 @@
 use std::sync::Arc;
 
-use async_trait::async_trait;
-use bytes::Bytes;
-use slatedb::{
-    Db, DbIterator, DbSnapshot, MergeOperator as SlateDbMergeOperator, MergeOperatorError,
-    WriteBatch, config::WriteOptions as SlateDbWriteOptions,
-};
-
 use crate::{
     BytesRange, Record, StorageError, StorageIterator, StorageRead, StorageResult,
     storage::{MergeOperator, RecordOp, Storage, StorageSnapshot, WriteOptions},
+};
+use async_trait::async_trait;
+use bytes::Bytes;
+use slatedb::config::ScanOptions;
+use slatedb::{
+    Db, DbIterator, DbSnapshot, MergeOperator as SlateDbMergeOperator, MergeOperatorError,
+    WriteBatch, config::WriteOptions as SlateDbWriteOptions,
 };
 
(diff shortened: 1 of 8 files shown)


common/src/storage/slate.rs
```diff
@@ -96,7 +96,16 @@ impl StorageRead for SlateDbStorage {
     ) -> StorageResult<Box<dyn StorageIterator + Send + 'static>> {
         let iter = self
             .db
-            .scan(range)
+            .scan_with_options(
+                range,
+                &ScanOptions {
+                    durability_filter: Default::default(),
+                    dirty: false,
+                    read_ahead_bytes: 1024 * 1024,
```
agavra, line 104: nice, so this was the magic? feels like we should document this somewhere in big and bold in slateDB, I didn't even know this eixsted

timeseries/src/tsdb.rs
```diff
@@ -142,12 +160,54 @@ impl Tsdb {
...
+                bucket_series_map
+                    .entry(bucket)
+                    .or_default()
+                    .push(bucket_series);
             }
         }
 
+        let buckets_touched = bucket_series_map.len();
+
+        // Second pass: ingest all series for each bucket in a single batch
```
agavra, line 172: nice catch, this makes sense to me. also 🤦 I introduced this regression in #24 -- that's probably why when I had run it I saw decent performance numbers when I tested it last. Probably makes sense to have some kind of CI/CD that runs an end-to-end test

common/src/storage/slate.rs
rodesai: I suspect this is why read performance was poor. For ingest I as just seeing lots of timeouts of the first request to ingest since it takes ages to load the minitsdb. But after that point ingest is not impacted.

PR #45, opened 2026-01-11
add a crude backpressure mechanism

rodesai:
## Summary

Adds a crude backpressure mechanism that forces tasks that are ingesting into a minitsdb to block if the current pending delta has not been flushed for 2xflush_interval. This should prevent the system from buffering samples indefinitely if ingestion is faster than flushing. This is just a stopgap. Longer-term we should rework ingestion to use a mechanism that lets the user configure a memory limit for the buffer.

## Test Plan

Ran with victoriametrics-prometheus-benchmark.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/minitsdb.rs
+++ b/timeseries/src/minitsdb.rs
@@ -103,6 +103,10 @@ pub(crate) struct MiniTsdb {
     next_series_id: AtomicU32,
     /// Pending delta accumulating ingested data not yet flushed to storage.
     pending_delta: Mutex<TsdbDelta>,
+    /// Receiver for waiting for updates to pending delta
+    pending_delta_watch_rx: tokio::sync::watch::Receiver<std::time::Instant>,
+    /// Sender for updating the pending delta when flushing.
+    pending_delta_watch_tx: tokio::sync::watch::Sender<std::time::Instant>,
     /// Storage snapshot used for queries.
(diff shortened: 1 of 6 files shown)


agavra (approved):
skimmed and it makes sense to me, let's open an issue for the improved backpressure mechanism. 

cc @apurvam might be interested because this would relate to how much we "require" using the log as the ingest mechanism, which would solve this naturally

rodesai:
> skimmed and it makes sense to me, let's open an issue for the improved backpressure mechanism.

Yep will do. I think there's a couple other things to clean up in the ingestion path wrt concurrency. I'm trying to summarize it into a chain of issues.

> cc @apurvam might be interested because this would relate to how much we "require" using the log

Not sure I follow. How will this solve the problem? We can log the records. From there we either need to
(1) consume the log for ingest, which solves the problem of using up too much memory in a buffer, but we still need some backpressure mechanism to prevent the consumer from lagging indefinitely. 
(2) just log for durability and still buffer in-memory. This seems better for perf, but we again need a backpressure mechanism.

agavra:
ah I just took "lagging indefinitely" is acceptable but you're totally right that's not appropriate. FWIW, if this is setup as a p8s-scraper then there is a natural backoff mechanism (we just don't have to issue a scrape)

PR #48, opened 2026-01-12
[log] First implementation of log segments

hachikuji:
This patch implements [RFC 2](https://github.com/opendata-oss/opendata/blob/main/log/rfcs/0002-logical-segmentation.md).


hachikuji:
Just one note. I did not do the varlength sequence number encoding yet from the RFC. We need to spell out the precise encoding scheme and make sure that it is compatible with lexicographic ordering. I will do this separately.

PR #49, opened 2026-01-12
fix: ensure query eval is scoped to right bucket (step 1)

rodesai:
## Summary

Before this patch, query evaluation was not scoped to a bucket. The evaluator would just take all the buckets and then merge the contained indexes. However each bucket uses its own series IDs, which results in garbage query results.

With this patch, we take the first step towards ensuring evaluation runs bucket-by-bucket:
- Renames QueryReader to BucketQueryReader to indicate that it reads within a single bucket. MiniQueryReader implements this trait.
- Adds a new trait QueryReader that has the same fns as BucketQueryReader but each fn also specifies the bucket within which the read occurs.
- Changes the query evaluation code to first validate that a given query only targets 1 bucket (given the start/end times and lookback), and then evaluates within that bucket.

I've also added a test that reproduces the problem and disabled it as it's currently failing (since we don't support cross-bucket queries in this patch).

In subsequent patches, we'll extend the various querying endpoints to support cross-bucket querying.

Also has a couple misc tweaks/cleanups:
- Changes TimeBucket to implement Copy since it just stores a couple ints
- Drops the lifetime parameter from MiniQueryReader and just has it store an owned TimeBucket.
- Changes the lifetime parameter from the returned indexes from BucketQueryReader to a static lifetime
- Note a couple TODOs of things to fix. I'll turn these into github issues

## Test Plan

Unit/integration tests inline

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -64,7 +64,12 @@ impl<'a, R: QueryReader> Evaluator<'a, R> {
     }
 
     pub(crate) async fn evaluate(&self, stmt: EvalStmt) -> EvalResult<Vec<EvalSample>> {
-        assert_eq!(stmt.start, stmt.end);
+        if stmt.start != stmt.end {
+            return Err(EvaluationError::InternalError(format!(
+                "evaluation must always be done at an instant.got start({:?}), end({:?})",
+                stmt.start, stmt.end
+            )));
+        }
         let result = self
             .evaluate_expr(
                 &stmt.expr,


timeseries/src/promql/evaluator.rs
```diff
@@ -64,6 +64,7 @@ impl<'a, R: QueryReader> Evaluator<'a, R> {
     }
 
     pub(crate) async fn evaluate(&self, stmt: EvalStmt) -> EvalResult<Vec<EvalSample>> {
+        assert_eq!(stmt.start, stmt.end);
```
agavra: did you intend to check this in?

timeseries/src/promql/tsdb_router.rs
```diff
@@ -359,6 +402,47 @@ impl PromqlRouter for Tsdb {
             }
         };
 
+        // Validate that query touches only one bucket
```
agavra, line 405: nit, can this be extracted (3x) into `get_single_bucket`? or are you just going to immediately follow this up with the PR that supports multi-bucket, then we can just leave it as is

timeseries/src/tsdb.rs
```diff
@@ -217,137 +221,86 @@ impl Tsdb {
...
-            for (sid, spec) in index.all_series() {
-                merged.series.insert(sid, spec);
-            }
-        }
-
-        Ok(Box::new(merged))
+    ) -> Result<Box<dyn ForwardIndexLookup + Send + Sync + 'static>> {
+        let mini = self.mini_readers.get(bucket).ok_or_else(|| {
+            crate::error::Error::Internal(format!("Bucket {:?} not found", bucket))
+        })?;
```
agavra, line 252: should this just return an empty forward index? (i.e. trying to find data that's expired isn't technically wrong, it'll just return empty) ditto for all the other methods

agavra (approved):
Thanks @rodesai - apologize for leaving this mess behind during the port

timeseries/src/promql/evaluator.rs
rodesai: yeah, we should never see an `EvalStmt` that has different start/end times. Evaluations are always done at a specific instant.

timeseries/src/promql/tsdb_router.rs
rodesai: I'm getting rid of all these in follow-ups

timeseries/src/tsdb.rs
rodesai: The caller should still never pass in the bucket here. It's supposed to discover the set of available buckets using the `list_buckets` fn and pass one of those in here.

timeseries/src/promql/evaluator.rs
rodesai: I turned it into an error

PR #50, opened 2026-01-13
support cross-bucket query evaluation

rodesai:
## Summary

Support executing promql queries that touch multiple buckets by iterating over the buckets in reverse-time order (newest first).

Also enables the cross-bucket tsdb eval tests.

Also extends the mock query reader to support multiple buckets

## Test Plan

Unit/integration tests

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -162,68 +162,70 @@ impl<'a, R: QueryReader> Evaluator<'a, R> {
             .as_millis() as i64;
         let start_ms = end_ms - (lookback_delta.as_millis() as i64);
 
-        // Use the selector module to find matching series
-        // Get the bucket to use for queries (for single-bucket operations, this will be the one bucket)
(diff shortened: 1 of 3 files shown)


timeseries/src/promql/evaluator.rs
```diff
@@ -162,68 +162,70 @@ impl<'a, R: QueryReader> Evaluator<'a, R> {
...
+                            series_id, bucket
+                        )));
+                    }
+                };
+                let fingerprint = self.compute_fingerprint(&series_spec.labels);
+
+                // Skip if we already found a sample for this series in a newer bucket
+                if series_with_results.contains(&fingerprint) {
+                    continue;
+                }
```
agavra, line 207: checking my understanding - we don't need to lookup the fingerprint in the older bucket forward idx if we found it in a newer one. the series is the same even if the id is different

rodesai: yep!

PR #51, opened 2026-01-13
[log] Varlength sequence numbers relative to segment start

hachikuji:
I implemented a varlength encoding for u32 and u64 which preserves numeric ordering in the encoded bytes. We use this in the log implementation to encode the sequence number in the suffix.

### DIFF
--- a/common/src/serde/varint.rs
+++ b/common/src/serde/varint.rs
@@ -100,7 +100,10 @@ pub mod var_u32 {
         let first_data_bits = ((value as u64) >> shift) as u8 & FIRST_BYTE_DATA_MASK;
         bytes[0] = (len_code << FIRST_BYTE_DATA_BITS) | first_data_bits;
 
-        // Remaining bytes in big-endian order
+        // Remaining bytes in big-endian order.
+        // Using indexed loop instead of iter_mut().enumerate() for better performance;
+        // the compiler optimizes bounds checks better with direct indexing here.
+        #[allow(clippy::needless_range_loop)]
         for i in 1..total_bytes {
             bytes[i] = ((value >> (8 * (total_bytes - 1 - i))) & 0xFF) as u8;
         }
@@ -340,7 +343,10 @@ pub mod var_u64 {
         };
         bytes[0] = (len_code << 4) | first_data_bits;
 
-        // Remaining bytes in big-endian order
+        // Remaining bytes in big-endian order.
+        // Using indexed loop instead of iter_mut().enumerate() for better performance;
+        // the compiler optimizes bounds checks better with direct indexing here.
+        #[allow(clippy::needless_range_loop)]
         for i in 1..total_bytes {
             bytes[i] = ((value >> (8 * (total_bytes - 1 - i))) & 0xFF) as u8;
         }


common/src/serde/varint.rs
```diff
@@ -0,0 +1,544 @@
+//! Variable-length integer serialization for lexicographically ordered keys.
+//!
+//! This module provides encoding and decoding for unsigned integers using a compact
+//! variable-length format that preserves lexicographic ordering.
+//!
+//! # Encoding scheme
+//!
+//! The first byte contains a length code in its high bits, with the remaining bits
```
agavra, line 8: to make sure I understand why this works, the length bits are lexicographically sorted "themselves", in other words `0b001` comes before `0b010` which comes before `0b011`, and the value ranges that those bits correspond to all come before each other respectively? thats a bunch of word vomit but I think I get it. it's quite simple!

agavra:
Actually now I'm wondering about the perf of decoding such variable length ints vs. the byte savings from having them 🤔 might be worth running some tests

hachikuji:
@agavra Yeah, I'll do a microbenchmark to compare. Usually the cost of varlength (just a few bit shifts) is negligible.

hachikuji:
@agavra I added a bench in the latest patch. I ran the tests locally. Here is an AI-generated summary from the bench results for u32 (similar story for u64):

**Encoding (serialize)**

Fixed-length encoding is basically constant at ~4.0 ns/value across all inputs.

Varint encoding scales with how many bytes the value needs:

- 1 byte values (0x0, 0x1f): ~3.2 ns → ~19% faster than fixed (because it writes less)
- 2 byte values (0x20, 0x1fff): ~4.78 ns → ~20% slower
- 3 byte values (0x2000, 0x1fffff): ~7.90 ns → ~2.0× slower
- 4 byte values (0x200000, 0x1fffffff): ~10.03 ns → ~2.5× slower
- 5 byte values (0x20000000, 0xffffffff): ~12.42 ns → ~3.1× slower

So: varint wins only when most of your values fit in 1 byte. Past that, fixed-length is faster and the gap grows with length.

**Decoding (deserialize)**

Fixed-length decode sits around ~1.5–1.7 ns/value.

Varint decode is always slower in your run:

- 1 byte values: ~2.23 ns → ~1.45–1.49× slower
- 2 byte values: ~2.40 ns → ~1.59–1.64× slower
- 3 byte values: ~2.45–2.51 ns → ~1.44–1.68× slower
- 4 byte values: ~2.92 ns → ~1.92–1.97× slower
- 5 byte values: ~2.65 ns → ~1.65–1.78× slower

So: decode pays the “branchy / loop / continuation-bit” tax even when the value is small.


My take is that all this is in line with expectations. The penalty for varlength is likely a rounding error in comparison with much heavier IO costs.

agavra:
Makes sense to me, the absolute numbers are totally within reason for me 👍

I do wonder what the effect would be when deserializing blocks of data (instead of a microbench that deserializes just one value). Branch prediction etc.. might actually come into affect. We might actually get into situations where serde is a bottleneck (it always was in Kafka Streams) but I suspect this will be minor compared to the serde the _user_ supplies. We can cross that bride if/when we get there

hachikuji:
One thing that stands out to me is that the penalty is mostly payed on encoding. Decode even in the worst case is less than 2x. For u64 in my local run, decode remained within 1.2x of the fixed length decode.

agavra:
Oh that is pretty interesting I didn't make that connection the first pass, I feel like decode is more important than encode for some reason. Anyway, I think saving space & I/O is certainly the good default option. LGTM

hachikuji:
@agavra I pushed a minor change in the encoding logic to write the results into an array first and then use `put_slice`. This narrows the margin considerably for encoding. For 5-byte payloads (the worst case for u32), we drop from about 12.6ns to 5.4ns per operation, which is closer to a 1.3x penalty over the fixed-length encoding.

PR #52, opened 2026-01-13
Fix lex ordering terminated bytes

hachikuji:
This patch fixes an issue in the terminated_bytes encoding. Since we encode `0xFF` as `0x01 0x03`, our encoding does not preserve the original lexicographic ordering of byte arrays. Here's a simple example: `0xFF > 0x07`, but `0x01 0x03 < 0x07`. We may or may not rely on this, but it makes things simpler if we have it and opens some doors. For example, with #51, we have a varlength encoding of integers which preserves lexicographic ordering. We could pass the output through `terminated_bytes` and then we can use varlength integer prefixes in keys. Then we could change the fixed-length u32 `segment_id` in the log key to a varlength u64 without paying all of its hefty price.

To fix the problem, we remove the escaping of `0xFF` and instead use lexicographic increment to compute the end bound of a prefix. This is also used in BytesRange. Finally, I've renamed `common::util` to `common::bytes` to emphasize focus on bytes utilities.

### DIFF
--- a/common/src/bytes.rs
+++ b/common/src/bytes.rs
@@ -59,8 +59,6 @@ impl BytesRange {
     }
 
     /// Creates a range that includes all keys with the given prefix.
-    ///
-    /// Uses [`lex_increment`] to compute the exclusive upper bound.
     pub fn prefix(prefix: Bytes) -> Self {
         if prefix.is_empty() {
             Self::unbounded()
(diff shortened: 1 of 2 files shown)


agavra:
I'm going to merge this so I can use it in the vector storage line of work. Thanks @hachikuji

PR #54, opened 2026-01-13
support cross bucket meta requests (series, labels, label values)

rodesai:
## Summary

Adds support for requests to the `series`, `labels` and `label/<label>/values` endpoints that query across multiple time buckets.

Builds on #50 so review that first.

## Test Plan

Unit/integration tests

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/tsdb_router.rs
+++ b/timeseries/src/promql/tsdb_router.rs
@@ -542,11 +542,9 @@ impl PromqlRouter for Tsdb {
     }
 
     async fn label_values(&self, request: LabelValuesRequest) -> LabelValuesResponse {
-        // Calculate time range (use defaults if not provided)
         let start_secs = request.start.unwrap_or(0);
         let end_secs = request.end.unwrap_or(i64::MAX);
 
-        // Get query reader for time range
         let reader = match self.query_reader(start_secs, end_secs).await {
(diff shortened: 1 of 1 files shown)


timeseries/src/promql/tsdb_router.rs
```diff
@@ -332,45 +335,77 @@ impl PromqlRouter for Tsdb {
...
-        // Extract label sets from each series
-        let mut result: Vec<HashMap<String, String>> = series_ids_vec
-            .iter()
-            .filter_map(|id| forward_index.get_spec(id))
-            .map(|spec| labels_to_map(&spec.labels))
-            .collect();
+        let mut result: Vec<HashMap<String, String>> = unique_series.into_values().collect();
+
+        // Sort for consistent output (by metric name first, then other labels)
+        result.sort_by(|a, b| {
```
agavra, line 395: feels like we're sorting more than we need to 🤔 anyway not a big deal and I can't immediately figure out a way to do less sorting, probably not a big deal

agavra (approved):
LGTM

timeseries/src/promql/tsdb_router.rs
rodesai: yeah it's very low on the list of bottlenecks right now 😄

PR #55, opened 2026-01-13
Move SeqBlock to common crate for reusable ID allocation

### LINKED ISSUE #42: move SeqBlock monotonic number generation to common
The sequence number generating algorithm used in the log (see #18) is useful in any context where we need to generate incrementing ids. can definitely be used for the vector db and potentially for the timeseries database as well.

Part of this ticket is figuring out how we want to model shared key schemas between databases. Another example is the record version/tag which all of the systems share a similar record type and currently duplicate into their own code.

agavra:
Extract the block-based sequence allocation mechanism from the log crate into common, making it a shared primitive for any OpenData system that needs monotonically increasing IDs (Log, Vector, etc.). I also added an RFC for it mostly for documentation of the mechanism so people don't have to dig it up within the Log RFC.

This should just have moved a bunch of code around, so I skimmed it to make sure that is what it did and that nothing was lost in translation. I did not review it super deeply and all the tests pass.

Fixes #42

### DIFF
--- a/common/src/sequence.rs
+++ b/common/src/sequence.rs
@@ -53,38 +53,38 @@ pub const DEFAULT_BLOCK_SIZE: u64 = 4096;
 
 /// Error type for sequence allocation operations.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub enum SeqBlockError {
+pub enum SequenceError {
     /// Storage operation failed
     Storage(StorageError),
     /// Deserialization failed
     Deserialize(DeserializeError),
 }
 
-impl std::error::Error for SeqBlockError {}
+impl std::error::Error for SequenceError {}
 
-impl std::fmt::Display for SeqBlockError {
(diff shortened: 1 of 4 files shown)


common/src/sequence.rs
```diff
@@ -0,0 +1,563 @@
...
+use crate::serde::DeserializeError;
+use crate::serde::seq_block::SeqBlock;
+use crate::{Record, Storage, StorageError};
+
+/// Default block size for sequence allocation.
+pub const DEFAULT_BLOCK_SIZE: u64 = 4096;
+
+/// Error type for sequence allocation operations.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SeqBlockError {
```
hachikuji: Wonder if we could generalize this slightly. Maybe SequenceError? We'll probably add more sequence utilities over time.

hachikuji (approved):
Makes sense to me.

PR #58, opened 2026-01-13
[common] Factor out common prefix utility

hachikuji:
Implements common [RFC-0001](https://github.com/opendata-oss/opendata/blob/main/rfcs/0001-record-key-prefix.md) as a small set of utilities which follow the prefix spec.


agavra (approved):
LGTM, I already merged this for vector but I can submit a separate PR to refactor that unless it's easy enough for you to do

hachikuji:
@agavra Yeah, it's easy to do. I'll just add it here.

PR #61, opened 2026-01-13
[log] Initial draft of listing api

hachikuji:
We should have an API to list keys with active logs. The approach here stores listing records within each segment that can be scanned independently from the log entries.

### DIFF
--- a/log/rfcs/0003-key-listing.md
+++ b/log/rfcs/0003-key-listing.md
@@ -65,7 +65,7 @@ On append(key, value):
 
 When a new segment starts, the cache is cleared.
 
-This approach avoids read-before-write overhead in the ingest path. When writers change (e.g., after failover), the new writer starts with an empty cache and may re-insert listing records for keys already present in the segment. However, since SlateDB overwrites duplicate keys, no additional storage overhead results from this.
+This approach avoids read-before-write overhead in the ingest path. When writers change (e.g., after failover), the new writer starts with an empty cache and may re-insert listing records for keys already present in the segment. SlateDB will overwrite duplicate keys, though the duplicates exist until compaction removes them. Since writers should not change frequently, we do not expect this to be a significant concern. If it does become problematic, we can introduce read-before-write in the future.
 
 ### List API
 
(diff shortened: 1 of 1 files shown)


log/rfcs/0003-key-listing.md
```diff
@@ -0,0 +1,149 @@
...
+ListingEntry Record:
+  Key:   | version (u8) | type (u8=0x04) | segment_id (u32 BE) | key (Bytes) |
+  Value: | (empty) |
+```
+
+The key structure places `segment_id` before the user key, ensuring that all listing records for a segment are contiguous. This enables efficient enumeration of keys within a segment via prefix scan.
+
+Unlike log entry keys, the user key is stored as raw `Bytes` without `TerminatedBytes` encoding. Since the key occupies the suffix position, no delimiter is needed to establish boundaries.
+
+The value is empty—presence of the record indicates the key exists in that segment.
```
agavra, line 52: does slate support empty values? I wonder if we could use this for other metadata eventually as well so we can just have a reserved byte

log/rfcs/0003-key-listing.md
```diff
@@ -0,0 +1,149 @@
...
+On append(key, value):
+  1. If key not in segment_key_cache:
+     a. Write ListingEntry record for (current_segment_id, key)
+     b. Add key to segment_key_cache
+  2. Write log entry as usual
+```
+
+When a new segment starts, the cache is cleared.
+
+This approach avoids read-before-write overhead in the ingest path. When writers change (e.g., after failover), the new writer starts with an empty cache and may re-insert listing records for keys already present in the segment. However, since SlateDB overwrites duplicate keys, no additional storage overhead results from this.
```
agavra: > no additional storage overhead results from this.

nit: there will be from garbage, so there's some space amplifciation but it gets cleared up on compaction

log/rfcs/0003-key-listing.md
```diff
@@ -0,0 +1,149 @@
...
+
+Maintaining a single global index of all keys (outside the segment structure) would simplify queries but complicates retention. When segments are deleted, identifying which keys are no longer present requires scanning the remaining segments. Per-segment listings naturally handle this—deleting a segment removes its listing records, and keys fall out of scope when their last segment is removed.
+
+## Open Questions
+
+### Deduplication Strategy
+
+The deduplication mechanism has two main approaches:
+
+1. **In-memory collection**: Accumulate all keys in memory, deduplicate, then return. Simple but memory usage scales with distinct key count. Enables sorted iteration order.
```
agavra: if we build up a prefix trie we can probably dramatically reduce the memory it takes to store the full keyset, I'm in favor of avoiding streaming unless necessary

hachikuji: Yeah, I agree. If we didn't need to deduplicate, I'd be tempted to stream, but if we need the memory overhead anyway, might as well make it synchronous.

log/rfcs/0003-key-listing.md
criccomini: Yea now that we have a deletion bit flag I think empty values are fine. Probably worth actually trying it, though. 😛

log/rfcs/0003-key-listing.md
```diff
@@ -0,0 +1,139 @@
...
+    fn list(&self, seq_range: impl RangeBounds<u64>) -> LogKeyIterator;
+    fn list_with_options(
+        &self,
+        seq_range: impl RangeBounds<u64>,
+        options: ListOptions,
+    ) -> LogKeyIterator;
+}
+```
+
+The sequence number range is mapped internally to the corresponding segment range. The iterator scans `ListingEntry` records across those segments and returns distinct keys.
```
rodesai, line 106: If listing takes a seq range, but that seq range doesn't align perfectly with the underlying segments, then wouldn't the listing then potentially return keys that are not present in the range?

PR #62, opened 2026-01-13
implement ingestion path for vector db

agavra:
**NOTE**: I attempted to break this up into multiple PRs but that just caused a lot of ??? in the actual implementation. Instead I just broke them up into different commits on this PR which you can review one by one if you want to though you may see a little bit of jitter and review some things that didn't end up making it in the end.

`storage/*.rs` and `model.rs` are pretty straightforward (I guess I could split those out into their own PRs if y'all want to review this deeply).

The complicated stuff is in `db.rs` where the ingest path mostly follows the timeseries ingest path:
```
1. buffer pending vectors into a delta
2. when it comes time to flush, mark the current delta "immutable" and allocate a new one
3. flush the delta (can still accept incoming writes into the new delta) this will:
  - assign the internal IDs for each external ids
  - turn the list of vectors into the corresponding index updates (and deletions if there were old ids assigned)
  - flush to slateDB
```

The backpressure mechanism is the same.

### DIFF
--- a/vector/src/storage/merge_operator.rs
+++ b/vector/src/storage/merge_operator.rs
@@ -26,8 +26,8 @@ impl common::storage::MergeOperator for VectorDbMergeOperator {
             return new_value;
         };
 
-        let prefix = KeyPrefix::from_bytes_versioned(key, KEY_VERSION)
-            .expect("Failed to decode key prefix");
+        let prefix =
+            KeyPrefix::from_bytes_versioned(key, KEY_VERSION).expect("Failed to decode key prefix");
 
         let record_tag = prefix.tag();
 


vector/src/storage/merge_operator.rs
```diff
@@ -0,0 +1,296 @@
...
+
+impl common::storage::MergeOperator for VectorDbMergeOperator {
+    fn merge(&self, key: &Bytes, existing_value: Option<Bytes>, new_value: Bytes) -> Bytes {
+        // If no existing value, just return the new value
+        let Some(existing) = existing_value else {
+            return new_value;
+        };
+
+        // Decode record type from key
+        if key.len() < 2 {
```
hachikuji: I think we have a helper which extracts the prefix. `KeyPrefix::from_bytes_versioned`?

vector/src/model.rs
```diff
@@ -0,0 +1,314 @@
...
+    /// attribute names or type mismatches will fail. If empty, any attribute
+    /// names are accepted with types inferred from the first write.
+    pub metadata_fields: Vec<MetadataFieldSpec>,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            storage: Arc::new(common::storage::in_memory::InMemoryStorage::new()),
+            dimensions: 0, // Must be set explicitly
```
hachikuji, line 189: nit: does Default make sense if it is not complete?

agavra: I was debating that as well, I thought it was a weird claude-ism but decided to keep it because I think you can use this syntax
```rs
    let vector = Vector {
        dimensions: 1234,        // change just this field
        ..Default::default()
    };
```

vector/src/storage/merge_operator.rs
agavra: note that I need to `expect` anyway because SlateDB's merge operator doesn't support returning a `Result` (though maybe we should fix that)

PR #63, opened 2026-01-14
[log] Add README for log with rough roadmap

hachikuji:
Attempted to write an initial README for the log system.

### DIFF
--- a/log/README.md
+++ b/log/README.md
@@ -25,7 +25,7 @@ Each log entry contains a key, sequence, and value. Entries are stored as `(segm
 > [!IMPORTANT]
 > Compared to the Kafka data model, our "key" is closer to a topic partition. The key identifies the log stream; the value is the payload. We have no separate notion of keys which distinguish entries within the same log stream.
 
-Segments partition the sequence space and scope compaction—entries within a segment are sorted by key, then sequence. The `seal_interval` configuration controls segment boundaries: smaller intervals reduce write amplification at the cost of read locality.
+Segments partition the sequence space and scope compaction—entries within a segment are sorted by key, then sequence. The `seal_interval` configuration controls segment boundaries: smaller intervals reduce write amplification at the cost of read locality. Conceptually, each segment represents a window of time across all keys. In the future, we may support size-based segmentation as well.
 
 What that buys us:
 - **Many independent logs, zero provisioning** — each `key` is its own log stream; creating new streams is just writing new keys (no pre-allocation or reconfiguration)


log/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Concepts
+
+**Key** — Identifies an independent log stream. Variable-length bytes.
+
+**Sequence** — A global 64-bit counter assigned at write time. Within each key, sequences are guaranteed to increase monotonically, but they are not contiguous—gaps appear when writes to other keys are interleaved.
+
+**Log Entry** — A record read from the log, containing the key, sequence, and value.
+
+**Segment** — Logical partitions of the sequence space, created periodically. Used internally for retention and efficient seeking.
```
apurvam: Why should a user care about this?

hachikuji: Yeah, I was debating whether a user should care about this. It is exposed in the configuration, so some awareness probably makes sense. Let me try to revise.

hachikuji: @apurvam I tried to clarify the structure in the latest revision at a level that is useful to users.

log/README.md
```diff
@@ -0,0 +1,78 @@
...
+             └───────┴───────┴───────┴───────┴───────┘
+
+Segment 0    ┌───────┬───────┬───────┬───────┬───────┐
+(seq 0–4)    │ A:0   │ A:3   │ B:1   │ C:2   │ C:4   │
+             └───────┴───────┴───────┴───────┴───────┘
+
+Each cell: key:sequence
+```
+
+Each log entry contains a key, sequence, and value. Entries are stored as `(segment_id, key, sequence)`, where the sequence is a global counter that increases monotonically per key but is not contiguous.
```
agavra, line 23: probably worth calling out what we mean by `key` with analogy to topic since this is different from what most people expect a `key` is in a log (e.g. key/value pairs within a topic)

it's hinted below, but I think making it explicit maybe with a github callout:
```
> [!IMPORTANT]
> Key information users need to know to achieve their goal.
```

log/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+Segments partition the sequence space and scope compaction—entries within a segment are sorted by key, then sequence. The `seal_interval` controls segment boundaries: smaller intervals reduce write amplification at the cost of read locality.
+
+What that buys us:
+- **Many independent logs, zero provisioning** — each `key` is its own log stream; creating new streams is just writing new keys (no pre-allocation or reconfiguration)
+- **Automatic locality via compaction** — the LSM continuously rewrites data so entries for the same key become clustered on disk over time
+- **Simple mental model** — scaling isn’t “partition management”; it’s just “write and read by key”
+
+Trade-offs:
+- **Write amplification** — LSM compaction rewrites data, which increases total bytes written compared to pure append-only layouts.
```
agavra: might be worth calling out TWCS-style compaction to mitigate this if we're explicitly noting the tradeoffs

log/README.md
apurvam: Makes sense. I think an example like 'one logical partitioning could be wall clock time, so any given segment maps to a time range. This allows us to seek by time range, delete blocks of time, etc. The actual function for creating is pluggable'. 

this example makes it more concrete and may click better for some users. Your call in the end.

hachikuji: Yeah, the example would help.

PR #64, opened 2026-01-14
add a request-level cache to the evaluator

rodesai:
## Summary

This patch adds a request-level cache to the evaluator. When running query_range, the db iterates over each evaluation instant and re-executes query evaluation. Most of the time, it's repeating the same work since all evaluations within a time bucket read the same indexes and series rows from slatedb. This patch adds CachedQueryReader, which stores the results of these reads and uses them for future reads. Each new request creates its own CachedQueryReader when it creates the evaluator.

## Test Plan

- unit test
- ran perf benchmarks. Perf improved from ~30 timeseries/second to 180 timeseries/second

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -2,19 +2,22 @@ use std::collections::{HashMap, HashSet};
 use std::fmt::{Display, Formatter};
 use std::future::Future;
 use std::pin::Pin;
+use std::sync::Arc;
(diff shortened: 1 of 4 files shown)


rodesai:
> I may have missed it, but where is the cache size bounded and when are entries evicted?

It isn't currently bounded. It lives for the duration of a request and is then dropped.

PR #67, opened 2026-01-14
improve AGENTS.md setup to help bootstrapping new conversastions

agavra:
creates an AGENTS.md for each subproject and updates the main one to give some instructions to avoid context pollution

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,7 @@
 # AGENTS.md
 
+**Start here**: Read [CONTRIBUTING.md](CONTRIBUTING.md) for development workflow, RFC process, code style, and testing patterns.
+
 ## Project Overview
 
 OpenData is a collection of deployable database systems that share common infrastructure. Each database is its own crate but leverages shared logic from `common`. All databases are built on [SlateDB](https://github.com/slatedb/slatedb).


agavra:
Great idea. I'll add it!

AGENTS.md
```diff
@@ -1,14 +1,29 @@
 # AGENTS.md
 
+**Start here**: Read [CONTRIBUTING.md](CONTRIBUTING.md) for development workflow, RFC process, code style, and testing patterns.
+
 ## Project Overview
 
-OpenData is a collection of deployable database systems that share common infrastructure. Each database is its own crate but leverages shared logic from `common`.
+OpenData is a collection of deployable database systems that share common infrastructure. Each database is its own crate but leverages shared logic from `common`. All databases are built on [SlateDB](https://github.com/slatedb/slatedb).
+
+**Important**: When working on a specific database, focus on that crate's directory. Each database has its own `AGENTS.md` with architecture details. Avoid loading context from other database crates unless explicitly needed.
```
hachikuji, line 9: Would it be helpful to have a note about keeping AGENTS.md up to date with changes?

agavra: I was thinking we should just do that manually batched on occasion if we notice drift. otherwise I think we risk it adding too many minute features into AGENTS.md

hachikuji: Fair enough. Wonder if we could turn that into a periodic job.

agavra: I like that idea 👍 and then it can just create a PR for us to review

hachikuji (approved):
LGTM!

PR #72, opened 2026-01-15
add RFC for vector read API

agavra:
This change also modifies the write-API to make it clear that `vector` is a valid attribute type in order to unify the data model.

### DIFF
--- a/vector/rfcs/0003-read-api.md
+++ b/vector/rfcs/0003-read-api.md
@@ -774,27 +774,6 @@ This is useful for:
 - Pipelines that perform a coarse search first, then refine with additional filters
 - Queries where the vector embedding is not needed in results
 
-## Comparison with LogRead
-
-| Aspect                  | LogRead                          | VectorDbRead                       |
-|-------------------------|----------------------------------|------------------------------------|
-| Primary operation       | `scan(key, seq_range)`           | `search(query)`                    |
-| Query parameters        | Key + sequence range             | Vector + limit + filter            |
-| Result ordering         | Sequence number (ascending)      | Similarity score (best first)      |
-| Iterator support        | `LogIterator` for streaming      | Vectorized results (no iterator)   |
-| Options variant         | `scan_with_options()`            | Builder pattern on `Query`         |
-| Point lookup            | N/A (scan with range)            | `get(id)`                          |
-| Batch retrieval         | N/A                              | Not provided (can add if needed)   |
-| Filtering               | N/A                              | Metadata filters with predicates   |
-| Field selection         | N/A                              | `FieldSelection` enum              |
-| Count operation         | `count(key, seq_range)`          | Not provided (can add if needed)   |
-| Lazy initialization     | N/A                              | Centroids loaded on first query    |
-| Thread safety           | Methods take `&self`             | Methods take `&self`               |
-| Snapshot semantics      | Reader operates on snapshot      | Reader operates on snapshot        |
-
-Both follow the same architectural pattern of separating read-only concerns from read-write
-operations.
-
 ## Examples
 
 ### Basic Search


vector/rfcs/0002-write-api.md
```diff
@@ -118,15 +118,29 @@ impl Attribute {
 ///
 /// These types align with the metadata field types defined in the storage
 /// layer (CollectionMeta). Type mismatches at write time will return an error.
+///
+/// The Vector variant stores embedding vectors and enables treating the vector
```
apurvam, line 122: Why special case embedding vectors in the api? It looks like at the storage level these are not distinguished from regular vectors. So how are they used differently from a user perspective?

vector/rfcs/0003-read-api.md
```diff
@@ -0,0 +1,1072 @@
...
+The vector embedding is stored under the reserved field name `"vector"` in the fields HashMap,
+enabling a unified data model where all attributes (including the vector) can be queried and
+retrieved consistently. The Vector variant is first as it is the primary data in a vector
+database. At the storage layer, vectors are stored in VectorData records while other attributes
+are stored in VectorMeta records.
+
+### VectorDbRead Trait
+
+Following the `LogRead` pattern from the log module, we define a trait that can be implemented by
+both `VectorDb` and `VectorDbReader`:
```
apurvam, line 83: I think maybe the better approach to the `Writer`, `Reader`, `Compactor` roles that's proposed in RFC #74 is to just have traits with those roles in every DB. I don't think a higher level abstraction is needed, as these traits are pretty composable on their own IMO.

vector/rfcs/0003-read-api.md
```diff
@@ -0,0 +1,1072 @@
...
+    - Cosine/DotProduct: Exclude results where score < threshold
+
+### Filter Evaluation
+
+Filters are evaluated using inverted indexes stored in the `MetadataIndex`:
+
+- **Equality (`Eq`)**: Direct lookup in `MetadataIndex[field][value]` returns vector IDs
+- **Inequality (`Neq`)**: All vectors except those in `MetadataIndex[field][value]`
+- **Range (`Lt`, `Lte`, `Gt`, `Gte`)**:
+  - Scan metadata index keys in sorted order (using sortable encoding)
```
apurvam, line 710: Refreshing my understanding: the inverted index has `key=value` pairs sorted in lexicographic order, and these range lookups will use that sort order to pull the right range?

vector/rfcs/0003-read-api.md
```diff
@@ -0,0 +1,1072 @@
...
+- Skip loading VectorData entirely if "vector" is not requested (saves 4 bytes × dimensions)
+- Skip loading VectorMeta entirely if no metadata fields are requested
+- Reduce deserialization overhead for large records
+
+This is useful for:
+- Applications that only need IDs and scores (e.g., for ranking before a second pass)
+- Pipelines that perform a coarse search first, then refine with additional filters
+- Queries where the vector embedding is not needed in results
+
+## Comparison with LogRead
```
apurvam: Not sure why we need this comparison table here. Is it to show the similarity in the API design or something else?

apurvam (approved):
Left minor comments. But this looks good to me!

vector/rfcs/0002-write-api.md
agavra: these are one and the same, the idea is that instead of return:
```
Result {
 vector: Vec<f32>,
 fields: Map<String, AttributeValue>
}
```
we can return just
```
Result {
 fields: Map<String, AttributeValue>
}
```
and have a field named `"vector"`. This matches the Turbopuffer API and I like having it be unified like that instead of special casing vector.

Vectors are stored differently in the storage layer though because you always need to fetch the vector to compute similarities but you may not always need to fetch the metadata for every vector using the forward index.

vector/rfcs/0003-read-api.md
agavra: yeah that's the plan for string ranges, but it doesn't work for numerical ranges (often you won't index something like "price" in an inverted index because there are too many term values). for those we'll need to post-filter using the forward index or add a min-max index option.

vector/rfcs/0003-read-api.md
agavra: yeah it's probably not relevant here. I had the comparison table in the write API RFC which was useful to see the similarities between the two and Claude generated it here to match. I guess it goes to show the point that unifying the read APIs may not make much sense

vector/rfcs/0002-write-api.md
agavra: the benefit of this, by the way, is that you can store multiple vectors in a single document if you wanted to. so you _could_ (if we update storage to support this) have a main vector and then some related embedding stored in the forward index for it as well

apurvam: Makes sense. thanks!

Issue #47, opened 2026-01-12
Flexible and configurable deployment options


apurvam:
## Summary

Design and implement first-class APIs, tooling, and conventions that enable OpenData databases to seamlessly operate across all deployment topologies—embedded, single-node, distributed, cross-account, and hybrid—with native support for ownership transfer and database branching.

## Motivation

Object storage as the sole persistence layer is OpenData's core architectural lever. However, the full implications of this architecture as it relates to database deployment options haven't yet been sketched out. This issue is the first pass at sketching out the problem and the options.

**Problems with traditional databases that OpenData can solve:**

1. **Durability without replication**: In traditional databases, embedded instances sacrifice durability because they can't create replicas. With object storage, an embedded OpenData instance can be fully durable—data survives process termination. This enables new use cases: durable per-agent databases, crash-recoverable local caches, etc.

2. **Migration-free scaling path**: Traditional database consumption follows this progression: start with managed multi-tenant offering → migrate to your own account for compliance → migrate to self-hosted for cost. Each step requires moving data between disk sets. The problem grows with scale even as the necessity to migrate grows with scale. With OpenData, readers and writers can spin up anywhere—the data never moves. Thus it's possible for such "Migrations" to become topology changes, not data movements.

3. **Flexible reader/writer placement**: Since coordination happens via manifests in object storage, readers and writers can live in different accounts, different regions, or different trust boundaries.  For example, a managed writer can populate data that customer-owned readers consume the data directly for any use case without using the vendor's query layer or API.

**What's missing:**

OpenData doesn't yet expose these capabilities as coherent, well-documented primitives. We should sketch out how we want to take advantage of this flexibilty, build the right common components, and then have each database be designed to leverage these common components. 

### 1. Topology-Agnostic Database Design

Every OpenData database should work across the full deployment spectrum:

| Topology | Writers | Readers | Coordination |
|----------|---------|---------|--------------|
| Embedded | In-process | In-process | Local manifest |
| Single-node | Single process | Same process | Local manifest |
| Distributed | Multiple processes | Multiple processes | Shared manifest |
| Cross-account | Account A | Account B | Bucket sharing + shared manifest |
| Hybrid | Managed service | Customer infra | Cross-account manifest |

This requires clear separation between framework abstractions and database-specific logic.

#### Framework-Level Abstractions

The framework provides primitives that behave differently based on deployment topology, so database authors write to stable interfaces:

**Component Communication**

[...]

PR #74, opened 2026-01-16
Add service framework rfc

apurvam:
##  Summary

  Adds RFC 0003: Service Layer, which defines a framework for consuming OpenData databases (Log, TSDB, Vector) in various deployment configurations—from fully embedded to fully service-oriented.

  The RFC covers:
  - Role Abstractions: Writer, Reader, and Compactor traits that decompose databases into composable pieces
  - Composition Patterns: Five deployment modes (Embedded, Embedded + Background Compactor, Single-Server, Scaled Readers, Serverless)
  - ServiceRuntime: Shared Axum-based HTTP infrastructure with health checks and metrics
  - Protocol Bindings: Per-database HTTP endpoints (Log, PromQL, Vector)
  - Catalog Abstraction: Foundation for future sharding support
  - Configuration Model: Builder pattern + TOML/YAML file support

##  Related Issues

  Relates to #47 

##  Test Plan

  RFC document only—no code changes. Reviewed for:
  - Consistency with existing RFC format (log/rfcs/0001-storage.md)
  - Alignment with current codebase patterns (Axum server in timeseries/src/promql/server.rs, storage traits in common/src/storage/)
  - Technical accuracy regarding SlateDB capabilities

##  Checklist

[] Tests added/updated
[] cargo fmt and cargo clippy pass
[] Documentation updated (if applicable)

### DIFF
--- a/rfcs/0003-service-layer.md
+++ b/rfcs/0003-service-layer.md
@@ -1057,6 +1057,8 @@ impl Catalog for LocalCatalog {
 
 #### S3Catalog (Multi-Shard)
 
+> **Note**: The catalog implementation sketched here is directional. There are many approaches. For instance the catalog could be stored in a special config-kv store that's backed by SlateDB itself, this could use a single-writer/many-readers pattern, with reader instances embedded alongside database writers and readers to quickly resolve key → shard → storage config. This architecture would scale well and is worth exploring in a follow-up RFC.
+
 For distributed deployments, the catalog is stored in S3 alongside the data:
 
 ```


rfcs/0003-service-layer.md
```diff
@@ -0,0 +1,1491 @@
...
+    /// Execute a read query
+    async fn query(&self, request: Self::QueryRequest) -> Result<Self::QueryResponse, ReadError>;
+
+    /// Refresh view of underlying storage
+    async fn refresh(&self) -> Result<(), ReadError>;
+}
+
+/// Compactor role - optimizes storage layout
+#[async_trait]
+pub trait Compactor: DatabaseRole {
```
agavra, line 154: we might want to punt on this role for a little until we've designed a little bit more of the compaction layer. right now all of them are just using the default SlateDB compaction algorithms.

I don't imagine it'll be as simple as this.

cc @hachikuji whose working on TWCS-esque compaction for Slate and probably has thought about this for a while.

rfcs/0003-service-layer.md
```diff
@@ -0,0 +1,1491 @@
...
+
+impl EmbeddedLog {
+    pub async fn open(config: Config) -> Result<Self, Error> {
+        let storage = create_storage(&config.storage, None).await?;
+
+        Ok(Self {
+            writer: LogWriter::with_storage(Arc::clone(&storage)),
+            reader: LogReaderRole::with_storage(Arc::clone(&storage)),
+            compactor: LogCompactor::with_storage(storage),
+        })
```
agavra, line 385: looking at this makes me think that a writer should always be a reader as well. separating the abstractions makes it difficult to allow reading unflushed state (i.e. communicate over memory). otherwise the API only allows for communication via storage

rfcs/0003-service-layer.md
```diff
@@ -0,0 +1,1491 @@
...
+
+// Usage:
+let log = EmbeddedLog::open(config).await?;
+log.append(records).await?;
+let entries = log.scan(key, ..).collect().await?;
+```
+
+#### B. Embedded with Background Compactor
+
+Writer and Reader in application process; Compactor runs as background task.
```
agavra, line 406: how is this different from embedded? is the idea that with truly embedded you'd need to manually call the compactor?

rfcs/0003-service-layer.md
```diff
@@ -0,0 +1,1491 @@
...
+        }
+
+        router.with_state(state)
+    }
+
+    fn protocol_name() -> &'static str { "vector" }
+}
+```
+
+### Catalog Abstraction
```
agavra, line 947: probably worth deferring this to a future RFC specifically about sharding / distribution

rfcs/0003-service-layer.md
```diff
@@ -0,0 +1,1491 @@
...
+
+    #[serde(default)]
+    pub distance_metric: DistanceMetric,
+
+    #[serde(default = "default_flush_interval", with = "humantime_serde")]
+    pub flush_interval: Duration,
+}
+```
+
+#### Configuration Builder
```
agavra, line 1211: nit: for actual construction of the config we should use some off the shelf configuration libraries, there are some really good ones in the rust ecosystem

apurvam:
Thanks for the review @agavra . I agree with all of that. 

> I'm a little uncertain about the Reader/Writer/Compactor abstractions.

I was already thinking about backing this out. It seems like a superfluous abstraction as presented. I think there is value in splitting out the reader / writer / compactor ,but these abstractions as presented in this RFC really don't add anything meaningful in that direction IMO.

rfcs/0003-service-layer.md
apurvam: Not very different. The compactor object runs in a background thread. Probably redundant since slate's compactor is in the background anyway.

apurvam:
I'm going to close this in favor of focused RFCs for each common component.

PR #75, opened 2026-01-16
add range vector selectors and rate fn

rodesai:
## Summary

Add support for range vector selectors in queries, and support a basic implementation of the rate fn over range vectors

required to support the grafana metrics explorer panel

## Test Plan

Unit tests
Tested plotting a graph on grafana

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -44,6 +44,10 @@ impl From<crate::error::Error> for EvaluationError {
 
 pub(crate) type EvalResult<T> = std::result::Result<T, EvaluationError>;
 
+/// Type alias for complex HashMap used in matrix selector evaluation.
+/// Maps from label key (sorted vector of label pairs) to samples vector
+type SeriesMap = HashMap<Vec<Label>, Vec<Sample>>;
+
 pub(crate) struct QueryReaderBucketEvalCache {
     // Map from terms (series_ids for forward, labels for inverted) to cached results
     forward_index_cache:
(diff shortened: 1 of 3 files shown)


timeseries/src/promql/evaluator.rs
```diff
@@ -193,6 +197,12 @@ pub struct EvalSample {
     pub(crate) labels: HashMap<String, String>,
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub struct EvalSamples {
+    pub(crate) values: Vec<(i64, f64)>, // (timestamp_ms, value) pairs
```
agavra: should we use `Vec<Sample>`? we could also consider reusing `Series` and just not setting any of the optional fields

timeseries/src/promql/evaluator.rs
```diff
@@ -44,6 +44,10 @@ impl From<crate::error::Error> for EvaluationError {
 
 pub(crate) type EvalResult<T> = std::result::Result<T, EvaluationError>;
 
+/// Type alias for complex HashMap used in matrix selector evaluation.
+/// Maps from label key (sorted vector of label pairs) to (labels hashmap, samples vector).
+type SeriesMap = HashMap<Vec<(String, String)>, (HashMap<String, String>, Vec<(i64, f64)>)>;
```
agavra: should we use `Vec<Label>` for the key and `Vec<Sample>` in the value? ditto throughout the PR

timeseries/src/promql/evaluator.rs
rodesai: Good call. Will do

PR #76, opened 2026-01-16
handle empty labels in selectors

rodesai:
## Summary

Handles empty labels in selectors as defined in the prometheus query spec. Required for loading the grafana metrics explorer panel

## Test Plan

- unit tests
- grafana quickstart

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -411,8 +411,9 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
                 let fut = self.evaluate_binary_expr(b, start, end, interval, lookback_delta);
                 Box::pin(fut)
             }
-            Expr::Paren(_p) => {
-                todo!()
+            Expr::Paren(p) => {
+                let fut = self.evaluate_expr(&p.expr, start, end, interval, lookback_delta);
+                Box::pin(fut)
             }
             Expr::Subquery(_q) => {
                 todo!()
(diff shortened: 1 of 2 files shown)


timeseries/src/promql/evaluator.rs
```diff
@@ -411,8 +411,9 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
                 let fut = self.evaluate_binary_expr(b, start, end, interval, lookback_delta);
                 Box::pin(fut)
             }
-            Expr::Paren(_p) => {
-                todo!()
+            Expr::Paren(p) => {
+                let fut = self.evaluate_expr(&p.expr, start, end, interval, lookback_delta);
```
rodesai, line 415: pulled in this quick change as well to handle paren expressions

timeseries/src/promql/selector.rs
```diff
@@ -228,8 +288,27 @@ fn find_candidates_with_regex_support(
...
+                name: METRIC_NAME.to_string(),
+                value: name.clone(),
+            };
+            let result_set: Vec<SeriesId> =
+                inverted_index.intersect(vec![metric_term]).iter().collect();
+            return Ok(result_set);
+        } else {
+            // No metric name specified - this would match all series, but that's expensive
+            // For now, return empty as this case should be rare
+            return Ok(Vec::new());
```
agavra, line 310: curious why this returns empty whereas the labels one returns an error. does grafana depend on one?

agavra (approved):

timeseries/src/promql/selector.rs
rodesai: I just missed it during review (Claude had originally implemented the other one to just return the wrong results as well). FWIW this is only called from `evaluate_on_indexes` which is only called from test code. We should probably clean it up.

PR #77, opened 2026-01-16
update quickstart to use docker-compose and Grafana

rodesai:
## Summary

Updates the timeseries quickstart to use docker-compose for 1-line setup/teardown, and have it deploy a Grafana node so users can interact with the timeseries server over a GUI. Also moves all quickstart related stuff to the quickstart directory

## Test Plan

Ran through quickstart manually

## Checklist

- [x] Tests added/updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ The flip side of this decoupled architecture is that you have higher end-to-end 
 
 # Quick Start
 
-* [TSDB](./timeseries/README.md)
+* [TSDB](./timeseries/README.md#quickstart)
 * Log: TODO
 * Vector: TODO
 


timeseries/quickstart/Dockerfile.timeseries
```diff
@@ -0,0 +1,47 @@
+# Multi-stage build for OpenData Timeseries
```
agavra, line 1: for now this is good be we should figure out if there's a way for us to share a dockerfile/build pipeline for all the databases, especially once we have a more unified entrypoint layer cc @apurvam

rodesai: Yep totally agreed here.

PR #78, opened 2026-01-16
Sketch out common bencher rfc

hachikuji:
Started a draft for a benching framework. I thought I would get some feedback first before proposing a specific design. High-level, it would be cool if we could run system benchmarks periodically and store the results somewhere public. An agent could inspect them and try to identify candidate regressions from the commit log.

### DIFF
--- a/rfcs/0003-benchmark-framework.md
+++ b/rfcs/0003-benchmark-framework.md
@@ -7,62 +7,35 @@
 
 ## Summary
 
-This RFC proposes a common benchmark framework for opendata that standardizes
-provisioning and configuration aspects while remaining flexible enough to
-accommodate the diverse needs of each database system (TSDB, Log, Vector, etc.). The
-framework introduces minimal abstractions, allowing each system to plug in its
-own workload generators and measurement concerns. Cross-cutting capabilities
-like telemetry collection are handled by the framework, and benchmark results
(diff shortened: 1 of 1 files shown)


agavra (approved):
LGTM! It'd be nice to have a good way to define workloads that is reusable for each system. Figuring out a reliable runner is something we'll probably have to do as well, don't think the GHA runners are super reliable from our slate experience.

apurvam (commented):
This looks good to me. I'm fine to merge it as is to articulate the problem. But one thing to add would be an example of defining a benchmark, running it, collecting results, etc. to make the work concrete.

hachikuji:
@agavra @apurvam I did another pass on this RFC. I wanted to focus on defining a basic bench framework which handles environment setup and not much else. I left out CI integration, regression analysis, etc. for the future. Not sure we really need RFCs for these efforts since they are not public anyway. I think they can be a bit more organic.

hachikuji:
I'm going to merge this. I wouldn't consider any of it set in stone, but nice to have a starting point. Initially I wanted to try writing a benchmark for the log since it should be the most straightforward to define workloads. I imagine we'll revise the API a lot after usage becomes clear.

apurvam:
Makes sense. Writing out a plan whether in an RFC or otherwise is good for discussion / brainstorming, if nothing else.

Issue #96, opened 2026-01-17
[tsdb] minor/cleanup: implement `Ord` for `Label`


rodesai:
There are a bunch of places in tsdb where we sort a set of labels using `sort_with`. Since `Label` doesn't implement `Ord`, each of these points specifies its own implementation of ordering (first by key, then by value). We should just implement `Ord` for `Label` and use `sort` everywhere we need to sort.

Anusha0501:
Thanks for opening this issue! Before I start, just wanted to confirm a couple of things:

1. Is the canonical ordering for `Label` always lexicographic by `key` first, then `value`?
2. Is default `String::cmp` (case-sensitive, no normalization) the intended behavior?
3. Would deriving `Ord`/`PartialOrd` be acceptable if field order matches, or do you prefer a manual impl?
4. Should this PR replace all existing label-specific sort logic with `sort()` within tsdb?

Once confirmed, I’m happy to take this up.

agavra:
Hi @Anusha0501 - I had missed your comment on this PR before noticing https://github.com/opendata-oss/opendata/pull/103 which LGTM. If you'd like you can follow up that PR with one that derives the traits (just add some tests to make sure the order remains consistent) and I'll approve that one, I like the suggestion.

also labels should be case sensitive, as that is the prometheus default behavior

PR #103, opened 2026-01-17
tsdb: implement Ord for Label and simplify canonical label sorting

siphonite:
This PR implements a canonical `Ord` / `PartialOrd` for `Label` using
lexicographic ordering by `(name, value)`, matching existing behavior.

It also replaces a small number of label-specific `sort_by` calls that
already used `(name, value)` ordering with `sort()`, relying on the new
canonical ordering.

Name-only and domain-specific sorts were intentionally left unchanged.

Happy to adjust if a different scope or ordering semantics are preferred. 

Refs #96

### DIFF
--- a/timeseries/src/model.rs
+++ b/timeseries/src/model.rs
@@ -73,7 +73,9 @@ impl Label {
 /// for fingerprint computation and series identification.
 impl Ord for Label {
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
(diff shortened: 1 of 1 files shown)


agavra:
Looks like there a bit of formatting issues. cargo fmt should fix that (and make sure to check clippy as well)

siphonite:
Thanks! I’ve run cargo fmt and clippy and pushed the fixes.

PR #104, opened 2026-01-17
Update Readme

apurvam:
## Summary

Update README to focus more on the problems we solve and how. Add a roadmap. Make it easier for potential contributors to engage.

## Related Issues
None.

## Test Plan
N/A - docs only.

## Checklist
- [ x] `cargo fmt` and `cargo clippy` pass
- [ x] Documentation updated (if applicable)


README.md
```diff
@@ -5,134 +5,78 @@
...
 
-OpenData is a collection of open source databases designed from ground up for object storage. We aim to deliver highly focused, objectstore-native versions of online databases that can power your application stack. As of today, we have a competitive timeseries database and a data streaming backend. We plan to ship vector search, text search, and other database types over time. Contributions are welcome!
+OpenData is a collection of databases built on a shared foundation. Same storage layer, same operational patterns, same tooling. The goal: make running multiple specialized databases feel like running one system.
 
-Building performant, cost-effective, and correct online database on object storage takes special care. Successful designs all have to solve the problem of write batching, multiple levels of caching, and snapshot isolation for correctness. OpenData databases build on a common foundation to solve these problems. This common foundation gives our databases a common set of operational tools, configuration systems, etc., that make our databases easier to operate in aggregate. 
+## The Problem OpenData is solving
```
agavra: I wouldn't lean so much on counter-positioning against "postgres for everything" but instead assume that people want multiple databases. There are better places to make the case for where OpenData fits in the larger ecosystem than in the README's first paragraph. 

This also feels a little choppy and markety. I much preferred the tone of the previous README which feltdry and  more technical(as I think a README should be).

apurvam: I agree. I'm trying to figure out a balance between going too deep technically too early, and being market-y. Tone aside, I think the right message to try and communicate in this README right now is: 

1. What problem we are solving and how. The postgres counter-positioning is probably unnecessary for this. Instead, we can just say that you eventually need specialized databases, and when you do so, they all come with their own unique baggage, making operations hard. 
2. Our solution is a shared storage and control plane (for lack of a beter word) foundation. 
3. Roadmap
4. contribute.

We don't need 'Why now' section to communicate those 4 things eithr. Do you think that makes it feel market-y / pitch-y ?  

If we agree on those bullets, then I can play with different tones.

apurvam: And if we want to keep in compact with and communicate those 4 things, I do feel it leans more toward being market-y. I feel that's a fundamental trade off, since these are 'product' talking points, not 'engineering' talking points.

apurvam: Another option is to add back the architecture section from the old readme (the high level architecture, not the detailed data flow). So that expands on the 'How', which is more engineering oriented.

apurvam: Once we align on this thread, I can rewrite the content accordingly.

apurvam: Looking at your sampling of READMEs, I think the entire content of this PR needs to go to a separate document. Something like 'Vision'. Or we need a page on the website to communicate the product vision. Because that's what this content is. There is also the technical vision, which is what the previous readme spent most of its time on. 

If we emulate the other readmes, then we basically have "useful links", "how to build", "how to contribute".  the links to the vision docs could be in the 'useful links' section.

agavra: I agree with the four points, even within this README. It's basically:

1. why / what
2. how
3. what's next
4. minutia (probably also makes sense to have links to quickstarts as they come up, with "Comming Soon" for vector/tsdb)

Points 3/4 have no risk of coming across as markety. Point 2 can pretty easily be made dry and technical by making it a "list of features", in our case the "common features" that back them.

Point 1 is definitely at risk of being markety, so I think we should limit ourselves to at most one paragraph for it. Here are some examples pulled from other READMEs:

> LanceDB is designed for fast, scalable, and production-ready vector search. It is built on top of the Lance columnar format. You can store, index, and search over petabytes of multimodal data and vectors with ease. LanceDB is a central location where developers can build, train and analyze their AI workloads.

> Claude Code is an agentic coding tool that lives in your terminal, understands your codebase, and helps you code faster by executing routine tasks, explaining complex code, and handling git workflows -- all through natural language commands. Use it in your terminal, IDE, or tag claude on Github.

> ClickHouse® is an open-source column-oriented database management system that allows generating analytical data reports in real-time.

I don't think that developers coming to this repo reading the README are going to read more than that anyway. Either way, I think it's a good exercise for us to distill the message to something that concise (which I think we've made progress with that in the manifesto, we can reuse a lot of that here).

apurvam: Makes sense. I think the the 'how' can shrink as a % of the readme and minutia can increase as time goes on. At this point this is our best real estate to get our message out. Most READMEs don't try to get the message out. So I expect this setup will be short lived.

apurvam: I'll rewrite to have a very short (3 sentence) why/what. A compact 'how' which covers a high level technical vision in 2 paragraphs or so (similar to what's in this PR). Then the rest: Roadmap, quickstarts, etc. Each progressive version can get toward moving the marketing out of the README and substituting in more tangible details (how to build, etc.).

WDYT?

agavra: Perfect 👍

apurvam: Ok. I rewrote it based on this thread. I'm feeling much better about this. LMKWYT.

README.md
```diff
@@ -5,134 +5,57 @@
...
+service infrastructure, service catalog,  admin tooling, distributed state infrastructure, configuration systems, 
+and testing frameworks. 
 
-* **TSDB**: An objectstore native timeseries database that can serve as a backend for Prometheus. Its a great option for a low cost, easy to operate, grafana backend. [Learn more](timeseries/rfcs/0001-tsdb-storage.md).
-* **Log**: Think of it as Kafka 2.0. An objecstore native event streaming backend that supports millions of logs, so you can finally get a replayable log per key. [Learn more](log/rfcs/0001-storage.md).
-* **Vector**: An objectstore native vector database for approximate nearest neighbor search. SPANN-style architecture with centroids in memory and posting lists on disk for billion-scale capacity. [Learn more](vector/rfcs/0001-storage.md).
+Taken together, with OpenData there is only one storage engine and one set of operational tooling to learn across
+every database. Individual databases focus only on what makes them unique—query semantics, data layout, use-case 
+optimizations—while inheriting everything else.
```
agavra: while I personally don't mind the emdash, it is a dead giveaway that this is AI generated which I think we should aim to avoid in the README purely from a perception standpoint.

Consider:
> Taken together, with OpenData there is only one storage engine and one set of operational tooling to learn across all systems. By inheriting these key components, individual databases focus only on the unique query semantics, data layout and optimizations.

README.md
```diff
@@ -5,134 +5,57 @@
...
+easier and cheaper to operate. 
 
-Building performant, cost-effective, and correct online database on object storage takes special care. Successful designs all have to solve the problem of write batching, multiple levels of caching, and snapshot isolation for correctness. OpenData databases build on a common foundation to solve these problems. This common foundation gives our databases a common set of operational tools, configuration systems, etc., that make our databases easier to operate in aggregate. 
+The common foundation has two distinct components:
 
-# Databases
+**SlateDB as the common storage layer:** [SlateDB](https://www.slatedb.io) is an object-store-native LSM tree that
+handles write batching, tiered caching, and compaction. It provides snapshot isolation via atomic manifest updates 
+using S3's compare-and-set. Each OpenData database is essentially a data structure built on top of SlateDB—defining 
+its own data layout and query implementation while inheriting the storage fundamentals.
```
agavra: this is duplicated below, I think we can just say: "Each individual OpenData database implements its own domain-specific data structure and query layer on top of SlateDB" and ending it with that.

README.md
```diff
@@ -5,134 +5,57 @@
...
-* Vector: TODO
+## Bigger ideas for later
+- [ ] Shared ingest layer: all writes got a shared service for better batching. A compactor framework creates queryable state.
+- [ ] Flexible deployment modes: embedded writers + hosted readers, fully embedded, fully distributed, etc.
+- [ ] Deterministic simulation testing.
 
+# Get Involved
 
-# Architecture
+We're early and building in the open.
```
agavra: I think we should drive people to the discord front and center. Something like "We are early and building in the open, with most discussions happening on [Discord](https://discord.gg/35xhun9EgZ)"

(Also make sure to use this new link, the old one expired b/c turns out the default links expire in 7 days)

README.md
```diff
@@ -5,134 +5,57 @@
 [![Discord](https://img.shields.io/badge/discord-join-7289DA?style=flat-square&logo=discord)](https://discord.gg/EUv3Bqcv)
 [![GitHub License](https://img.shields.io/github/license/opendata-oss/opendata?style=flat-square)](LICENSE)
 
-OpenData is a collection of open source databases designed from ground up for object storage. We aim to deliver highly focused, objectstore-native versions of online databases that can power your application stack. As of today, we have a competitive timeseries database and a data streaming backend. We plan to ship vector search, text search, and other database types over time. Contributions are welcome!
+OpenData is a collection of open source databases built on a common storage and infrastructure foundation. This shared
```
agavra:
```suggestion
OpenData is a collection of open source databases built on a common, object-native storage and infrastructure foundation. This shared
```
I know "object-storage" isn't a big deal for a good chunk of people, but I think having those two words here won't distract anyone that doesn't care but will pull in people who do. In other words it's not _that_ much real estate but it could be valuable.

README.md
```diff
@@ -5,134 +5,57 @@
...
+foundation means every database has a virtually identical operational profile, which makes our database fleet materially 
+easier and cheaper to operate.
```
agavra: english nit: 
```suggestion
easier and cheaper to operate than alternatives. 
```

PR #106, opened 2026-01-18
use FoundationDB-style tuple encoding for inverted index keys

### LINKED ISSUE #26: update timeseries storage to use FoundationDB style tuples for inverted index
## Summary

Currently the index uses length-prefix encoding for utf8 strings in the keys for the inverted index (e.g. term:value are encoded as `[len | utf8][len | utf8]`) Instead we should use FoundationDB-style escaped tuples with 0x00 delimiters.

## Motivation

This keeps lexicographical sorting of terms and reduces the total index size. I'm not certain that lexicographical sorting helps that much, but it's better than length-ordered in situations when many terms for the same key are examined (e.g. `{env:prod OR env:dev}

## Proposed Solution

See #23 for an example in the log implementation

## Alternatives Considered

- The existing approach (has issues above)
- A dictionary approach saves space but adds significant complexity and doesn't guarantee ordering

## Additional Context

see https://github.com/apple/foundationdb/blob/main/design/tuple.md

siphonite:
This PR updates the timeseries inverted index key encoding to use a
FoundationDB-style tuple encoding instead of length-prefixed UTF-8.

### Why
Length-prefixed UTF-8 encoding does not preserve lexicographical ordering,
which can cause incorrect behavior during range scans over inverted index
keys (e.g. label enumeration).

Tuple-style, delimiter-based encoding preserves ordering and avoids key
interleaving issues, consistent with the approach used in #23.

### What changed
- Added tuple-style encoder/decoder scoped to inverted index keys
- Updated inverted index key encoding and attribute range scans
- Kept shared UTF-8 encoding utilities unchanged
- Updated tests to reflect delimiter-based semantics

### Testing
- `cargo test`
- `cargo fmt`
- `cargo clippy`

Fixes #26

### DIFF
--- a/timeseries/rfcs/0001-tsdb-storage.md
+++ b/timeseries/rfcs/0001-tsdb-storage.md
@@ -350,10 +350,6 @@ unambiguously separated from the value field. The value does not require a
 terminator since it extends to the end of the key. This minimizes key size while
 preserving the ability to do efficient prefix scans by label name.
 
-> [!WARNING]
-> **Backward Incompatibility**: This key encoding is not backward compatible with
-> the previous length-prefixed UTF-8 encoding. Existing inverted index data must
-> be migrated or discarded before upgrading.
 
 Note that since the label names and values are stored lexicographically,
 SlateDB's prefix encoding will be very effective without the use of a


timeseries/src/serde/key.rs
```diff
@@ -6,6 +6,62 @@ use bytes::{Bytes, BytesMut};
 use common::BytesRange;
 use common::serde::key_prefix::KeyPrefix;
 
+/// Encodes a string as a tuple element for inverted index keys.
+///
+/// This uses a FoundationDB-style tuple encoding with a `0x00` delimiter
+/// to preserve lexicographical ordering for range scans. This helper is
+/// intentionally scoped to inverted index key encoding and does not replace
+/// shared UTF-8 encoding utilities.
+fn encode_tuple_string(buf: &mut Vec<u8>, s: &str) {
```
agavra: we should reuse `terminated_bytes.rs`

timeseries/src/serde/key.rs
```diff
@@ -176,26 +232,30 @@ pub struct InvertedIndexKey {
...
             .prefix_with_bucket_size(self.bucket_size)
-            .write_to(&mut buf);
+            .write_to(&mut prefix_buf);
+        buf.extend_from_slice(&prefix_buf);
         buf.extend_from_slice(&self.time_bucket.to_be_bytes());
-        encode_utf8(&self.attribute, &mut buf);
-        encode_utf8(&self.value, &mut buf);
-        buf.freeze()
+        encode_tuple_string(&mut buf, &self.attribute);
+        encode_tuple_string(&mut buf, &self.value);
```
agavra: we only need the attribute to be encoded with foundationDB style since the `value` is delimited by the end of the key. please add a test to confirm this

timeseries/src/serde/key.rs
agavra: we should also make sure to update `timeseries/rfcs/0001-tsdb-storage.md` with the changes. it's OK that it's backwards incomaptible for now

agavra (commented):
Thanks @siphonite - this is a good start, left some comments inline

agavra:
Also unrelated, but if you're planning on contributing regularly I recommend you come hang out in our discord! A lot of the interesting discussions come here: https://discord.com/invite/35xhun9EgZ (if you tried earlier and got an expired link I apologize, I didn't notice the default links last only 7 days)

siphonite:
Thanks for the feedback @agavra. 

I will: 

- Reuse `terminated_bytes` instead of a local tuple helper
- Encode only the attribute with FoundationDB-style encoding (keep the value as-is) and add a test to confirm this behavior
- Update `timeseries/rfcs/0001-tsdb-storage.md` to reflect the change 

I will push an update shortly.

siphonite:
Hi @agavra, I’ve pushed an update addressing the comments.
Let me know if it looks good or if anything needs adjusting. 

- Thanks!

agavra:
Hey @siphonite I'll take a look at this later today, thanks for addressing the comments!

timeseries/rfcs/0001-tsdb-storage.md
```diff
@@ -323,20 +323,37 @@ and similarly for `B=2`.
...
+
+Only the label (attribute) field uses terminated encoding because it must be
+unambiguously separated from the value field. The value does not require a
+terminator since it extends to the end of the key. This minimizes key size while
+preserving the ability to do efficient prefix scans by label name.
+
+> [!WARNING]
+> **Backward Incompatibility**: This key encoding is not backward compatible with
+> the previous length-prefixed UTF-8 encoding. Existing inverted index data must
+> be migrated or discarded before upgrading.
```
agavra: this is fine, we can remove it for now since this isn't a production ready system - just will be distracting if anyone reads this as documentation

siphonite: Thanks for clarifying.

I have removed the warning and pushed an update.

agavra:
Thanks for the contribution @siphonite :) happy to have you contributing regularly.

PR #109, opened 2026-01-19
Fix parsing of match[] query parameters for label values API

### LINKED ISSUE #14: tsdb p8s server match[] parameters cannot be parsed over curl
## Description

## Steps to Reproduce

```

```

## Expected Behavior

Rohan-flutterint:
## Summary

What does this PR do?
This PR fixes the parsing of Prometheus-style match[] query parameters for the
/api/v1/label/{name}/values endpoint.

Axum’s Query extractor cannot correctly deserialise repeated match[] parameters
in URL-encoded queries, causing valid requests to fail before reaching the handler.
The handler now parses match[] parameters directly from the raw query string while
preserving existing support for start, end, and limit.

## Related Issues
Fixes #14

## Test Plan

How was this tested?

- Reproduced the issue locally using curl as described in the issue
- Verified single and multiple match[] parameters work correctly
- Verified behavior with optional start, end, and limit parameters

**Ran:**
> - `cargo check`
> - `cargo fmt`
> - `cargo clippy`

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -423,8 +423,14 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
                 let val = l.val;
                 Box::pin(async move { Ok(ExprResult::Scalar(val)) })
             }
-            Expr::StringLiteral(_l) => {
-                todo!()
+            Expr::StringLiteral(l) => {
+                let val = l.val.clone();
+                Box::pin(async move {
+                    Err(EvaluationError::InternalError(format!(
+                        "string literal \"{}\" is not directly evaluatable",
+                        val
+                    )))
+                })
             }
             Expr::VectorSelector(vector_selector) => {
                 let fut = self.evaluate_vector_selector(vector_selector, end, lookback_delta);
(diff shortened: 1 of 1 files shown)


timeseries/src/promql/server.rs
```diff
@@ -279,12 +295,57 @@ async fn handle_labels(
 }
 
 /// Handle /api/v1/label/{name}/values
+// NOTE: Axum's Query extractor cannot correctly deserialize repeated `match[]`
+// parameters in URL-encoded queries. We manually parse them from the raw query
+// string to restore Prometheus-compatible behavior.
 async fn handle_label_values(
```
agavra, line 282: can we add unit tests to ensure the behavior works as intended?

timeseries/src/promql/server.rs
```diff
@@ -35,6 +38,19 @@ pub(crate) struct AppState {
...
 }
 
+/// Helper params struct used to deserialize label values query parameters
+/// *excluding* `match[]`.
+///
+/// Axum's `Query` extractor cannot deserialize repeated `match[]` parameters,
+/// so `match[]` is parsed manually from the raw query string and the remaining
+/// parameters are deserialized using this struct.
+#[derive(Debug, Deserialize, Default)]
+struct LabelValuesParamsNoMatch {
```
rodesai: can we drop `LabelValuesParams`? It's not used anywhere anymore.

Issue #83, opened 2026-01-17
[tsdb] support string literals in PromQL expressions


rodesai:
## Summary

We should add support for string literals in PromQL expressions.

## Motivation

Full PromQL support

## Proposed Solution

This involves filling in the `todo` under the `StringLiteral` branch under `evaluate_expression` in the evaluator.

siphonite:
I’d like to take this up.

Before implementing, I want to confirm scope:
- Is the intent to support string literals primarily as evaluatable values that can be passed to functions (e.g. label_replace), rather than as standalone query results?

- For a top-level string literal expression (e.g. `"foo"`), should evaluation return an error or a typed value?

I’m planning to start by filling in the StringLiteral branch in the evaluator with minimal, spec-aligned support.

agavra:
Typically we should copy what promql does on any of these, so you can always spin up a local p8s instance and then check the behavior if you are unsure. 

I'm not actually sure though what @rodesai intended with this, but I think "the intent to support string literals primarily as evaluatable values that can be passed to functions" sounds correct to me.

Top-level string expressions are probably lower priority (if prometheus even supports those, I haven't double checeked)

siphonite:
Thanks, this makes sense.

I have implemented a minimal fix that replaces the `todo!()` with a structured evaluation error for standalone string literals, matching PromQL behavior and avoiding a panic.

This keeps scope limited while leaving room for future work to support string literals as function arguments (e.g. label_replace).

I’ll open a PR shortly for review.

rodesai:
> I'm not actually sure though what @rodesai intended with this, but I think "the intent to support string literals primarily as evaluatable values that can be passed to functions" sounds correct to me.

Yeah that's right. That's pretty much the only use case for string literals inside of queries afaict.

PR454D:
This issue seems to be stale, resolved by #110. Should be closed.

apurvam:
You're right. Thanks for pointing it out.

PR #110, opened 2026-01-19
[tsdb] Handle standalone PromQL string literals without panic

siphonite:
## Summary

This PR replaces a `todo!()` in the PromQL evaluator for `Expr::StringLiteral` with a structured evaluation error.

String literals are valid PromQL syntax but cannot be evaluated as standalone query results. Previously, attempting to evaluate such an expression would panic.

This change:
- Avoids a runtime panic
- Returns a clear, user-facing evaluation error
- Keeps scope minimal and evaluator-only

## Related Issues

Relates to #83

## Test Plan

- Added a regression test asserting that evaluating a standalone string literal returns an error
- Ran existing literal evaluation tests

## Checklist

- [x] Tests added/updated
- [x] Documentation not required

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -1020,8 +1020,8 @@ mod tests {
     use crate::query::test_utils::MockQueryReaderBuilder;
     use crate::test_utils::assertions::approx_eq;
     use promql_parser::label::{METRIC_NAME, Matchers};
-    use promql_parser::parser::value::ValueType;
     use promql_parser::parser::EvalStmt;
+    use promql_parser::parser::value::ValueType;
     use rstest::rstest;
 
     use std::time::{Duration, SystemTime, UNIX_EPOCH};


timeseries/src/promql/evaluator.rs
```diff
@@ -423,8 +423,14 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
-            Expr::StringLiteral(_l) => {
-                todo!()
+            Expr::StringLiteral(l) => {
+                let val = l.val.clone();
+                Box::pin(async move {
+                    Err(EvaluationError::InternalError(format!(
+                        "string literal \"{}\" cannot be evaluated as a standalone PromQL expression",
+                        val
+                    )))
+                })
```
agavra, line 433: thanks @siphonite - I think it'd be helpful to implement one of the method contexts where string literals are used. That'll help us illustrate how `StringLiteral`s should be used

Since this is a recursive function it's possible that this is called within the context of something else (in other words, the error here would be misleading because it wasn't a standalone PromQL expression).

siphonite: Thanks.

I agree that emitting a “standalone expression” error can be misleading if the string literal is evaluated as part of a larger expression (e.g. inside a function call).

I can update this to either:
- make the error context-agnostic, or
- minimally support string literals flowing through a function-call context (without implementing full string semantics yet)

I’ll take a stab at the smallest useful context (function arguments) unless you’d prefer a different one.

agavra: your plan sounds good to me! Feel free to ping me when this is ready for another round of review.

siphonite: Thanks for review @agavra

I have pushed an update that:
- makes the StringLiteral error context-agnostic,
- adds a minimal function-argument context in `evaluate_call`, and
- fixes the CI issues 

Would appreciate another look when you have time.

agavra: Thanks I apologize for the late review I just saw this! Taking a look now.

agavra (approved):
Thanks @siphonite - returning a nice error is definitely better than the previous panic!

PR #115, opened 2026-01-21
[catalog] Sketch out initial RFC scope

hachikuji:
It would be useful to have a catalog for all opendata systems. This PR attempts to sketch out the potential scope for an initial catalog. We can complete design details once there's general consensus about it.

### DIFF
--- a/rfcs/0003-catalog.md
+++ b/rfcs/0003-catalog.md
@@ -11,25 +11,27 @@ This RFC proposes a catalog system for OpenData that serves as a central managem
 
 ## Motivation
 
-OpenData comprises multiple storage subsystems (log, timeseries, vector, etc.) that share a common foundation on SlateDB. As users deploy multiple instances of these systems, several cross-cutting operational concerns emerge:
+OpenData comprises multiple storage subsystems (log, timeseries, vector, etc.) that share a common foundation on SlateDB. As users deploy multiple instances of these systems, several operational concerns emerge that are not well-served by existing orchestration systems:
 
 1. **Discovery** — There is no unified way to enumerate which slates exist within an environment. Operators must track this information externally or inspect object storage directly.
 
-2. **Configuration management** — Each slate requires object storage configuration (bucket, path prefix, credentials). Without a central registry, this configuration is scattered and difficult to audit.
+2. **Usage tracking** — Orchestration systems like Kubernetes know what pods are running, but not which applications are actively reading or writing to a given slate. This information lives at the storage layer.
 
-3. **Lifecycle management** — Creating and deleting slates involves coordinating metadata across multiple locations. A catalog provides a single point of control for these operations.
+3. **Safe deprovisioning** — Before removing a slate, operators need to know whether it has active consumers. Without usage tracking, this requires manual coordination or risks data loss.
 
 4. **Operational tooling** — Cross-cutting tools (backup, monitoring, migration) need a way to discover and enumerate slates without subsystem-specific knowledge.
 
 A catalog addresses these concerns by providing:
 
-- A single source of truth for slate metadata
-- Consistent registration and deletion workflows
-- A foundation for future operational tooling that works across subsystem types
+- A registry of running slates via self-registration
+- Tracking of active readers and writers for each slate
+- A foundation for safety checks and operational tooling
 
 The catalog is itself a slate backed by SlateDB, ensuring it benefits from the same durability and operational characteristics as the systems it manages.
 
 ### Hypothetical Workflow
+
+Slates and their consumers self-register when they start up. The CLI is primarily for discovery and usage tracking:
 
 ```
 $ opendata slate list
@@ -37,8 +39,7 @@ NAME        TYPE        BUCKET
 events      log         s3://acme-data/events
 metrics     timeseries  s3://acme-data/metrics
 
-$ opendata slate register --name orders --type log --bucket s3://acme-data/orders
-Registered slate 'orders'
+# A new slate is provisioned and starts up, self-registering with the catalog...
 
 $ opendata slate list
 NAME        TYPE        BUCKET
@@ -47,21 +48,19 @@ metrics     timeseries  s3://acme-data/metrics
 orders      log         s3://acme-data/orders
 
 $ opendata slate describe orders
-Name:   orders
-Type:   log
-Bucket: s3://acme-data/orders
+Name:    orders
+Type:    log
+Bucket:  s3://acme-data/orders
+Writers: order-service
+Readers: analytics, billing
 
-$ opendata slate delete orders
-Deleted slate 'orders'
-
-$ opendata slate config events
-storage:
-  type: SlateDb
-  path: events
-  object_store:
-    type: Aws
-    region: us-west-2
-    bucket: acme-data
+# Before deprovisioning, check if it's safe to remove
+$ opendata slate describe metrics
+Name:    metrics
+Type:    timeseries
+Bucket:  s3://acme-data/metrics
+Writers: telemetry-collector
+Readers: dashboard, alerting
 ```
 
 ## Goals
@@ -69,13 +68,14 @@ storage:
 - Define "slate" as the fundamental unit of an OpenData system
 - Specify the metadata required to describe a slate (name, type, object storage location)
 - Establish the catalog as a SlateDB-backed registry of slate metadata
-- Define the registration process for adding new slates to the catalog
-- Define the deletion process for removing slates from the catalog
+- Define the self-registration process by which slates register themselves on startup
+- Define the deregistration process for when slates are shut down
+- Track active readers and writers for each slate
 - Ensure the catalog can manage slates of all types (log, timeseries, vector)
 
 ## Non-Goals
 
-_To be completed in a future revision._
+- **Provisioning** — The catalog does not manage target state or drive provisioning. It reflects current state only. Provisioning is the responsibility of external systems (Kubernetes, Terraform, etc.).
 
 ## Design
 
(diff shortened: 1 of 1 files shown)


rfcs/0003-catalog.md
```diff
@@ -0,0 +1,100 @@
...
+storage:
+  type: SlateDb
+  path: events
+  object_store:
+    type: Aws
+    region: us-west-2
+    bucket: acme-data
+```
+
+## Goals
```
apurvam, line 66: This is great! I wonder, however, if we should add a concept of the 'owner' of a slate. for example, if you have a prometheus server that's backed by a slate TSDB. Simply deleting the s3 buckets will orphan the server. We need to be able to know that there is a service that depends on the slate, where it lives, etc. so that the operations maintain the integrity of the system.

hachikuji: Makes sense. This gets into liveness I guess. I was hoping we could go with a model where the catalog only communicates with object storage. Perhaps we could add some kind of explicit fencing marker to each opendata system, which must be added by the writer. The purpose is to fence writers/readers and signal the catalog that it is safe to delete.

apurvam: To make sure I understand, the marker would be added by the writer to indicate that the slate is no longer being written to? And that makes it safe to delete? If so, I'd presume the readers would also have to write a marker. This is more like a lease system than fencing markers if that's how you are thinking it would work. 

From a user perspective, if you have a service, you want to manage the lifecycle of the service as a whole. For example, I think a useful admin `delete` operation would deprovision the service and optionally delete the data. Not just delete the data.

hachikuji: The marker could be like a poison pill inserted into the slatedb manifest. It would kill readers and writers. I think the main point is trying to define the communication model. How does the catalog interact with provisioning systems? How does the catalog interact with system readers/writers? The ideal from my perspective is that all communication with the catalog is done through object storage. It is a return of our "storage as protocol" idea at its heart. For example, the catalog could write provisioning requests as files in object storage. Some kind of k8s service could watch that file and do the actual provisioning work. Deletion workflows could follow a similar pattern.

hachikuji: Downstream systems could just be catalog readers I guess. They might follow changes to the catalog as any SlateDb reader does and act when necessary.

apurvam: I like this line of thinking. So then the catalog needs to hold enough metadata to make those flows possible. I _think_ this is captured under 'define process for registering/deleting slates' in your goals. So I can imagine this type of flow in the longer term. 

1. User issues a deprovisioning request for some OpenData database `db0`. this could be through a UI, terraform, whatever.
2. This request is received by some control plane. 
3. The control plane writes this poison pill into the catalog.
4. A k8s operator detects the poison pill and instructs the readers/writers services to shut down. This means it needs a mapping from the slate to the reader and writer pods. Alternately, the the readers and writer pods of `db0` could catalog readers, read the catalog, and mark themselves for deprovisioning with the operator simply executing the action. I like the latter approach because it works for really any deployment model and doesn't require the operator to maintain additional metadata.

Does that match what you had in mind?

hachikuji: Yeah, I think that's right. I guess the main point is that downstream systems are just Readers in our Slatedb-backed framing. So provisioning systems would follow the catalog as readers. Perhaps they could even modify the catalog themselves by temporarily assuming the Writer role. Perhaps we do not need long-lived writers at all. If we could get a model like this to work, it would remove a huge amount of complexity. You don't need to have catalog as a persistent service.

hachikuji: I added some text to the RFC about the communication model. I like it. It leans into the advantages of slatedb/object storage.

apurvam: Nice. I like the direction. I think SlateDB transactions will be crucial to making this work.

rfcs/0003-catalog.md
```diff
@@ -0,0 +1,154 @@
...
+
+4. **Operational tooling** — Cross-cutting tools (backup, monitoring, migration) need a way to discover and enumerate slates without subsystem-specific knowledge.
+
+A catalog addresses these concerns by providing:
+
+- A single source of truth for slate metadata
+- Consistent registration and deletion workflows
+- A foundation for future operational tooling that works across subsystem types
+
+The catalog is itself a slate backed by SlateDB, ensuring it benefits from the same durability and operational characteristics as the systems it manages.
```
agavra: probably worth specifying whether or not we want the catalog to be a requirement to run the opendata stack. My preference is making it optional, but I could see the argument for requiring it in certain deployment modes so we can assume it exists

rfcs/0003-catalog.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Hypothetical Workflow
+
+```
+$ opendata slate list
+NAME        TYPE        BUCKET
+events      log         s3://acme-data/events
+metrics     timeseries  s3://acme-data/metrics
+
+$ opendata slate register --name orders --type log --bucket s3://acme-data/orders
```
agavra: I guess to my previous question this is something that (at the moment) is manually run? It probably makes sense for the service layer to accept a catalog endpoint so services can auto-register with the catalog. WDYT?

cc @apurvam

rfcs/0003-catalog.md
```diff
@@ -0,0 +1,154 @@
...
+**Status**: Draft
+
+**Authors**:
+- [Jason Gustafson](https://github.com/hachikuji)
+
+## Summary
+
+This RFC proposes a catalog system for OpenData that serves as a central management plane for OpenData storage systems. The catalog provides a single point to manage metadata about the systems ("slates") a user has installed, including their names, types, and object storage configuration. The catalog itself is implemented as a slate backed by SlateDB, following the same patterns as other OpenData subsystems.
+
+## Motivation
```
agavra, line 12: I had a review of this lined up but now I'm wondering whether this RFC is over-prescribed to a dogfood philosophy. What if instead we tried to design this as k8s-native?

I posed this question to Claude and here's an alternative we came up with:

The current RFC essentially builds a bespoke control plane on top of SlateDB, but if your primary deployment target is Kubernetes, you'd be reinventing machinery that K8s already provides (watches, reconciliation loops, status subresources, RBAC, etc.).

Instead of the catalog being a SlateDB-backed store that components poll, the Kubernetes API server becomes the catalog. Each slate is represented as a Custom Resource, and an operator reconciles desired state to actual state.

Then using the CRDs we could use:

```
$ kubectl get slates
NAME      TYPE        BUCKET                     PHASE
events    log         s3://acme-data/events      Provisioned
metrics   timeseries  s3://acme-data/metrics     Provisioned

$ kubectl apply -f - <<EOF
apiVersion: opendata.io/v1alpha1
kind: Slate
metadata:
  name: orders
spec:
  type: log
  objectStore:
    bucket: s3://acme-data/orders
EOF
slate.opendata.io/orders created

$ kubectl get slate orders -o jsonpath='{.status.phase}'
Provisioned

$ kubectl delete slate orders
slate.opendata.io/orders deleted
```

### Tradeoffs

| Aspect | SlateDB-backed Catalog | K8s-native CRDs |
|--------|------------------------|-----------------|
| **Dependency** | Only object storage | Requires Kubernetes |
| **Discovery** | Must know catalog location | Standard K8s API discovery |
| **Watches** | Polling (or custom mechanism) | Native watch support |
| **Auth/RBAC** | Custom | K8s RBAC out of the box |
| **Tooling** | Custom CLI | kubectl, GitOps, Helm, etc. |
| **Dogfooding** | Uses OpenData's own primitives | External control plane |
| **Portability** | Runs anywhere with object storage | K8s-only (or needs abstraction) |

The big benefit, then, of using opendata is that you install just one operator and one common language for CRDs instead of learning a new operator and new CRD for each of the data systems your deploy in your k8s stack.

agavra: I also think philosophically it's OK to lean into kubernetes + object storage as the two primitives we rely on. The 'pitch' in my mind is that those two solve the hardest distributed systems problems: the former solves elastic compute and the latter solves elastic storage/consistency. Without both opendata's vision can't come to fruition. 

Another big win with using `kubectl` as the primary control plane CLI is that the AI agents are really good with it.

apurvam: We should definitely design to be k8s native, but I don't think a catalog as being discussed here is occupying the same place as k8s. Any system will need some storage to figure out what's deployed, where it's deployed etc. The deployments very often span regions and k8s clusters. The question is: where is _that_ information going to live. We need a catalog for that, which drives the k8s actions in a particular region.

agavra: I guess I conflated the two. I believe we should start by figuring out the deployment models in a single k8s cluster and work up from there. The CLI as proposed here has a lot of overlap with the type of things that k8s should handle for me if its all within a single k8s cluster.

I'm not convinced that multi-region/multi-k8s is something we should figure out until we have a solid understanding of the single-k8s, multi-AZ design. A single k8s cluster can span multiple AZs, which is likely where 99% of data systems stop.

hachikuji: I agree with the concern. I don't think we want the catalog directly involved in provisioning. At the same time, I'm not too comfortable being super dogmatic about k8s and sticking it at the heart of the system. I took a shot at reframing the catalog in the latest patch. Rather than tracking a target state, the catalog might simply track the current state. It might be aware of active readers/writers in the system.  Kubernetes could consult the catalog prior to deprovisioning a resource rather than having the catalog drive deprovisioning itself for example. Not sure if this is enough value to justify the catalog's existence just yet. I suspect we need to let this stew for a while.

rfcs/0003-catalog.md
```diff
@@ -0,0 +1,163 @@
...
+#### Example: Self-Registration on Startup
+
+The catalog sits downstream of provisioning. Each slate can be configured with an optional reference to a catalog. When a slate starts up, it registers itself, ensuring the catalog reflects the current state of running systems.
+
+Consider a flow where a Kubernetes operator provisions a new slate:
+
+1. A user creates a Kubernetes Custom Resource specifying a new slate.
+2. The K8s operator provisions the slate with a catalog reference in its configuration.
+3. When the slate starts, it assumes the **Writer** role to register itself in the catalog.
+4. CLI tooling or other components can observe the catalog as **Readers** to discover running slates.
```
apurvam, line 100: I like the new version a lot. I agree this needs time to bake. And I think the crux of what needs to bake is this 4th item. What do the _readers_ actually do with the data in the catalog. I think this, more than anything, is what will inform if and how the catalog co-exists with orchestration systems like k8s.

PR #116, opened 2026-01-22
Add HTTP server for the log

apurvam:
## Summary

Add an HTTP server for the log. If the `ACCEPT` or `Content-Type` HTTP header is `application/x-protobuf` a protobuf payload will be expected or returned. 

It works like so: 

```zsh

# Append
apurva@Apurvas-MacBook-Pro opendata %   curl -X POST http://localhost:8080/api/v1/log/append \
    -H "Content-Type: application/json" \
    -d '{"records": [{"key": "ZXZlbnRz", "value": "ZXZlbnQtMQ=="}, {"key": "ZXZlbnRz", "value":
  "ZXZlbnQtMg=="}]}'

{"recordsAppended":2,"startSequence":0,"status":"success"}

# Scan 
apurva@Apurvas-MacBook-Pro opendata %   curl "http://localhost:8080/api/v1/log/scan?key=events" | jq
{
  "entries": [
    {
      "sequence": 0,
      "value": "ZXZlbnQtMQ=="
    },
    {
      "sequence": 1,
      "value": "ZXZlbnQtMg=="
    }
  ],
  "key": "ZXZlbnRz",
  "status": "success"
}


# scan range

apurva@Apurvas-MacBook-Pro opendata %   curl "http://localhost:8080/api/v1/log/scan?key=events&start_seq=0&end_seq=10&limit=100" | jq

{
  "entries": [
    {
      "sequence": 0,
      "value": "ZXZlbnQtMQ=="
    },
    {
      "sequence": 1,
      "value": "ZXZlbnQtMg=="
    }
  ],
  "key": "ZXZlbnRz",
  "status": "success"
}

# list keys

apurva@Apurvas-MacBook-Pro opendata %   curl "http://localhost:8080/api/v1/log/keys" | jq

{
  "keys": [
    "ZXZlbnRz"
  ],
  "status": "success"
}


# list segments

apurva@Apurvas-MacBook-Pro opendata %   curl "http://localhost:8080/api/v1/log/segments" | jq
{
  "segments": [
    {
      "id": 0,
      "startSeq": 0,
      "startTimeMs": 1769478662550
    }
  ],
  "status": "success"
}

```

the API is like so: 
```
  ┌──────────────────────┬────────┬────────────────────────────────────────────┐
  │       Endpoint       │ Method │                Description                 │
  ├──────────────────────┼────────┼────────────────────────────────────────────┤
  │ /api/v1/log/append   │ POST   │ Append records to the log                  │
  ├──────────────────────┼────────┼────────────────────────────────────────────┤
  │ /api/v1/log/scan     │ GET    │ Scan entries by key and sequence range     │
  ├──────────────────────┼────────┼────────────────────────────────────────────┤
  │ /api/v1/log/keys     │ GET    │ List distinct keys within a segment range  │
  ├──────────────────────┼────────┼────────────────────────────────────────────┤
  │ /api/v1/log/segments │ GET    │ List segments overlapping a sequence range │
  ├──────────────────────┼────────┼────────────────────────────────────────────┤
  │ /api/v1/log/count    │ GET    │ Count entries for a key                    │
  ├──────────────────────┼────────┼────────────────────────────────────────────┤
  │ /metrics             │ GET    │ Prometheus metrics                         │
  └──────────────────────┴────────┴────────────────────────────────────────────┘

```

## Related Issue

Implements: #117 

## Test Plan

Added unit and integration tests. Also manually test on the command line with a local log directory. 

## Checklist

- [x] Tests added/updated

### DIFF
--- a/log/src/server/handlers.rs
+++ b/log/src/server/handlers.rs
@@ -10,7 +10,7 @@ use axum::extract::{Query, State};
 use axum::http::HeaderMap;
 
 use super::error::ApiError;
-use super::metrics::{AppendLabels, Metrics, OperationStatus, ScanLabels};
+use super::metrics::Metrics;
 use super::proto::{
     AppendResponse, CountResponse, KeysResponse, ScanResponse, Segment, SegmentsResponse, Value,
 };
@@ -46,37 +46,15 @@ pub async fn handle_append(
         await_durable: request.await_durable,
     };
 
-    match state
+    let result = state
         .log
         .append_with_options(request.records, options)
-        .await
-    {
-        Ok(result) => {
-            state.metrics.log_append_records_total.inc_by(count as u64);
-            state
-                .metrics
-                .log_append_requests_total
-                .get_or_create(&AppendLabels {
-                    status: OperationStatus::Success,
-                })
-                .inc();
+        .await?;
 
-            let response =
-                AppendResponse::success(result.records_appended as i32, result.start_sequence);
-            Ok(to_api_response(response, format))
-        }
-        Err(e) => {
-            state
-                .metrics
-                .log_append_requests_total
-                .get_or_create(&AppendLabels {
-                    status: OperationStatus::Error,
-                })
-                .inc();
+    state.metrics.log_append_records_total.inc_by(count as u64);
 
-            Err(ApiError::from(e))
-        }
-    }
+    let response = AppendResponse::success(result.records_appended as i32, result.start_sequence);
+    Ok(to_api_response(response, format))
 }
 
 /// Handle GET /api/v1/log/scan
@@ -92,46 +70,24 @@ pub async fn handle_scan(
     let range = params.seq_range();
     let limit = params.limit.unwrap_or(32);
 
-    match state.log.scan(key.clone(), range).await {
-        Ok(mut iter) => {
-            let mut entries = Vec::new();
-            while let Some(entry) = iter.next().await.map_err(ApiError::from)? {
-                entries.push(entry);
-                if entries.len() >= limit {
-                    break;
-                }
-            }
-
-            state
-                .metrics
-                .log_scan_requests_total
-                .get_or_create(&ScanLabels {
-                    status: OperationStatus::Success,
-                })
-                .inc();
-
-            let values: Vec<Value> = entries
-                .iter()
-                .map(|e| Value {
-                    sequence: e.sequence,
-                    value: e.value.clone(),
-                })
-                .collect();
-            let response = ScanResponse::success(key, values);
-            Ok(to_api_response(response, format))
-        }
-        Err(e) => {
-            state
-                .metrics
-                .log_scan_requests_total
-                .get_or_create(&ScanLabels {
-                    status: OperationStatus::Error,
-                })
-                .inc();
-
-            Err(ApiError::from(e))
+    let mut iter = state.log.scan(key.clone(), range).await?;
+    let mut entries = Vec::new();
+    while let Some(entry) = iter.next().await.map_err(ApiError::from)? {
+        entries.push(entry);
+        if entries.len() >= limit {
+            break;
         }
     }
+
+    let values: Vec<Value> = entries
+        .iter()
+        .map(|e| Value {
+            sequence: e.sequence,
+            value: e.value.clone(),
+        })
+        .collect();
+    let response = ScanResponse::success(key, values);
+    Ok(to_api_response(response, format))
 }
 
 /// Handle GET /api/v1/log/keys
--- a/log/src/server/metrics.rs
+++ b/log/src/server/metrics.rs
@@ -6,25 +6,6 @@ use prometheus_client::metrics::counter::Counter;
 use prometheus_client::metrics::family::Family;
 use prometheus_client::registry::Registry;
 
-/// Labels for append operation metrics.
-#[derive(Clone, Debug, Hash, PartialEq, Eq, EncodeLabelSet)]
-pub struct AppendLabels {
-    pub status: OperationStatus,
-}
-
-/// Labels for scan operation metrics.
-#[derive(Clone, Debug, Hash, PartialEq, Eq, EncodeLabelSet)]
-pub struct ScanLabels {
-    pub status: OperationStatus,
-}
-
-/// Operation status for metrics.
-#[derive(Clone, Debug, Hash, PartialEq, Eq, EncodeLabelValue)]
-pub enum OperationStatus {
-    Success,
-    Error,
-}
-
 /// Labels for HTTP request metrics.
 #[derive(Clone, Debug, Hash, PartialEq, Eq, EncodeLabelSet)]
 pub struct HttpLabelsWithStatus {
@@ -68,12 +49,6 @@ pub struct Metrics {
     /// Counter of records successfully appended.
     pub log_append_records_total: Counter,
 
-    /// Counter of append requests by status.
-    pub log_append_requests_total: Family<AppendLabels, Counter>,
-
-    /// Counter of scan requests by status.
-    pub log_scan_requests_total: Family<ScanLabels, Counter>,
-
     /// Counter of HTTP requests.
     pub http_requests_total: Family<HttpLabelsWithStatus, Counter>,
 }
@@ -97,22 +72,6 @@ impl Metrics {
             log_append_records_total.clone(),
         );
 
-        // Log append requests counter
-        let log_append_requests_total = Family::<AppendLabels, Counter>::default();
-        registry.register(
-            "log_append_requests_total",
-            "Total number of append requests by status",
-            log_append_requests_total.clone(),
-        );
-
-        // Log scan requests counter
-        let log_scan_requests_total = Family::<ScanLabels, Counter>::default();
-        registry.register(
-            "log_scan_requests_total",
-            "Total number of scan requests by status",
-            log_scan_requests_total.clone(),
-        );
-
         // HTTP requests total counter
         let http_requests_total = Family::<HttpLabelsWithStatus, Counter>::default();
         registry.register(
@@ -124,8 +83,6 @@ impl Metrics {
         Self {
             registry,
             log_append_records_total,
-            log_append_requests_total,
-            log_scan_requests_total,
             http_requests_total,
         }
     }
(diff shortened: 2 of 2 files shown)


apurvam:
It's pretty cool you can just start the server and append using HTTP+JSON.. no creating topics, partitions. No cumbersome CLI tools.

log/Cargo.toml
```diff
@@ -3,12 +3,29 @@ name = "log"
...
+axum.workspace = true
 bytes.workspace = true
+clap.workspace = true
 common.workspace = true
+serde.workspace = true
+serde_json.workspace = true
 slatedb.workspace = true
 tokio.workspace = true
+tower.workspace = true
+tracing.workspace = true
```
apurvam, line 28: the tracing layer is lifted from the promql server. I _think_ it has to do with configuring log destinations, log levels, but not sure.

apurvam:
hmm. clippy passes locally 🤔

apurvam:
> hmm. clippy passes locally 🤔

I had to update main since `HEAD` had different log apis, which my local copy didn't have.

log/src/server/response.rs
```diff
@@ -0,0 +1,190 @@
...
+    /// Sequence number.
+    pub sequence: u64,
+    /// Value (UTF-8 string).
+    pub value: String,
+}
+
+impl From<&LogEntry> for ScanEntry {
+    fn from(entry: &LogEntry) -> Self {
+        Self {
+            key: String::from_utf8_lossy(&entry.key).into_owned(),
```
agavra: we'll eventually need to figure out the serde layer here so that people can use things other than utf8 keys/values

log/Cargo.toml
agavra: yup!

agavra (approved):
This LGTM! 

I do think that since this is designing a new HTTP API (unlike p8s, which is just matching an existing HTTP API) its probably worth a quick RFC for the REST API that'll double as documentation before we merge.

I also think you pointed out the interesting question which is what leverage is there? I think the biggest point is convention (again, as you noted). The actual "http server routing" part is pretty straightforward here, but having a common configuration, common CLI and common metric handling is a pretty big win from a mental overhead perspective. The ideal outcome is if we can enforce the convention via code so that the convention isn't open to drift between projects (it may actually end up in more LOC, but that's OK in my opinion). Either way, I wouldn't underestimate the importance of convention.

I mentioned this earlier too but i think when we start building out more ingest APIs we'll also be able to build more into the common service layer.

hachikuji:
This is awesome. I agree that an RFC would be helpful. Also, perhaps we should have an `http-server` feature flag?

apurvam:
> This is awesome. I agree that an RFC would be helpful. Also, perhaps we should have an `http-server` feature flag?

@hachikuji what do you mean by `http-server` feature flag? Do you mean that the server should have a feature flag that can specify it's protocol (Http, grpc, etc.)? Or something else?

hachikuji:
We can define 'http-server' as a [feature](https://doc.rust-lang.org/cargo/reference/features.html) within the log crate. Users who are just using the log API itself do not need to build the http server or take its dependencies.

apurvam:
Got it, thanks. That's a good idea. Will do!

apurvam:
RFC based on the APIs implemented here is available at: https://github.com/opendata-oss/opendata/pull/117

apurvam:
I updated the PR to reflect the merged RFC #117

log/src/server/handlers.rs
```diff
@@ -0,0 +1,203 @@
...
+/// Returns response in format matching the `Accept` header.
+pub async fn handle_scan(
+    State(state): State<AppState>,
+    headers: HeaderMap,
+    Query(params): Query<ScanParams>,
+) -> Result<ApiResponse, ApiError> {
+    let format = ResponseFormat::from_headers(&headers);
+    let key = params.key();
+    let range = params.seq_range();
+    let limit = params.limit.unwrap_or(1000);
```
agavra: nit: maybe default to something like `32` - i suspect the only time it'll be called without a limit is from some curl-like endpoint which is in development. returning 1k events feels like a lot for a default

log/src/server/handlers.rs
```diff
@@ -0,0 +1,203 @@
...
+            while let Some(entry) = iter.next().await.map_err(ApiError::from)? {
+                entries.push(entry);
+                if entries.len() >= limit {
+                    break;
+                }
+            }
+
+            state
+                .metrics
+                .log_scan_requests_total
```
agavra: we might want to consider `log_scan_events_total` and `log_scan_bytes_total` metrics (we can do this in a follow-up PR, but we should create a ticket to track it)

log/src/server/handlers.rs
```diff
@@ -0,0 +1,203 @@
...
+    match state
+        .log
+        .append_with_options(request.records, options)
+        .await
+    {
+        Ok(result) => {
+            state.metrics.log_append_records_total.inc_by(count as u64);
+            state
+                .metrics
+                .log_append_requests_total
```
agavra: we might want to consider `log_append_[events/bytes]_total` (can be done in a future PR)

log/src/server/handlers.rs
```diff
@@ -0,0 +1,203 @@
...
+
+            Err(ApiError::from(e))
+        }
+    }
+}
+
+/// Handle GET /api/v1/log/keys
+///
+/// Returns response in format matching the `Accept` header.
+pub async fn handle_list_keys(
```
agavra, line 96: missing metrics from this and the below, though honestly I think for simple "count responses by response code" we can use a middleware that does that for all of these instead of implementing it inline for each handler

log/src/server/metrics.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub log_append_records_total: Counter,
+
+    /// Counter of append requests by status.
+    pub log_append_requests_total: Family<AppendLabels, Counter>,
+
+    /// Counter of scan requests by status.
+    pub log_scan_requests_total: Family<ScanLabels, Counter>,
+
+    /// Counter of HTTP requests.
+    pub http_requests_total: Family<HttpLabelsWithStatus, Counter>,
```
agavra, line 53: related to my previous suggestion, why do we have the other totals when we have this?
```
pub struct HttpLabelsWithStatus {
    pub method: HttpMethod,
    pub endpoint: String,
    pub status: u16,
}
```
this struct seems to cover everything we need, right?

log/src/server/request.rs
```diff
@@ -0,0 +1,376 @@
...
+    }
+}
+
+/// A single record in a JSON append request.
+///
+/// Uses camelCase field names and base64-encoded bytes per RFC 0004.
+#[serde_as]
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct AppendRecordJson {
```
agavra: do we need to manually keep these in sync with the protobuf `proto.rs` implementations?

log/src/server/response.rs
```diff
@@ -0,0 +1,453 @@
...
+const CONTENT_TYPE_PROTOJSON: &str = "application/protobuf+json";
+
+/// Desired response format based on Accept header.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ResponseFormat {
+    Json,
+    Protobuf,
+}
+
+impl ResponseFormat {
```
agavra, line 27: nit: seems like this is similar to `is_protobuf_content` - I wonder if we should just have `parse_content_type` and use it for both `Accept` and `ContentType`

log/src/server/response.rs
```diff
@@ -0,0 +1,453 @@
...
+            ResponseFormat::Json => ApiResponse::Json(Json(serde_json::to_value(self).unwrap())),
+            ResponseFormat::Protobuf => ApiResponse::Protobuf(self.into_proto().encode_to_vec()),
+        }
+    }
+}
+
+/// Key wrapper for JSON serialization and deserialization.
+#[serde_as]
+#[derive(Debug, Serialize, serde::Deserialize)]
+pub struct KeyJson {
```
agavra: same question, seems unfortunate that we need to duplicate all of these with whats in `proto.rs` - would be nice if we could just implement `serde::Serialize` and `serde::Deserialize` for all of the types in `proto.rs`

log/src/server/mod.rs
```diff
@@ -0,0 +1,17 @@
+//! HTTP server for OpenData Log.
```
agavra, line 1: I agree with @hachikuji that it would be nice to gate all of this on a feature flag so people that embed the log don't need to pull in proto or axum dependencies

log/src/server/response.rs
apurvam: ugh. You are right. I had done a bunch of deduplication already, but obviously not enough. Will fix.

apurvam: Duplicates have been dropped. The handlers.rs and the request/response.rs files are much cleaner now. thanks for catching this!

apurvam: actually. I have more work to do in request.rs. Hold off on a further review for now.

apurvam: I'd like your thoughts on this. the `request.rs` has the request parameter type definitions. The `response.rs` doesn't have any type definitions, since they are all moved to `proto.rs` (since all responses can be serialized to protobuf). This feels asymmetric. 

I'm wondering what's best:

1. Rename `proto.rs` to `types.rs` and move the request parameter types there. 
2. Change `request.rs` to `params.rs` or something like that. 
3. Leave as is. 

I'm leaning toward option 1, but maybe there is a better option yet.

log/src/server/mod.rs
apurvam: done.

log/src/server/metrics.rs
apurvam: I think so.  I think the append events/append bytes/scan events/scan bytes metrics make sense. Everything else is probably covered here. I'll file issues to track the addition of those metrics, and back out the other metric definitions in this PR.

log/src/server/handlers.rs
apurvam: https://github.com/opendata-oss/opendata/issues/127

apurvam:
I'm going to go ahead and merge this. We can address any followups in a new PR.

log/src/server/response.rs
agavra: I like option 1 as well, I was also a little confused about the asymmetry between request/response/proto

PR #117, opened 2026-01-23
[LOG] RFC: HTTP API

apurvam:
## Summary

Proposes a HTTP API for OpenData-Log.

## Related Issues

Relates to #116 

## Test Plan

N/A doc only

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/log/rfcs/0004-http-apis.md
+++ b/log/rfcs/0004-http-apis.md
@@ -36,8 +36,8 @@ The API supports both binary protobuf and ProtoJSON formats, with a single proto
 truth for all message definitions. 
 
 **Content Types:**
-- Binary protobuf: `Content-Type: application/x-protobuf`
-- JSON: `Content-Type: application/json`
+- Binary protobuf: `Content-Type: application/protobuf`
+- JSON: `Content-Type: application/protobuf+json`
 
 **Field Naming:**
 - JSON field names use `lowerCamelCase` per the ProtoJSON specification
@@ -54,8 +54,12 @@ message AppendRequest {
   bool await_durable = 2;
 }
 
+message Key {
+  bytes value = 1;
+}
+
 message Record {
-  bytes key = 1;
+  Key key = 1;
   bytes value = 2;
 }
 
@@ -68,11 +72,11 @@ message AppendResponse {
 
 message ScanResponse {
   string status = 1;
-  bytes key = 2;
-  repeated Entry entries = 3;
+  Key key = 2;
+  repeated Value values = 3;
 }
 
-message Entry {
+message Value {
   uint64 sequence = 1;
   bytes value = 2;
 }
@@ -90,7 +94,13 @@ message Segment {
 
 message KeysResponse {
   string status = 1;
-  repeated bytes keys = 2;
+  repeated Key keys = 2;
+}
+
+message CountRequest {
+  Key key = 1;
+  optional uint64 start_seq = 2;
+  optional uint64 end_seq = 3;
 }
 
 message CountResponse {
@@ -116,7 +126,7 @@ let request = AppendRequest {
 };
 let binary = request.encode_to_vec();
 client.post("/api/v1/log/append")
-    .header("Content-Type", "application/x-protobuf")
+    .header("Content-Type", "application/protobuf")
     .body(binary)
     .send();
 ```
@@ -124,9 +134,9 @@ client.post("/api/v1/log/append")
 **HTTP client writing JSON:**
 ```bash
 curl -X POST http://localhost:8080/api/v1/log/append \
-  -H "Content-Type: application/json" \
+  -H "Content-Type: application/protobuf+json" \
   -d '{
-    "records": [{"key": "bXkta2V5", "value": "bXktdmFsdWU="}],
+    "records": [{"key": {"value": "bXkta2V5"}, "value": "bXktdmFsdWU="}],
     "awaitDurable": false
   }'
 ```
@@ -180,7 +190,7 @@ Request Body:
 ```json
 {
   "records": [
-    { "key": "bXkta2V5", "value": "bXktdmFsdWU=" }
+    { "key": {"value": "bXkta2V5"}, "value": "bXktdmFsdWU=" }
   ],
   "awaitDurable": false
 }
@@ -224,8 +234,8 @@ Query Parameters:
 ```json
 {
   "status": "success",
-  "key": "bXkta2V5",
-  "entries": [
+  "key": {"value": "bXkta2V5"},
+  "values": [
     { "sequence": 0, "value": "bXktdmFsdWU=" }
   ]
 }
@@ -290,7 +300,7 @@ Query Parameters:
 ```json
 {
   "status": "success",
-  "keys": ["ZXZlbnRz", "b3JkZXJz"]
+  "keys": [{"value": "ZXZlbnRz"}, {"value": "b3JkZXJz"}]
 }
 ```
 
@@ -332,9 +342,8 @@ Query Parameters:
 
 The current proposal adopts `ProtoJSON` as the JSON format. This brings support for binary payloads. It also means
 that keys and values are base64 encoded. The alternative was to allow plain JSON over HTTP without any protobuf schema.
-The latter approach was rejected because we will to support typed binary payloads for the log, and moving from a
-flexible JSON payload will make the transition to a properly typed payload with a fixed schema impossible.  
-]
+The latter approach was rejected because we want to support typed binary payloads for the log, and moving from a
+flexible JSON payload will make the transition to a properly typed payload with a fixed schema impossible.
 ## Open Questions
 
    * The proposal as presented mimics the Rust API for OpenData-Log. Should we consider APIs that are more HTTP-native? 


log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,184 @@
...
+## Design
+
+### Log HTTP Server API Summary
+
+| Endpoint | Method | Description |
+|----------|--------|-------------|
+| `/api/v1/log/append` | POST | Append records to the log |
+| `/api/v1/log/scan` | GET | Scan entries by key and sequence range |
+| `/api/v1/log/keys` | GET | List distinct keys within a segment range |
+| `/api/v1/log/segments` | GET | List segments overlapping a sequence range |
```
hachikuji, line 158: We don't have to do it now, but it might be nice to move the versioning convention into a common RFC. Similar for the /metrics endpoint which can probably be standardized.

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,184 @@
...
+| `/api/v1/log/count` | GET | Count entries for a key |
+| `/metrics` | GET | Prometheus metrics |
```
hachikuji, line 160: Not for now, but it would be cool if we had the alternative to export directly to our timeseries. It would be a nice way to leverage the platform to simplify setup. Perhaps once we have stateless ingest.

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Alternatives
+
+### Supporting binary protocols and payloads
+
+We should support binary APIs and protocols in the future, but they can complement the HTTP API. Similarly support
+for a non json payloads within the HTTP protocol can be added as a future extension. 
+
+
+## Open Questions
```
hachikuji, line 347: How do we return errors? Will we rely purely on HTTP error codes? It would be nice to have some discussion on this.

log/rfcs/0004-http-apis.md
apurvam: Agreed. Getting this type of integration right would make the platform more compelling.

log/rfcs/0004-http-apis.md
apurvam: Great point. I think HTTP ERROR codes + payload messages would be way. I'll update to include details for each api based on the implementation.

apurvam: Updated.

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,224 @@
...
+- [Apurva Mehta](https://github.com/apurvam)
+
+## Summary
+
+This RFC introduce HTTP APIs for the OpenData-Log, enabling users to append and scan events using HTTP with JSON 
+payloads.
+
+## Motivation
+
+A native HTTP API will make OpenData-Log maximally accessible to developers, enabling access to the long in any language
```
agavra:
```suggestion
A native HTTP API will make OpenData-Log maximally accessible to developers, enabling access to the log in any language
```

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,224 @@
...
+    { "key": "my-key", "value": "my-value" }
+  ],
+  "await_durable": false
+}
+```
+
+**Success Response (200):**
+
+```json
+{ "status": "success", "records_appended": 1 }
```
agavra: do we want it to return the sequence numbers that were appended?

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,224 @@
...
+{
+  "status": "error",
+  "message": "Missing required parameter: key"
+}
+```
+
+### APIs 
+
+#### Append
+`POST /api/v1/log/append`
```
agavra, line 184: I'm wondering if we should have a way to specify the encoding of the key/value (utf8 or binary-base64) so that there's a way for you to use this with non-string key/values.

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,224 @@
...
+`GET /api/v1/log/segments`
+
+List segments overlapping a sequence range. Use this to discover segment boundaries before calling /keys.
+
+Query Parameters:
+
+| Param | Type | Required | Description |
+|-------|------|----------|-------------|
+| start_seq | u64 | no | Start sequence (inclusive), default: 0 |
+| end_seq | u64 | no | End sequence (exclusive), default: u64::MAX |
```
agavra, line 263: should we add a limit (with a default) to this as well?

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,224 @@
...
+{ "status": "success", "records_appended": 1 }
+```
+
+**Error Responses:**
+- `400` - Invalid request body or empty records array
+- `500` - Storage write failure
+
+#### Scan
+
+`GET /api/v1/log/scan`
```
agavra, line 219: might make sense to also support a POST version of this so you don't have to urlencode the `key`

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,224 @@
...
+| limit | usize | no | Max entries to return, default: 1000 |
+
+**Success Response (200):**
+
+```json
+{
+  "status": "success",
+  "entries": [
+    { "key": "my-key", "sequence": 0, "value": "my-value" }
+  ]
```
agavra: seems like including the `key` in every row returned is redundant, might make sense to just have it returned once at the top-level of the response

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,224 @@
...
+| limit | usize | no | Max keys to return, default: 1000 |
+
+**Success Response (200):**
+
+```json
+{
+  "status": "success",
+  "keys": [
+    { "key": "events" },
+    { "key": "orders" }
```
agavra: curious why we don't just have:
```json
"keys": ["events", "orders"]
```
Is there other metadata about the key we intend to return here?

log/rfcs/0004-http-apis.md
apurvam: This is a good idea. I’ll update.

log/rfcs/0004-http-apis.md
apurvam: Good call. Will add.

log/rfcs/0004-http-apis.md
apurvam: The only thing I can think of are sequence numbers of each key. We can keep it compact to start with and evolve it if needed.

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,224 @@
+# RFC 0004: HTTP APIs
+
+**Status**: Draft 
+
+**Authors**:
+- [Apurva Mehta](https://github.com/apurvam)
+
+## Summary
+
+This RFC introduce HTTP APIs for the OpenData-Log, enabling users to append and scan events using HTTP with JSON
```
rodesai, line 10: What http protocol version are we using here? We should just build for HTTP2.

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,224 @@
...
+will improve the out-of-the-box developer experience. It does not preclude more performant binary APIs from being added
+later.
+
+## Goals
+
+- Add a basic HTTP API with a JSON payload
+
+## Non-Goals
+
+- Supporting binary keys and value via HTTP.
```
rodesai: might be good to call out streaming here (e.g. pushing a stream of log entries, or subscribing to the log as a stream) as something we want to support but are omitting for now

log/rfcs/0004-http-apis.md
apurvam: Makes sense. I'll call it out.

log/rfcs/0004-http-apis.md
apurvam: good call.

rodesai: this seems like a pretty major constraint for the log. did you look at non-json options for the request/response? We could do something like grpc which sends protobufs over HTTP2

log/rfcs/0004-http-apis.md
agavra: For this one i wonder if keeping your structure makes sense because it's evolvable without a backwards incompatible change. if we return only keys we can't evovle it easily.

log/rfcs/0004-http-apis.md
apurvam: I agree that it's a pretty big constraint. A very rough idea in my head is that we should be able to support multiple IO modes: 

1. HTTP/Json : great for development / demo / learning the ropes. Simple to implement.
2. GRPC: interoperable protocol, performant payload. Could be used for production. 
3. Other / custom protocols : high perf use cases. 

A fourth option is to specify some sort of key/value serde in the HTTP/JSON API. then the keys and values could be base64 encoded strings.

log/rfcs/0004-http-apis.md
apurvam: actually @hachikuji I was looking at the code. write now the [append method](https://github.com/opendata-oss/opendata/blob/main/log/src/log.rs#L244) and it doesn't return sequences. Probably the best thing from an API perspective would be to return the `base_sequence` in the response. Can the user count on the end sequence to be `base_sequence + records_appended`  ?

apurvam:
Hello all, I updated the RFC to support ProtoJson, as well as binary protobuf payloads. The keys and values are now `Bytes` and will be Base64 encoded in JSON payloads according to the protobuf spec. 

I stacked the implementation PR #116 on top of this, and updated it to support ProtoJson and protobuf, with extensive integration tests for both formats. PTAL.

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,346 @@
...
+
+#### Proto Schema
+
+```protobuf
+syntax = "proto3";
+package opendata.log.v1;
+
+message AppendRequest {
+  repeated Record records = 1;
+  bool await_durable = 2;
```
agavra, line 54: this is good for V1 but I just had the thought that we probably want to figure out how we want to handle "await durable" across all the systems. I suspect there might be a few patterns we can discuss (e.g. return a token that allows you to follow up with a different request to await it). that way there can be multiple levels of durability etc... it's more in line with github.com/slatedb/slatedb/issues/1138

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,346 @@
...
+  uint64 end_sequence = 4;
+}
+
+message ScanResponse {
+  string status = 1;
+  bytes key = 2;
+  repeated Entry entries = 3;
+}
+
+message Entry {
```
agavra:
```suggestion
message Value {
```
just a nit preference, Entry/Record sound a bit too similar

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,346 @@
...
+
+message Segment {
+  uint32 id = 1;
+  uint64 start_seq = 2;
+  int64 start_time_ms = 3;
+}
+
+message KeysResponse {
+  string status = 1;
+  repeated bytes keys = 2;
```
agavra: since we're doing protobuf I think there's very little overhead to doing 

```suggestion
  repeated Key keys = 2;
```

and then adding a dedicated wrapper for `Key` so that we can add other things in the future if we want

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,346 @@
...
+}
+
+message CountResponse {
```
agavra, line 106: missing `CountRequeset`?

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,346 @@
...
+
+The HTTP server will use HTTP/2 as the transport protocol.
+
+### Message Format
+
+The API supports both binary protobuf and ProtoJSON formats, with a single protobuf schema serving as the source of 
+truth for all message definitions. 
+
+**Content Types:**
+- Binary protobuf: `Content-Type: application/x-protobuf`
```
rodesai: I think `application/protobuf` and `application/protobuf+json` are standard here: https://protobuf.dev/reference/protobuf/mime-types/

log/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,346 @@
...
+- `400` - Missing required `key` parameter or invalid sequence range
+- `500` - Storage read failure
+
+## Alternatives
+
+### Supporting plain JSON over HTTP
+
+The current proposal adopts `ProtoJSON` as the JSON format. This brings support for binary payloads. It also means
+that keys and values are base64 encoded. The alternative was to allow plain JSON over HTTP without any protobuf schema.
+The latter approach was rejected because we will to support typed binary payloads for the log, and moving from a
```
rodesai: nit: we s/will/want/ to support

log/rfcs/0004-http-apis.md
apurvam: I made this change, but it does mean that there is an additional layer of nesting that makes the json representation look a little more clunky. Eg. the key listing API now returns: 

```json
{
  "status": "success",
  "keys": [{"value": "ZXZlbnRz"}, {"value": "b3JkZXJz"}]
}
```

instead of 

```json
{
  "status": "success",
  "keys": ["ZXZlbnRz", "b3JkZXJz"]
}
```

log/rfcs/0004-http-apis.md
apurvam: This is addressed with the use of protobufs and protojson.

log/rfcs/0004-http-apis.md
apurvam: With the base64 encoding for keys by default, adding a post method is less necessary. We can always add it later if more ergonomic.

Issue #95, opened 2026-01-17
[tsdb] serialize all ingestion to a given time bucket


rodesai:
## Description

We currently allow multiple tasks/threads to ingest into a given time bucket concurrently, which can lead to new series being written durably without the associated dictionary changes that update the ID:
    - Task A arrives which adds a new series S which gets id 10
    - Task A adds S to the in-memory dictionary in `MiniTsdb`
    - Task B arrives with writes a new sample for S
    - Task B updates the in-memory delta
    - The flush task flushes the in-memory delta before Task A updates it
    - If the process crashes here, we will have never written the dictionary update that adds S with id 10

## Steps to Reproduce

TODO

## Expected Behavior

It should not be possible for a write batch to write a series with an id that is either not added to the dictionary as part of that batch or an earlier batch.

## Proposed Solution

We should just serialize all writes to a given tsdb by either calling ingest under a lock, or forcing ingest to go through a single task. I like the second option as it allows us to naturally scale ingest by sharding the keyspace and using a task per shard to parallelize delta maintenance.

## Additional Context

Any other relevant information (logs, screenshots, etc.).

Issue #82, opened 2026-01-17
[tsdb] properly synchronize/coordinate ingestion and querying


rodesai:
## Description

There are a few issues with the way that we synchronize ingest and query, and the associated caches. These are all kind of related, and I think the solution should just be to rework the model for ingest and query tsdbs, so I'm filing this all under one issue even though we could probably break it up:

Ingest cache expiry does not flush. `MiniTsdb`s cached for ingest are evicted after 15 minutes of inactivity, but are flushed on a configurable interval. It's possible for an ingest tsdb to be dirty but inactive for 15 minutes, and then be evicted without flushing.

The ingest cache and query cache can have `MiniTsdb`s representing the same bucket. This can cause the db to return stale query results after a sequence like:
    1. Some `MiniTsdb` M for bucket B is present in the query cache but not ingest.
    2. A write comes in for B, causing a new `MiniTsdb` for B M' to be inserted into the ingest cache.
    3. M' is evicted after 15 minutes of inactivity
    4. A new read that hits B will use M, which uses a stale snapshot for the read and so returns stale query results

When loading a query reader we use different snapshots for resolving the set of buckets and loading each individual minitsdb. I don't think this causes any issues right now, but seems dangerous. All reads for a given request should use the same storage snapshot.

## Steps to Reproduce

TODO: I haven't actually written tests to repro these bugs - right now these are just some things I observed in the current implementation that we should clean up.

## Expected Behavior

Reads should be consistent with each other at least. So once a read has observed a timeseries, all later reads should observe it.
Ideally reads should observe writes either after they are flushed, and even more ideally after they are written.
All reads within a single request should observe the exact same db state
Writes should be durable once the flush interval has expired

I'm proposing we rework the tsdb layer as follows to fix these issues:

First, we’ll separate `MiniTsdb` into separate ingest and query types called `MiniTsdbIngest` and `MiniTsdbQueryReader`, respectively.

`MiniTsdbIngest` will not store any read-specific state. So it won't store any db snapshot. It just needs the current dictionary, next series id, and current delta. In the future, if we want to support reading from the current delta then `Tsdb` will pass an `Arc` to the current delta to `MiniTsdbQueryReader`. We’ll also need to support reading from deltas that are mid-flush, so we’ll need to change to a model where the delta is frozen and stored in `MiniTsdbIngest` while the flush task is flushing it (similar to SlateDB memtable).

[...]

PR #118, opened 2026-01-23
[common] RFC for write coordination

agavra:
You can see a prototype implementation here for timeseries: https://github.com/opendata-oss/opendata/commit/c982b9486b2cdb3a96f4871d5c79aa3ac1accd49 and I also tested the prototype implementation on https://github.com/responsivedev/p8s-bench and noticed no regression in ingestion performance.

I considered doing this in slatedb (turning it into a separate WAL + memtable + manual L0 flush) but this turned out to be pretty complicated so I think I'll do it this way and just use slateb as is. There is the obvious downside that data doesn't become durable until it is flushed to slatedb. we could consider disabling the slateDB WAL and using our own WAL (based on slate's WAL).

Relates to #82 and #95

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -319,6 +319,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
+name = "bencher"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "async-trait",
+ "chrono",
+ "clap",
+ "common",
+ "humantime-serde",
+ "metrics",
+ "metrics-util",
+ "serde",
+ "timeseries",
+ "tokio",
+ "toml 0.8.23",
+]
+
+[[package]]
 name = "bincode"
 version = "1.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -467,7 +485,7 @@ dependencies = [
  "num-traits",
  "serde",
  "wasm-bindgen",
- "windows-link",
+ "windows-link 0.2.1",
 ]
 
 [[package]]
@@ -566,6 +584,7 @@ dependencies = [
  "async-trait",
  "bytes",
  "criterion",
+ "futures",
  "proptest",
  "serde",
  "serde_yaml",
@@ -967,9 +986,9 @@ dependencies = [
 
 [[package]]
 name = "endian-type"
-version = "0.2.0"
+version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "869b0adbda23651a9c5c0c3d270aac9fcb52e8622a8f2b17e57802d7791962f2"
+checksum = "c34f04666d835ff5d62e058c3995147c06f42fe86ff053337632bca83e42702d"
 
 [[package]]
 name = "enum-as-inner"
@@ -1605,6 +1624,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "135b12329e5e3ce057a9f972339ea52bc954fe1e9358ef27f95e89716fbc5424"
 
 [[package]]
+name = "humantime-serde"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "57a3db5ea5923d99402c94e9feb261dc5ee9b4efa158b0315f788cf549cc200c"
+dependencies = [
+ "humantime",
+ "serde",
+]
+
+[[package]]
 name = "hyper"
 version = "1.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1681,7 +1710,7 @@ dependencies = [
  "js-sys",
  "log 0.4.29",
  "wasm-bindgen",
- "windows-core",
+ "windows-core 0.62.2",
 ]
 
 [[package]]
@@ -2017,6 +2046,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5e5032e24019045c762d3c0f28f5b6b8bbf38563a65908389bf7978758920897"
 
 [[package]]
+name = "log-bench"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "async-trait",
+ "bencher",
+ "bytes",
+ "log 0.1.0",
+ "tokio",
+]
+
+[[package]]
 name = "lrlex"
 version = "0.13.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2209,6 +2250,34 @@ dependencies = [
 ]
 
 [[package]]
+name = "metrics"
+version = "0.24.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5d5312e9ba3771cfa961b585728215e3d972c950a3eed9252aa093d6301277e8"
+dependencies = [
+ "ahash",
+ "portable-atomic",
+]
+
+[[package]]
+name = "metrics-util"
+version = "0.18.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "15b482df36c13dd1869d73d14d28cd4855fbd6cfc32294bee109908a9f4a4ed7"
+dependencies = [
+ "aho-corasick",
+ "crossbeam-epoch",
+ "crossbeam-utils",
+ "hashbrown 0.15.5",
+ "indexmap 2.12.1",
+ "metrics",
+ "ordered-float",
+ "quanta",
+ "radix_trie",
+ "sketches-ddsketch",
+]
+
+[[package]]
 name = "mime"
 version = "0.3.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2255,7 +2324,7 @@ dependencies = [
  "sysctl",
  "thiserror 1.0.69",
  "widestring",
- "windows",
+ "windows 0.48.0",
 ]
 
 [[package]]
@@ -2327,6 +2396,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ntapi"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c70f219e21142367c70c0b30c6a9e3a14d55b4d12a204d897fbec83a0363f081"
+dependencies = [
+ "winapi",
+]
+
+[[package]]
 name = "nu-ansi-term"
 version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2367,6 +2445,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5c7398b9c8b70908f6371f47ed36737907c87c52af34c268fed0bf0ceb92ead9"
 dependencies = [
  "libc",
+]
+
+[[package]]
+name = "objc2-core-foundation"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2a180dd8642fa45cdb7dd721cd4c11b1cadd4929ce112ebd8b9f5803cc79d536"
+dependencies = [
+ "bitflags 2.10.0",
+]
+
+[[package]]
+name = "objc2-io-kit"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "33fafba39597d6dc1fb709123dfa8289d39406734be322956a69f0931c73bb15"
+dependencies = [
+ "libc",
+ "objc2-core-foundation",
 ]
 
 [[package]]
@@ -2428,6 +2525,15 @@ name = "openssl-probe"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d05e27ee213611ffe7d6348b942e8f942b37114c00cc03cec254295a4a17852e"
+
+[[package]]
+name = "ordered-float"
+version = "4.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7bb71e1b3fa6ca1c61f383464aaf2bb0e2f8e772a1f01d486832464de363b951"
+dependencies = [
+ "num-traits",
+]
 
 [[package]]
 name = "ordered-multimap"
@@ -2514,7 +2620,7 @@ dependencies = [
  "libc",
  "redox_syscall",
  "smallvec",
- "windows-link",
+ "windows-link 0.2.1",
 ]
 
 [[package]]
@@ -2809,6 +2915,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "quanta"
+version = "0.12.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f3ab5a9d756f0d97bdc89019bd2e4ea098cf9cde50ee7564dde6b81ccc8f06c7"
+dependencies = [
+ "crossbeam-utils",
+ "libc",
+ "once_cell",
+ "raw-cpuid",
+ "wasi",
+ "web-sys",
+ "winapi",
+]
+
+[[package]]
 name = "quick-error"
 version = "1.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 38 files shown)


rfcs/0003-write-coordination.md
```diff
@@ -0,0 +1,386 @@
...
+pub trait Flusher: Send + Sync + 'static {
+    type Image: Send + Sync + Clone + 'static;
+    type Delta: Delta<Image = Self::Image>;
+
+    /// Flush a delta to storage and return the new image.
+    fn flush(
+        &self,
+        delta: Self::Delta,
+        image: &Self::Image,
+    ) -> impl Future<Output = Result<Self::Image>> + Send;
```
agavra: we might want to split this API into two: one to return the new image (and unblock creating the new pending delta) and one to flush to storage.

also no idea why Claude changes this to return a `Future` instead of just calling it `async` 😵‍💫

rfcs/0004-write-coordination.md
```diff
@@ -0,0 +1,415 @@
...
+- Reader/Writer coordination (this RFC sets the stage for this)
+- Distributed write coordination
+
+## Design
+
+### Components & Terminology
+
+- `Write` is an insertion of data into the system
+- `WriteQueue` is a queue that accepts `Write` as input and applies an epoch-based ordering
+- `WriteBatch` is a collection of `Write`s each stamped with an epoch
```
rodesai: is each batch given an epoch, or each write? I'm not quite following what a `WriteBatch` is. If a caller wants to atomically apply a group of writes, do they model that as one `Write` and `WriteBatch` is an internal construct, or is it exposed to the users of this interface?

rfcs/0004-write-coordination.md
```diff
@@ -0,0 +1,415 @@
...
+- `Snapshot` is a point-in-time reference to the storage state, broadcast to readers after each flush
+- `Image` is the in-memory state needed to initialize a delta (e.g., series dictionary, ID counters);
```
rodesai: is it fair to say that `Image` is the materialized result of applying the current `Delta`?

rfcs/0004-write-coordination.md
```diff
@@ -0,0 +1,415 @@
...
+the APIs provided by this RFC:
+
+```rust
+pub struct WriteBatch<W> {
+    pub writes: Vec<W>,
+    pub max_epoch: u64,
+}
+
+pub trait Delta: Send + 'static {
+    type Image: Send + Sync + Clone + 'static;
```
rodesai: why does `Image` need to be `Clone`?

rfcs/0004-write-coordination.md
```diff
@@ -0,0 +1,415 @@
...
+}
+
+pub trait Flusher: Send + Sync + 'static {
+    type Delta: Delta;
+
+    /// Flush a delta to storage and return the new snapshot.
+    fn flush(
+        &self,
+        delta: Self::Delta,
+    ) -> impl Future<Output = Result<Snapshot>> + Send;
```
rodesai: what is `Snapshot` here? Is this a type parameter or literally a slatedb snapshot?

rfcs/0003-write-coordination.md
```diff
@@ -0,0 +1,415 @@
...
+/// Handle for interacting with the write coordinator.
+pub struct WriteCoordinatorHandle<D: Delta> { /* ... */ }
+
+impl<D: Delta> WriteCoordinatorHandle<D> {
+    /// Submit a write and receive a handle to track its durability.
+    pub async fn write(&self, event: D::Write) -> Result<WriteHandle> { /* ... */ }
+    /// Request a flush up to the specified epoch (or all pending if None).
+    pub async fn flush(&self, epoch: Option<u64>) -> Result<()> { /* ... */ }
+    /// Subscribe to snapshot updates.
+    pub fn subscribe(&self) -> watch::Receiver<Arc<Snapshot>> { /* ... */ }
```
rodesai: todo: rethink this to allow for reading unflushed state

rfcs/0004-write-coordination.md
```diff
@@ -0,0 +1,415 @@
...
+
+    /// Initialize the delta with state from the image.
+    fn init(&mut self, image: &Self::Image);
+    /// Apply a batch of writes to this delta.
+    fn apply(&mut self, batch: WriteBatch<Self::Write>) -> Result<()>;
+    /// Estimate the memory size of this delta for backpressure.
+    fn estimate_size(&self) -> usize;
+    /// Extract state needed for the next delta's initialization.
+    /// Called before flush to enable non-blocking writes.
+    fn fork_image(&self) -> Self::Image;
```
rodesai, line 124: IIUC this is called before flush to essentially freeze the current delta and extract the `Image` instance used to initialze a new delta. I'd consider having this consume `self` and return a `(Self::ImmutableDelta, Self::Image)` where `ImmutableDelta` just has an `estimate_size` fn that can be used for backpressure.

rfcs/0004-write-coordination.md
```diff
@@ -0,0 +1,441 @@
...
+                     └───────────────────────┘
+                                 │
+                                 ▼
+┌Write Coordinator────────────────────────────────────────────────┐
+│ ┌─────────────────────────────┐ ┌──────────────┐┌──────────────┐│             ┌───────────┐
+│ │        select! loop         │ │ EpochTracker ││   Flusher    │├FlushEvent──▶│  Reader   │
+│ ├─────────────────────────────┤ └──────────────┘└───────▲──────┘│             └───────────┘
+│ │                             │         ┌───────────────┤       │
+│ │                             │ ┌Delta──┴──────┐┌Image──┴──────┐│
+│ │1 ─► handle flush complete   │ │ ┌──────────┐ ││              ││
```
rodesai, line 77: presumably this is a separate event because flushes are done in the background so flushing and delta building can happen concurrently? will we ensure that there are no concurrent flushes?

rfcs/0004-write-coordination.md
```diff
@@ -0,0 +1,441 @@
...
+|-----------|-------------------------------------------------------------------------|
+| `applied` | Highest epoch that has been applied to the pending delta                |
+| `flushed` | Highest epoch reflected in the current image (delta applied to SlateDB) |
+| `durable` | Highest epoch persisted to object storage (SlateDB WAL flush complete)  |
+
+### Implementation
+
+The coordinator runs as a single-threaded async loop that processes commands and manages flushes.
+
+1. **Ordering**: all writes are serialized through a single `mpsc` channel (`WriteQueue`). Epochs
```
rodesai: One thing I'm trying to think through is how this approach can be extended to support multiple "input" channels. The immediate need for this is for the vector index maintenance stuff. In that case I want to be able do something like this:
    - pass in a command to the write coordinator to split a given centroid c into c_0 and c_1. This is just a lightweight split that creates the new centroids and marks the old one as "draining".
    - wait until the resulting changes are readable
    - read vector c and neighbours to compute reassignments.
I don't want to treat this as a write on the write channel because I want these commands to take priority, and it probably doesn't make sense to log them (if/when we write the WAL, though maybe writing the WAL can be upstream of the coordinator as described above). But it is useful to be able to sequence so the mutations respect snapshot reads. But we can't really do that if the sequence number / epoch is set on enqueue.

But generally there is perhaps a pattern of internal data structure maintenance that needs to be sequenced in the series of writes to storage, but is independent of the the logical writes into the system.

rodesai (commented):
> we could consider disabling the slateDB WAL and using our own WAL (based on slate's WAL).

I think this makes sense eventually. The write coordinator feels a lot like the WAL compactor from some of our earlier decomposed slatedb discussions. If we do this, I'm wondering if it makes sense to have the WAL writer be a separate entity that you can plug in in front of the write coordinator. That would give you a lot of freedom in how you compose things - e.g. you could have stateless writers that just append new WAL files and run the write coordinator separately, feeding it the data from the WAL.

rfcs/0004-write-coordination.md
agavra: Fair enough, I evolved the RFC a little bit and this is somewhat of an artifact of that. I intended `Write` to be 1:1 with an epoch and then `WriteBatch` is basically just `Vec<StampedWrite>`, and it's what the main loop polls from the queue in the case that many writes have backed up while something else was happening.

I'll just remove this concept from the RFC, it's an implementation detail whether or not we decide to give it a special type.

rfcs/0004-write-coordination.md
agavra: Almost, the initial `Image_0` is initialized from the snapshot (e.g. build up the in-memory map of fingerprint -> ids in timeseries). Then after that `Image_1` is the result of applying `Delta_0` to `Image_0`. I'll make that clear in the RFC.

rfcs/0004-write-coordination.md
agavra: Literally a SlateDB Snapshot. I think we call it `StorageRead`. I'll update the RFC to reflect that.

rfcs/0004-write-coordination.md
agavra: In order to support a non-blocking `flush` we need have `fork_image` on delta:
```
2. **Non-Blocking Flush**: when a flush is triggered:
   1. `fork_image()` is called on the current delta to extract state for the next delta
   2. The current delta is cloned for subscribers
   3. A new delta is created and initialized from the forked image
   4. The old delta is passed to the flusher (ownership transferred)
   5. Writes continue to the new delta while the flush runs in the background
   6. When flush completes, a `FlushEvent` is broadcast with the snapshot, cloned delta, and epoch range
```
It's what's needed to support ingesting new data while keeping a snapshot of the old Image for flushing data.

I'll double check to see whether or not access to the Image during flush is necessary. If we can flush everything we need just with access to the Delta then we don't need it and we can just reuse the same Image across flushes.

rfcs/0003-write-coordination.md
agavra: yeah, this is in "future extensions" suggested at the bottom

rfcs/0004-write-coordination.md
agavra: yes and yes 👍 

```
3. **Flush Triggers**: flushes can be triggered manually via `flush()` or based on conditions
   evaluated on a timer. Flush requests are queued if a flush is in progress. A call to
   `flush(epoch)` is a no-op if `epoch` ≤ the already-flushed epoch.
```

rfcs/0003-write-coordination.md
rodesai: lol my bad this was a todo for myself that I left while reviewing 😆

rfcs/0004-write-coordination.md
agavra: I'm not fully sure I understand - why wouldn't we want to log them, and why do we want them to take priority? In other words, whats wrong with the naive approach? (and then you can use `await_durable(Flushed)` to wait until they are readable)

rodesai: I want them to take priority because we'll need some kind of backpressure mechanism. If we're writing vectors faster than we can maintain the index then we need some way to pause ingest while still applying splits/merges.

I think not logging them is just an optimization. There's no real need to log them. The caller blocks on them being applied so it can then observe the old centroid once it knows know more vectors will be written to it. And it doesn't need to rely on the log to recover after a crash.

We should be able to do this by just assigning the epoch from the coordinator task. The only limitation this adds is that the call into write needs to block on the coordinator task picking up the write and assigning the epoch.

agavra: Would adding an additional priority channel be sufficient? a bit of a brute force approach but since im thinking of implementing the coordinator loop as a tokio `select!` with `biased` anyway this would be an easy way to have a priority channel. we can implement selective logging outside of this RFC

rfcs/0004-write-coordination.md
agavra: It doesn't technically need `Clone` but we still need to copy the image.

*(Response written by Claude)*

After tracing through the flow, `Image` doesn't actually need `Clone`:

1. `Image_0` loaded from storage (initial bootstrap)
2. `delta.init(&image)` takes a reference
3. `fork_image()` creates a *new* Image from the Delta
4. Coordinator stores one Image at a time, replaces it after each fork

No cloning of Image is required — the coordinator only passes `&Image` references. The `Clone` bound was vestigial.

Removed `Clone` from `Image` in the RFC. `Delta` still needs `Clone` for `FlushEvent`.

rfcs/0004-write-coordination.md
hachikuji: It sounds like the problem is that ordering is determined prior to queuing? Could we assign an unordered ID at that point for tracking and defer epoch assignment to dequeue time?

rodesai: Yep exactly. Using an unordered ID would work. We could also use a per-write channel like slatedb uses and trigger it once the epoch associated with the write reaches the desired level.

> Would adding an additional priority channel be sufficient?

It's sufficient for providing another channel for the higher-priority commands. The problem is that those commands transform the internal state, and those transformations should really be tied to the snapshot used by the readers to observe the state. So those transformations need an epoch to associate with.

agavra: Maybe I'm misunderstanding the unordered writes, but it's really nice to know that if I do the following `foo` is written before `bar`. Would I lose the ability to do that?
```rs
write(foo).await;
write(bar).await;
```

> We could also use a per-write channel like slatedb uses and trigger it once the epoch associated with the write reaches the desired level.

not sure I understood this either. What do you mean by "trigger it"?

agavra: I guess we get the write ordering because `.await` wouldn't return until its on the queue? so from a single thread I get ordering?

hachikuji: One way to model it might be to have separate input streams. Each stream is ordered sequentially, but the streams may be interleaved. So one stream might correspond to externally generated writes and one for internally generated writes.

agavra: OK, so `WriteHandle` would give you:
```rs
impl WriteHandle {
    /// Returns the id assigned to this write.
    pub fn id(&self) -> u64 { /* ... */ }
    /// Returns the epoch assigned to this write
    pub async fn epoch() -> u64 { /* ... */}
    /// Wait until the write reaches the specified durability level.
    pub async fn wait(&self, durability: Durability) -> Result<()> { /* ... */ }
}
```
And you could choose to just `wait` on the id _or_ you could get the `epoch` as soon as its read from the queue so you can wait on epochs in bulk.

For now I'll keep it simple and say you only get ordering within the thread if you await. If we have need for bulk writes + ordering we can introduces multiple write streams.

How does that sound?

rodesai: > I guess we get the write ordering because .await wouldn't return until its on the queue? so from a single thread I get ordering?

Yep. This is exactly how slatedb's writes work btw. You're guaranteed ordering of sequential calls to `put` because all puts are funneled through a single channel, even though they aren't assigned sequence numbers until the write task picks them up.

> One way to model it might be to have separate input streams. Each stream is ordered sequentially, but the streams may be interleaved. So one stream might correspond to externally generated writes and one for internally generated writes.

Yep, this is what I had in mind. The coordinator can read from different channels, and maintain ordering within a channel even if operations from different channels are interleaved. This seems sufficient to get the semantics we want.

> How does that sound?

That works. Here's the channel based idea I had in mind so we wouldn't need to include another id to the mix:

```rust
// some type that holds a reference to where the current durable/flushed/applied epochs are bookkept
// by the write coordinator
struct FlushWatchers {
}

impl FlushWatchers {
    async fn wait(&self, epoch: u64, durability: Durability) {
        // wait on tokio::watch instances for current durable/flushed/applied epochs
        todo!()
    }
}

enum WriteHandle {
    Queued(tokio::sync::oneshot::Receiver<Result<(u64, FlushWatchers), String>>),
    Failed(String),
    Sequenced {
        epoch: u64,
        watchers: FlushWatchers
    }
}

impl WriteHandle {
    // called when enqueuing a write, the handle is returned to the caller, the tx is included
    // in the enqueued write msg
    fn new() -> (Self, tokio::sync::oneshot::Sender<Result<(u64, FlushWatchers), String>>) {
        let (tx, rx) = tokio::sync::oneshot::channel();
        (Self::Queued(rx), tx)
    }

    fn epoch(&self) -> Option<u64> {
        match self {
            WriteHandle::Queued(_) => None,
            WriteHandle::Failed(_) => None,
            WriteHandle::Sequenced { epoch, .. } => Some(*epoch),
        }
    }

    async fn wait(&mut self, durability: Durability) -> Result<(), String> {
        loop {
            match self {
                WriteHandle::Queued(chan) => {
                    let result = chan.await.map_err(|_| "hung up".to_string())?;
                    match result {
                        Ok((epoch, watchers)) => {
                            *self = WriteHandle::Sequenced { epoch, watchers };
                        },
                        Err(e) => {
                            *self = WriteHandle::Failed(e.clone());
                            return Err(e);
                        }
                    }
                },
                WriteHandle::Failed(err) => return Err(err.clone()),
                WriteHandle::Sequenced { epoch, watchers } => {
                    watchers.wait(*epoch, durability).await;
                }
            };
        }
    }
}
```

agavra: I think I prefer adding the extra id for now, it seems simpler to me to track a mapping than handling a state machine but I'll give it more thought when it gets to implementation.

agavra: argh, looks like codex caught a real issue with introducing another id:

```
 The `id -> epoch` map has no cleanup strategy. If entries aren’t
  removed after `epoch()` resolves (or after a write is fully durable), it will grow unbounded over long runs. Suggest specifying
  when entries are removed, or replace the shared map with per-handle channels to avoid global state.
```

I don't think there's a super clean solution to this, so I might end up going with the multi-channel suggestion you had @rodesai

rodesai (approved):
LGTM!

PR #120, opened 2026-01-23
fix(tsdb): support unary promql operators from tsdb

### LINKED ISSUE #80: [tsdb] support unary promql operators from tsdb
Support unary promql operators from tsdb. Currently the only supported operator is negation (`-`).

Rohan-flutterint:
## Summary

This PR adds support for unary expressions (`-expr`) in the PromQL evaluator.

Unary negation is now correctly evaluated for:
- Scalar values (e.g. `-5`)
- Instant vectors (e.g. `-metric_name`)
- Range vectors (e.g. `-metric_name[5m]`)

This brings the evaluator closer to Prometheus-compatible PromQL behavior.

---

## Related Issues

Fixes #80

---

## Test Plan

Added and verified new tests for unary expression evaluation:

- `should_evaluate_unary_scalar_literal`
- `should_evaluate_unary_vector`

Tested using:
- `cargo test`
- Targeted test runs:
  - `cargo test should_evaluate_unary_scalar_literal`
  - `cargo test should_evaluate_unary_vector`

All existing tests pass successfully.

---

## Checklist

- [x] Tests added/updated
- [ ] Documentation updated (not applicable)

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -411,7 +411,20 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
                         .evaluate_expr(&u.expr, start, end, interval, lookback_delta)
                         .await?;
 
-                    Ok(self.apply_unary_negation(inner))
+                    match inner {
+                        ExprResult::Scalar(v) => Ok(ExprResult::Scalar(-v)),
+
+                        ExprResult::InstantVector(mut samples) => {
+                            for s in &mut samples {
+                                s.value = -s.value;
+                            }
+                            Ok(ExprResult::InstantVector(samples))
+                        }
+
+                        ExprResult::RangeVector(_) => Err(EvaluationError::InternalError(
+                            "Unary operators are not supported for range vectors".to_string(),
+                        )),
+                    }
                 };
                 Box::pin(fut)
             }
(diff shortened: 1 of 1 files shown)


timeseries/src/promql/evaluator.rs
```diff
@@ -900,6 +907,28 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
+            ExprResult::Scalar(v) => ExprResult::Scalar(-v),
+
+            ExprResult::InstantVector(mut samples) => {
+                for s in &mut samples {
+                    s.value = -s.value;
+                }
+                ExprResult::InstantVector(samples)
+            }
+
+            ExprResult::RangeVector(mut series) => {
```
rodesai, line 934: the spec restricts unary operators to scalars and instant vectors, so we should return an error in this case: https://prometheus.io/docs/prometheus/latest/querying/operators/

rohankumardubey: Thanks for pointing this out, you're absolutely right. I will update the implementation to return an error when a unary operator is applied to a range vector. Also, I will be adding a test to ensure this behavior is enforced.

rohankumardubey:
@rodesai Can you please take a look at this?

PR #121, opened 2026-01-23
fix(tsdb): correctly parse match[] query parameters for label value

### LINKED ISSUE #14: tsdb p8s server match[] parameters cannot be parsed over curl

Rohan-flutterint:
## Summary

What does this PR do?
This PR fixes parsing of Prometheus-style match[] query parameters for the
/api/v1/label/{name}/values endpoint.

Axum’s Query extractor cannot correctly deserialize repeated match[] parameters
in URL-encoded queries, causing valid requests to fail before reaching the handler.
The handler now parses match[] parameters directly from the raw query string while
preserving existing support for start, end, and limit.

## Related Issues
Fixes #14

## Test Plan

How was this tested?

- Reproduced the issue locally using curl as described in the issue
- Verified single and multiple match[] parameters work correctly
- Verified behavior with optional start, end, and limit parameters
- Added unit tests around the label values query parsing logic to cover single and multiple `match[]` parameters as well as optional params.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/mod.rs
+++ b/timeseries/src/promql/mod.rs
@@ -7,6 +7,7 @@ pub(crate) mod metrics;
 mod middleware;
 pub(crate) mod openmetrics;
 mod parser;
+pub mod query_parser;
 #[cfg(feature = "remote-write")]
 pub(crate) mod remote_write;
 mod request;
--- a/timeseries/src/promql/query_parser.rs
+++ b/timeseries/src/promql/query_parser.rs
@@ -1,0 +1,82 @@
+use serde::de::DeserializeOwned;
+
+/// This helper enables reuse of existing `*Params` structs across endpoints
+/// without duplicating parsing logic for repeated query keys.
+pub fn parse_query_with_repeated<T>(
+    raw_query: &str,
+    repeated_key: &str,
+) -> Result<(Vec<String>, T), serde_urlencoded::de::Error>
+where
+    T: DeserializeOwned,
+{
+    // Parse raw query into key-value pairs
+    let params: Vec<(String, String)> = serde_urlencoded::from_str(raw_query).unwrap_or_default();
+
+    // Collect repeated params (e.g. match[])
+    let repeated: Vec<String> = params
+        .iter()
+        .filter(|(k, _)| k == repeated_key)
+        .map(|(_, v)| v.clone())
+        .collect();
+
+    // Filter out repeated params
+    let remaining: Vec<(String, String)> = params
+        .into_iter()
+        .filter(|(k, _)| k != repeated_key)
+        .collect();
+
+    // Deserialize remaining params into T
+    let remaining_query = serde_urlencoded::to_string(remaining).unwrap_or_default();
+    let other_params: T = serde_urlencoded::from_str(&remaining_query)?;
+
+    Ok((repeated, other_params))
+}
+
+// TODO!
+// Temporary Tests for the SeriesParams to test the helper function 'parse_query_with_repeated'
+// is working as expected, will move this text back to 'SeriesParams' handler later.
+#[cfg(test)]
+mod series_tests {
+    use crate::promql::query_parser::parse_query_with_repeated;
+    use crate::promql::request::SeriesParams;
+
+    #[test]
+    fn parses_series_params_with_match_and_options() {
+        let raw = concat!(
+            "match%5B%5D=%7Bjob%3D%22api%22%7D&",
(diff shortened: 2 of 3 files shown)


rodesai:
@Rohan-flutterint is this the same as https://github.com/opendata-oss/opendata/pull/109 ? Why close that one and reopen?

rohankumardubey:
> @Rohan-flutterint is this the same as #109 ? Why close that one and reopen?

Yes, this is the same as #109. My fork was deleted on my end, which automatically closed the original PR, so I recreated the fork and reopened the PR.

timeseries/src/promql/server.rs
```diff
@@ -278,17 +294,143 @@ async fn handle_labels(
...
 /// Handle /api/v1/label/{name}/values
+///
+/// NOTE: Axum’s `Query` extractor cannot correctly deserialize repeated `match[]`
+/// parameters in URL-encoded queries. This handler parses `match[]` manually from
+/// the raw query string to preserve Prometheus-compatible behavior.
 async fn handle_label_values(
     State(state): State<AppState>,
     Path(name): Path<String>,
-    Query(params): Query<LabelValuesParams>,
+    RawQuery(raw_query): RawQuery,
```
agavra, line 291: I think we need a more generic solution, this only helps us with label_values but repeated matches can happen in any of the requests.

Also if you use AI to generate any of these PRs please disclose the model and the manual testing/verification done to confirm the behavior works as intended.

rohankumardubey: I agree a more generic solution would make sense here. I started with `label_values` since that’s where the issue surfaced, but I’ll explore factoring this into a reusable helper so other endpoints with repeated `match[]` parameters can benefit as well. 

Regarding the AI part, ChatGPT 5.2 was used mainly as a reference and for sanity-checking ideas, but the implementation and validation of the final design, coding, and testing were done manually.

agavra: Makes sense, totally fine to use AI (I do it a bunch) but it's helpful for us to know what model / techniques were used so that we can understand a little bit better before going into the PR review.

rohankumardubey: Thanks. Just to clarify, would you prefer that I introduce the generic parsing approach as part of this PR, or should I follow up with a separate change once this fix is merged?

agavra: I think it should be part of this PR so we can see the big picture. An example of where it's relevant is `QueryOptions`, it probably makes more sense to use the existing classes like `LabelValuesParams` and I want to make sure that we can do that in a way that works for all of the endpoints

rohankumardubey: That makes sense. I’ll refactor the current implementation to make it work for all endpoints.

rohankumardubey:
@agavra Made the changes for the generic parsing. Can you take a look?

timeseries/src/promql/mod.rs
```diff
@@ -7,6 +7,7 @@ pub(crate) mod metrics;
 mod middleware;
 pub(crate) mod openmetrics;
 mod parser;
+pub mod query_parser;
```
agavra, line 10: this doesn't need to be pub

timeseries/src/promql/query_parser.rs
```diff
@@ -0,0 +1,82 @@
+use serde::de::DeserializeOwned;
+
+/// This helper enables reuse of existing `*Params` structs across endpoints
+/// without duplicating parsing logic for repeated query keys.
+pub fn parse_query_with_repeated<T>(
```
agavra, line 5: this feels a little brittle, it requires you to know the repeated keys up front and assumes non of the other keys are repeated

timeseries/src/promql/server.rs
```diff
@@ -279,16 +281,109 @@ async fn handle_labels(
...
 ) -> Result<Json<LabelValuesResponse>, ApiError> {
+    let raw_query = raw_query.unwrap_or_default();
+
+    let (matches, mut params): (Vec<String>, LabelValuesParams) =
+        parse_query_with_repeated(&raw_query, "match[]")
+            .map_err(|e| Error::InvalidInput(format!("Failed to parse query: {}", e)))?;
```
agavra, line 297: it would be nice to implement `From RawQuery` for `LabelValuesParams` (and the others) so that we can make sure we can parse them all if necessary

agavra:
Closing this in favor of #203 which fixes the issue using axum's built-in Query/Form extractors

PR #122, opened 2026-01-24
Bencher prototype

hachikuji:
This patch implements a first cut of a really basic bencher. See the example benchmark in log/bench/src/ingest.rs for usage.

### DIFF
--- a/bencher/src/bench.rs
+++ b/bencher/src/bench.rs
@@ -9,6 +9,7 @@
 use std::sync::Arc;
 use std::time::Instant;
 
+use common::display::format_number;
 use metrics::Recorder;
 use timeseries::{Label, MetricType, Sample, Series};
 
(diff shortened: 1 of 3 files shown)


hachikuji:
There are a few unsatisfying details I'm still working on. First, I dislike the tight coupling of the benchmark parameters in the benchmark itself. It would be nice to separate that so that the values could be overridden simply through config or CLI parameters. The second thing I dislike is the approach for specifying the run duration. It would be nice to similarly allow that to be configurable. I'll push another patch shortly.

hachikuji:
I tested this locally with the prototype ingest benchmark. This is what it looks like:
```
cargo run -p log-bench -- -c config.toml                                                                              
  [benchmark=ingest, commit=a84f348f9b8d94d072b17bee3cf3e7e6f274f668, branch=bencher-impl, batch_size=1, key_length=16, num_keys=10, val
ue_size=256]                                                                                                                            
    throughput_ops    34.82K                                        
    throughput_bytes  9.47M                                                                                                             
    elapsed_ms        5.00K
  [benchmark=ingest, commit=a84f348f9b8d94d072b17bee3cf3e7e6f274f668, branch=bencher-impl, batch_size=10, key_length=16, num_keys=10, va
lue_size=256]
    throughput_ops    100.99K
    throughput_bytes  27.47M
    elapsed_ms        5.00K
...
```
Still lots of room for improvement. But maybe a useful starting point.

agavra:
That's some decent perf numbers! Very few use cases went that much beyond 27MB/s (I'm assuming these are per second numbers). I assume this is on your laptop + in-memory/local object storage? 

I'll skim the PR now, but don't feel the need to block on me if I don't finish the review tonight.

bencher/src/lib.rs
```diff
@@ -0,0 +1,828 @@
...
+    /// Create a new Bencher with the given configuration, benchmark name, and static labels.
+    async fn new(config: Config, name: &str, labels: Vec<Label>) -> anyhow::Result<Self> {
+        let reporter: Option<Arc<dyn Reporter>> = match &config.reporter {
+            Some(reporter_config) => {
+                let ts_config = timeseries::Config {
+                    storage: reporter_config.to_storage_config(),
+                    ..Default::default()
+                };
+                let ts = timeseries::TimeSeries::open(ts_config).await?;
+                Some(Arc::new(TimeSeriesReporter::new(ts)))
```
agavra, line 106: ha! this is cool dogfooding. hopefully it doesn't interfere at all with the experiment (not sure how the object store libraries work and if this may cause some sort of contention)

bencher/src/lib.rs
```diff
@@ -0,0 +1,828 @@
...
+    /// Access the benchmark specification.
+    pub fn spec(&self) -> &BenchSpec {
+        &self.spec
+    }
+
+    /// Get a counter handle for recording incremental values.
+    ///
+    /// The counter is identified by name and will include the bench's labels
+    /// when exported.
+    pub fn counter(&self, name: &'static str) -> Counter {
```
agavra: nit: we can do this another time but I feel like it'd be nice to break up this file into various modules instead of putting it all into `lib.rs`

bencher/src/lib.rs
```diff
@@ -0,0 +1,828 @@
...
+}
+
+impl Default for Summary {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Format a number for display with thousands separators and appropriate precision.
+fn format_number(value: f64) -> String {
```
agavra: I wonder if this would be useful in `common`, may not be the only place we need to format numbers like this

common/src/storage/factory.rs
```diff
@@ -7,13 +7,46 @@ use std::sync::Arc;
...
 use super::config::{ObjectStoreConfig, StorageConfig};
 use super::in_memory::InMemoryStorage;
 use super::slate::SlateDbStorage;
 use super::{MergeOperator, Storage, StorageError, StorageResult};
 
+/// Creates an object store from configuration without initializing SlateDB.
+///
+/// This is useful for cleanup operations where you need to access the object store
+/// after the database has been closed.
+pub fn create_object_store(config: &ObjectStoreConfig) -> StorageResult<Arc<dyn ObjectStore>> {
```
agavra, line 21: should we reuse this below in `create_slatedb_storage`?

common/src/storage/util.rs
```diff
@@ -0,0 +1,63 @@
...
+
+/// Deletes all data for the given storage configuration.
+///
+/// For SlateDB storage, this deletes all objects under the configured path prefix.
+/// For InMemory storage, this is a no-op (memory is released when storage is dropped).
+///
+/// Call this after the storage/database has been closed.
+pub async fn delete(config: &StorageConfig) -> StorageResult<()> {
+    match config {
+        StorageConfig::InMemory => {
```
agavra, line 21: (totally separate musing) I wonder if we should get rid of this and just use slatedb::in_memory for all testing/in memory use caes

log/bench/src/ingest.rs
```diff
@@ -0,0 +1,135 @@
...
+        let mut key_idx = 0;
+
+        while runner.keep_running() {
+            let records: Vec<Record> = (0..batch_size)
+                .map(|_| {
+                    let key = keys[key_idx % keys.len()].clone();
+                    key_idx += 1;
+                    Record {
+                        key,
+                        value: value.clone(),
```
agavra, line 97: we can do this in a follow-up, but I'm curious to test throughput with different value templates so that we don't get ridiculously good compression

bencher/src/lib.rs
hachikuji: Yeah, it is an optimistic addition. Not sure if it will be useful yet, but it was easy to add.

common/src/storage/util.rs
hachikuji: Probably makes sense. It would be nice to have something for test cases which behaved a bit close to a real object storage. Not necessarily introducing latency, but perhaps giving control to the test case to define when operations complete. Maybe I'll try sketching something up.

log/bench/src/ingest.rs
hachikuji: Yeah, I didn't put much thought into this bench. It was just a starter to get the end-to-end flow. I'll work on this next.

PR #123, opened 2026-01-24
some tweaks to the vector storage rfc

rodesai:
makes a few changes to the vector storage design
- changes the value of the postings to include the vector ids, and to serialize it as a simple array.
- don't split the vector data and metadata.
- track the deletion vector in its own key rather than as centroid 0
- drop references to the workings of LIRE. This will be addressed in a future PR

### DIFF
--- a/vector/rfcs/0001-storage.md
+++ b/vector/rfcs/0001-storage.md
@@ -755,7 +755,9 @@ Future RFCs will address:
 
 | Date       | Description                                                          |
 |------------|----------------------------------------------------------------------|
-| 2025-01-07 | Initial draft                                                        |
+| 2026-01-07 | Initial draft                                                        |
+| 2026-01-24 | Store full vectors in postings. Store all attributes in `VectorData` |
+|            | and drop `VectorMeta`                                                |
 
 ## References
 


vector/rfcs/0001-storage.md
```diff
@@ -216,8 +252,7 @@ with future schema changes that may require sub-type discrimination.
 - `Vector<D>`: `D` elements of `f32`, where `D` is the dimensionality stored in collection metadata.
   Total size is `D * 4` bytes.
 - `Array<T>`: `count: u16` followed by `count` serialized elements of type `T`.
-- `FixedElementArray<T>`: Serialized elements back-to-back with no count prefix; length derived from
-  buffer size divided by element size.
+- `FixedElementArray<T>`: Serialized elements back-to-back with no count prefix;
```
rodesai, line 263: I simplified this to just an array with no count prefix. The elements can have variable length as long as the buffer's end point is known.

vector/rfcs/0001-storage.md
```diff
@@ -377,29 +445,11 @@ boundary classification are defined in the LIRE rebalancing RFC.
 └─────────┴─────────────┴──────────────┘
 ```
 
-**Centroid ID Addressing:**
```
rodesai, line 380: I just moved this to the centroid chunks section. Felt out of place here.

vector/rfcs/0001-storage.md
```diff
@@ -75,8 +75,8 @@ index structures are stored as key-value pairs in the LSM tree.
 │   │   Centroids:     [chunk 0] [chunk 1] ... [chunk N]  │   │
 │   │                  (loaded into HNSW for navigation)  │   │
 │   │                                                     │   │
-│   │   Posting Lists: centroid_id → RoaringTreemap       │   │
-│   │                  (vector IDs per cluster)           │   │
+│   │   Posting Lists: centroid_id →                      │   │
+│   │                  [(vector_id, f32[dimensions])]     │   │
```
agavra, line 79: I wonder if we want it to be be `[vector_id],[f32[dimensions]]` so that we can load just the vector ids if we want to apply filtering first (and then we can pull up just the vectors we want based on indexes). Downside is that we can't just append to the end, so maybe this is better. Just worth considering I can go either way.

vector/rfcs/0001-storage.md
```diff
@@ -85,8 +85,9 @@ index structures are stored as key-value pairs in the LSM tree.
 │   │                  Vector Storage                     │   │
 │   │                                                     │   │
 │   │   IdDictionary:  external_id → internal vector_id   │   │
-│   │   VectorData:    vector_id → f32[dimensions]        │   │
-│   │   VectorMeta:    vector_id → external_id + metadata │   │
+│   │   VectorData:    vector_id →                        │   │
+│   │                  external_id + metadata             │   │
+│   │                    + f32[dimensions]                │   │
```
agavra, line 89: if we do it this way I think we should just consider adding a `Vector` type to the metadata field types so you don't have to have just one vector, we treat it like any other field. this aligns with the read API that treats them identical and just inserts a `"vector"` field with a Vector type

vector/rfcs/0001-storage.md
```diff
@@ -130,6 +134,37 @@ To illustrate, here's how the query
...
+Trade-offs:
+
+- Requires a custom in-memory buffer (essentially a memtable)
+- Writes are not immediately visible during the buffer interval
+- For queryable buffers, a separate WAL and atomic flush protocol would be needed (out of scope)
+
+**Delete Operation:**
+
+Deleting a vector requires four atomic operations via `WriteBatch`: (1) add vector ID to the
+deleted bitmap (centroid_id = 0), (2) tombstone the vector data, (3) tombstone the vector metadata,
```
agavra: its no longer `centroid_id = 0` after this, right? we also don't have `vector data` anymore (just metadata)

vector/rfcs/0001-storage.md
```diff
@@ -75,8 +75,8 @@ index structures are stored as key-value pairs in the LSM tree.
 │   │   Centroids:     [chunk 0] [chunk 1] ... [chunk N]  │   │
 │   │                  (loaded into HNSW for navigation)  │   │
 │   │                                                     │   │
-│   │   Posting Lists: centroid_id → RoaringTreemap       │   │
-│   │                  (vector IDs per cluster)           │   │
+│   │   Posting Lists: centroid_id →                      │   │
+│   │                  [(vector_id, f32[dimensions])]     │   │
 │   │                                                     │   │
 │   │   Deleted:       centroid_id=0 → deleted vector IDs │   │
```
agavra: let's update this to `Deletions` instead of special centroid id

vector/rfcs/0001-storage.md
```diff
@@ -308,7 +343,36 @@ Unsupported changes (require creating a new collection):
 - Removing metadata fields or changing their types
 - Disabling indexing on a field (index entries would become stale)
 
-### `CentroidChunk` (`RecordType::CentroidChunk` = `0x02`)
+### `Deletions` (`RecordType::Deletions` = `0x02`)
```
agavra, line 354: i forgot, did we decide whether we were going to do something segmented here or just clean it up incrementally?

vector/rfcs/0001-storage.md
```diff
@@ -409,50 +459,42 @@ from candidate sets.
 ┌────────────────────────────────────────────────────────────────┐
 │                     PostingListValue                           │
 ├────────────────────────────────────────────────────────────────┤
-│  vector_ids:  RoaringTreemap                                   │
+│  postings: FixedElementArray<Posting>                          │
+│                                                                │
+│  Posting                                                       │
+│  ┌──────────────────────────────────────────────────────────┐  │
+│  │  type:        u8 (0x0 => Append, 0x1 => Delete)          │  │
```
agavra, line 474: curious why we're adding deletions directly to these lists - is it so we can merge them on compaction? we shouldn't need that if we have the deletion vector right?

vector/rfcs/0001-storage.md
rodesai: Yeah I was thinking it's better to optimize for appends since vector workloads seem to have a very high write volume. For queries I'm thinking these postings should be relatively short (around 50 vectors if we try to maintain ~10 vectors per posting and vectors are replicated to ~5 neighbors on average), so it shouldn't be too expensive to intersect these. The format is also simpler. We can revisit when we benchmark.

vector/rfcs/0001-storage.md
rodesai: yep, forgot to update this section

vector/rfcs/0001-storage.md
rodesai: The approach I had in mind was to always write these as merges, clean up entries during compaction of at least the last sorted run, and use a specialized filter that just removes them for the last sorted run. I'm writing another RFC to cover the details and list alternatives.

vector/rfcs/0001-storage.md
rodesai: > is it so we can merge them on compaction?

Yes. The deletion vector is for vectors that are actually deleted. The use case here is for LIRE which needs to be able to move vectors between centroids without actually deleting them. We can also use `Delete` entries to remove vectors as part of update/delete in the future, but the main reason at the moment is for LIRE.

rodesai: Also somewhat non-obvious point - LIRE could just try to write a new value for the posting list, but it would then need to read-modify-write the postings while maintaining isolation with ingest. My thinking was instead it could just read from a snapshot, and then apply `Append/Delete` operations for any vectors that need to move.

Issue #43, opened 2026-01-09
[queue] Initial design for Opendata-Queue


hachikuji:
## Summary

We’d like to add a **message queue abstraction** to OpenData. The key idea is to separate:

- A **message store + metadata log** (canonical source of truth)
- One or more **queues** that *tail the log* and maintain their own bookkeeping about consumption

The message store should not need to be aware of the downstream queues. Queues are readers of the log and derive their own state.

Treat the sketch below as a rough starting point. The ultimate design can be very different, but it's useful to get some initial thoughts on the board that we can start discussing.

---

## Core data structures

At the logical level, we have four families of keys.

### 1. Message store

Message bodies live once, keyed by ID:

```
/msg/{msg_id} → payload bytes (or pointer to blob)
```
The message store is queue-agnostic.

### 2. Metadata log

The log is a **metadata-only** stream that feeds any downstream queues:
```
/log/{seq} → { msg_id, headers?, created_at?, etc... }
```

### 3. Queue state: available

For each queue, track messages that are ready to be claimed:

```
/avail/{queue_id}/{msg_id} → { priority?, visible_at? }
```

Multiple queues may reference the same message ID.

### 4. Queue state: in-flight

Messages currently leased by workers:

```
/inflight/{queue_id}/{msg_id} → { lease_deadline, worker_id }
```

Queue operations (claim/ack/retry) manipulate `/avail` and `/inflight`, and only read `/msg/{msg_id}` after a claim succeeds.

## Key points

- Log entries are small and sequential.
- Storage is optimized to randomly access individual messages.
- Queues **tail the log** to discover new messages and add to availability tracking
- Messages are not read until they have been dequeued into /inflight.

## Open Questions

In principle, we could build the message store and the log by stitching together a standard key-value representation in SlateDb with our log implementation. If both of these data structures could reside within the same backing store, then we would have transactional writes. Should we allow data structures to occupy the same backing store? e.g. we could use a prefix to distinguish the two stores. 

Alternatively, we can build the key-value and log abstractions into a single data store using different record types. This fits closer to the data systems that we have today, but it probably results in some duplicate work since we still need a log abstraction.

agavra:
I wonder how the performance of heavy churn (moving messages from /avail -> /inflight) would work with Slate (or an LSM in general). There's probably some changes we'd need to make to support that pattern well (e.g. disable storing versions for MVCC in the memtable so we don't flush to L0 as often for those key types)

hachikuji:
[...]

PR #125, opened 2026-01-26
add queue RFC

cadonna:
## Summary

This RFC proposes a storage/data model for a queue system within the OpenData collection.

## Related Issues

Relates to https://github.com/opendata-oss/opendata/issues/43

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/queue/rfcs/0001-storage.md
+++ b/queue/rfcs/0001-storage.md
@@ -260,6 +260,10 @@ One strategy could be to provide a metadata log for each created queue.
 Another strategy could be to group messages according to the queue definitions.
 While this RFC enables multiple metadata logs it does not specify any distribution strategy 
 and leaves the specification for future RFCs.
+The generation of both, `instance ID` and `log ID`, is still to be determined since they depend on how 
+OpenData-Queue system instances will be deployed and on the distribution strategy of the metadata over the metadata logs.
+The important aspect is that the key allows multiple OpenData-queues on the same SlateDB instance and that
+an OpenData-queue allows to use multiple metadata logs.
 
 The value of the metadata record is an encoded array of message metadata:
 ```


queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+If multiple queue systems are maintained by the same slateDB instance, the key for the 
+metadata log needs to be suffixed by an identifier of the actual OpenData-Queue 
+instance, e.g., `/newsfeeds`.
+
+The record value contains the following metadata:
+- the message ID assigned to the payload by the payload store
+- the timestamp the message was created
+- the timestamp the message was received
+- the timestamp after which the message becomes visible the first time to consumers
+- the headers of the message (key-value pairs) that contain data for filtering and routing
```
hachikuji: This is interesting. I've been preferring minimal byte representations when possible for opendata. But since this is part of the metadata, it suggests that some enrichment of the data model is necessary. I guess an alternative might be to store an opaque byte array in the metadata. So a message in our queue might be defined like this:
```
struct Message {
  key: Bytes, // specifies queue
  value: Bytes, // message payload
  meta: Bytes, // routing/provenance, stored in metadata log
}
```
Then we wouldn't need to take a firm stance on headers. If we wanted to implement compatibility with some messaging spec, we could support headers within the metadata in a format that is compatible with the messaging spec.

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+
+An `Array<T>` is encoded as a `count (u16, little endian)` followed by `count` serialized elements of type T,
+encoded back-to-back with no additional padding. 
+The type is defined in [OpenData-Timeseries](timeseries/rfcs/0001-tsdb-storage.md).
+
+### Metadata log
+
+The metadata log stores the metadata of the published messages.
+The metadata log is an [OpenData-Log](log/rfcs/0001-storage.md).
+The key of the log is `opendata-queue/messages/metadata`.
```
hachikuji: Ah, interesting. All of the queue state is stored separately, but we just have a single log that they can follow. Alternatively, we could use the key to represent logically distinct streams of messages. That would allow for finer-grained following.

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+- the list of in-flight messages, i.e., the messages that are currently claimed by consumers 
+
+The encoding of the active messages value is as follows:
+```
+Queue State Value:
+    | available messages (Array<AvailableMessage>)
+    | in-flight messages (Array<InFlightMessage>)   
+```
+
+An `AvailableMessage` consists of:
```
hachikuji, line 359: Have you considering whether batching could be built into the design? I imagine many use cases would benefit from some flexibility defining their unit of work. Ideally we could feed more messages into batches so that we could keep the churn of the queue constant. A worker might expand the batch size dynamically so that it only had to hit the queue once every N seconds. That prevents the queue itself from becoming a bottleneck.

queue/rfcs/0001-storage.md
cadonna: How would you logically distinguish the streams of messages? 
Should the user decide by formulating an expression over the metadata?
Or should it be a partitioning scheme, like round-robin? But then how should a queue decide what metadata log to follow?
If the goal is to ensure scalability, replication of the metadata log could also be an option. Then the queues could be assigned round-robin to the replicas.

cadonna: My assumption was that the publishing of the messages is not per queue. A producer does not publish a message into a queue, but a queue decides what published messages it offers to its consumers.
That has the benefit that a single message can be consumed from multiple queues.
However, I am not opposed to bind the publishing to one or more queues.
If we want to bind the publishing to queues then we can of course have a metadata log per queue.

queue/rfcs/0001-storage.md
cadonna: Let me rephrase this to be sure I understand you:

The `key` in `Message` struct specifies the queue in which the message should be published. See my comment https://github.com/opendata-oss/opendata/pull/125/changes#r2732002747 regarding publishing messages.

The `value` is the payload that is stored in the payload store.

The `meta` basically replaces the headers in the current proposal. So the metadata log value would be encoded as follows:
```
Metadata Record Value:
    | message ID (u64)
    | created at (i64)
    | received at (i64)
    | visible at (i64)
    | retention duration in seconds (u32)
    | metadata type (u16)
    | metadata (Array<Byte>)
```   
where `metadata type` tells you how the bytes in `metadata` should be interpreted.

Is this how you had envisioned it?

queue/rfcs/0001-storage.md
cadonna: Do you see anything that would prevent batching in the proposed storage model?
We can add batches of available messages to the array, we can move batches of available messages to the in-flight messages as well as the other way around, and we can move batches of in-flight messages to the done and failed messages. So the storage model does not prevent batching. Or am I missing something?

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+need to be copied into each queue.
+Payloads are not read until they are claimed by consumers.
+
+### Payload store
+
+The payload store stores the payload of the messages that are published to the message 
+queue system. 
+The payload store assigns an ID to each published messages.
+The assigned ID is provided by a counter.
+For the counter, the same technique as for the sequence number in [OpenData-Log](log/rfcs/0001-storage.md) shall be used.
```
apurvam: Nit: it's not just the same technique. I believe that entire code base is reusable in `common`. If not, it should be made so.

queue/rfcs/0001-storage.md
hachikuji: Pretty much. But I'm not sure we necessarily need the type field. From the perspective of the queue data structure, it is just a chunk of bytes to pass through. From the perspective of a system implemented on top, there would be a schema, which may specify headers. I am imagining that protocols such as MQTT would be built as overlays on top on this data structure. The MQTT protocol including its header specification would be built as schemas on top of the raw byte fields.

cadonna: I see. However, without the readable metadata a queue could not use the metadata within its filter expression to add messages to itself.
I guess, we need to decide how a queue gets its messages. Does the message decide to which queue it belongs or does the queue specify which messages it contains?

queue/rfcs/0001-storage.md
hachikuji: I would say that all of these options are possible if the general primitive is write the message once, publish to one or more streams. Simple use cases might use a single key. All downstream users would see every message published. Other use cases might treat each key as a separate partition of data. A dumb example might be an email-like system where each key represents a separate user inbox. One message naturally may route to multiple inboxes. This might make sense when the logs for each partition become substantial in size, so filtering becomes unattractive. I guess my feeling is why not allow the most generality at the base layer?

queue/rfcs/0001-storage.md
hachikuji: Yeah, that might work. It looked to me like the tracking metadata was always O(num messages), but I might have missed something. Perhaps that is still fine as long as we can do batched updates. Seems like a useful discussion point for the RFC.

cadonna: Yeah, you are right. Tracking metadata is O(num messages). Let me revisit that.

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+## Goals
+- Specify a storage layout/data model for the queue
+
+## Non-Goals
+- Specification of a publish API for producers   
+- Specification of a consumption API for consumers
+- Specification of a admin API to specify queues
+- Specification of background processes that maintain the queues 
+
+## Design
```
agavra, line 49: I'm curious if this is inspired by any other queue system? I'm not very familiar with the space but it'd be good to have some references to what other systems do.

queue/rfcs/0001-storage.md
agavra: I'm guessing the reason for this is because you typically read messages that were produced together (i.e. FIFO order?). I'm wondering if there are any other groupings that make sense to prepend so that consumers aren't doing a bunch of random point lookups.

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+## Summary
+
+This RFC defines the storage format for OpenData-Queue. 
+The design separates message storage from queue state, where each queue maintains its 
+own append-only sequence of records tracking message consumption state.
+
+## Motivation
+
+OpenData-Queue is a message queue that allows producers to publish messages and 
+consumers to receive and process them.
```
agavra: I might be a little dense, but it'd be good to make it clear when you'd want `log` vs `queue` especially since we have both.

queue/rfcs/0001-storage.md
agavra: this sounds like we might want to support [Column Families](github.com/facebook/rocksdb/wiki/column-families) in SlateDB so that we can have multiple types of keys for different systems in the same slatedb instance for txn purposes.

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+Metadata Record Value:
+    | message ID (u64)
+    | created at (i64)
+    | received at (i64)
+    | visible at (i64)
+    | retention duration in seconds (u32)
+    | number of headers N (u8)
+    | key 0 (Utf8) | value 0 (Utf8)   
+    ...
+    | key N-1 (Utf8) | value N-1 (Utf8)
```
agavra: nit: I go confused for a second, we might want to call these `header_key` and `header_value` 😅

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+- the headers of the message (key-value pairs) that contain data for filtering and routing
+
+The record value is encoded as follows:
+```
+Metadata Record Value:
+    | message ID (u64)
+    | created at (i64)
+    | received at (i64)
+    | visible at (i64)
+    | retention duration in seconds (u32)
```
agavra: nit: why not use Slate's TTL feature to get cleanup for free? also we should consider exposing slatedb's create timestamp and expire_at timestamp so we can remove these from the record value itself

EDIT: I see there's the cleanup of the active/inflight lists as well. This feels pretty similar to handling deletions in `vector` cc @rodesai

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+
+The key of the active messages consists of:
+- the version of the key
+- the type of key, i.e., key of the active messages
+- The identifier of the queue
+
+The key is encoded as follows:
+```
+Active Message Key:
+    | version (u8) | type (u8) | queue ID (TerminatedBytes)
```
agavra: nit: we shouldn't need to use TermiantedBytes at the end of the key, we can just use bytes and read until the end of the key

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+The exact grammar and encoding of the filter expressions are still to be determined.
+
+After the maximum attempts the message is moved to the failed messages.  
+
+The retention duration for done messages is computed from the timestamp `done at` in the `DoneMessage` record value.
+If the current timestamp exceeds the `done at` timestamp plus the retention duration, the done message is 
+removed from the array of done messages in the `DoneMessages` value.
+A retention duration of `0` means infinite retention.
+
+### Record types overview
```
agavra: nit: helpful to move this up top so we can see all the record types before reviewing them all in detail

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+
+The value of the active messages consists of:
+- the list of available messages, i.e., the messages that can be claimed by consumers
+- the list of in-flight messages, i.e., the messages that are currently claimed by consumers 
+
+The encoding of the active messages value is as follows:
+```
+Queue State Value:
+    | available messages (Array<AvailableMessage>)
+    | in-flight messages (Array<InFlightMessage>)
```
agavra: maybe this is what @hachikuji is suggesting below with "Batching" but I'm a little nervous about having a single "god record" that contains the full state of the queue in one record and then relying on merges to update these.

Have we considered splitting it up into different keys? e.g. 
```
| version (u8) | type (u8) | queue ID (TerminatedBytes) | available/inflight (bool) | msg_id (u64)
```

OR a chunked approach where chunk corresponds to [start_ms, end_ms] ranges

```
| version (u8) | type (u8) | queue ID (TerminatedBytes) | chunk (u32, u32)
```

This is similar to `vector` and the way it stores centroids.

Either way it would be nice to write why we chose to do that and what order of magnitude messages we expect to be in each queue.

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+```
+InFlightMessage:
+    | message ID (u64) | priority (u8) | | claimed at (i64) | lease deadline (i64) | attempts (u16)   
+```
+
+There are multiple options to set the lease deadline. 
+Either the consumer that claims the message sets the deadline or the deadline is set 
+by the queue config.  
+
+Besides the available and in-flight messages, the queue store also maintains the list of done messages per queue.
```
agavra, line 408: nit: it would be nice to follow the pattern of other RFCs and add a sub-header for each record type

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+A message is done if a consumer claimed the message and acknowledged the successful processing. 
+The done message have a retention period after which the done messages are removed from the queue.
+The payload and the metadata of the messages are not removed from the system since other queues could still use them.
+The key for the done messages is encoded as follows:
+```
+Done Messages key:
+    | version (u8) | type (u8) | queue ID (TerminatedBytes)
```
agavra: same comment as above, why not include the message id in the key? that will avoid requiring us to continuously read/update the same key which is pretty bad for WA

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+```
+
+The config consists of:
+- the filters that specify which messages the queue contains.
+- the maximum attempts to process a message
+- the retention duration for done messages
+
+```
+Queue Config Value:
+    | filters (FilterExpression)
```
agavra: I'm guessing this should never change but the other two seem like they could change?

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+   3. the consumer acknowledges the successful processing of a message
+      1. the message is removed from the in-flight messages and added to the done messages
+   4. the consumer reports the unsuccessful processing of a message
+      1. if the message has still attempts, the message is removed from the in-flight messages and added to the available messages
+      2. if the message has no attempts left, the message is removed from the in-flight messages and added to the failed messages
+   
+
+## Alternatives
+
+### Use a record for each available and in-flight message
```
agavra, line 595: looks like you considered this alternative, what made you decide against it?

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,374 @@
...
+    | type of payload (u8) | payload (variable-length bytes)
+```
+The type of payload specifies the format of the payload.
+Initially, there will be two payload types:
+- `0x01` `Array<Byte>`  
+- `0x02` URI (encoding still to be determined)
+
+An `Array<T>` is encoded as a `count (u16, little endian)` followed by `count` serialized elements of type T,
+encoded back-to-back with no additional padding. 
+The type is defined in [OpenData-Timeseries](timeseries/rfcs/0001-tsdb-storage.md).
```
apurvam: nit: do you mean the count type? If not, which type is defined in the timeseries RFC?

apurvam (commented):
Thanks for this RFC. It's quite an effort and it's great to see how queue semantics could be implemented on top of an LSM tree. 

My big picture comment is this: the queue actually feels like the most complicated data structure we are trying to build, especially because each read could result in multiple writes. I think for this RFC, it may make sense to show how the underlying slateDB will be modified for some combination of read/write examples. 

This would help us get a sense of the perf. I'm not aware of queues-on-LSM trees being proven in the wild, so perf and costs is something we should try to derisk earlier than later. It would also be helpful to mention what the data structure designs for popular queues are.

queue/rfcs/0001-storage.md
cadonna: That is a good idea!
I added a paragraph about the distinction.

queue/rfcs/0001-storage.md
cadonna: I added a section with existing systems and protocols at the end of the design section.

queue/rfcs/0001-storage.md
cadonna: Actually, that is what I meant. I will re-formulate it.

cadonna: I did not consider messages that were produced together. I just wanted to avoid continuously (for each counter value) updating the counter.

queue/rfcs/0001-storage.md
cadonna: Type `Array<T>` is specified in timeseries. 
I added it to [the common types RFC](https://github.com/opendata-oss/opendata/blob/main/rfcs/0004-common-encodings.md).

cadonna: Anyways, I believe I made a mistake here. The payload for type `0x01` should simply be `Bytes`.

queue/rfcs/0001-storage.md
cadonna: Prefixes for sub-systems in OpenData might also do the trick. 
For example, all sub-systems might use the following prefix for all their records:
```
system (u8) | version (u8) | type (u8) | 
```
`system` could be:
- `0x00` log
- `0x01` timeseries
- `0x02` vector
- `0x03` queue
- ...

queue/rfcs/0001-storage.md
cadonna: My thoughts were the following:

With a record for each message, I have the following operations:

```
/avail/{queue ID}/{message ID}
/in-flight/{queue ID}/{message ID}
/done/{queue ID}/{message ID}
/failed/{queue ID}/{message ID}

- claim message: 
	- prefix scan /avail/{queue ID}
	- put /in-flight/{queue ID}/{message ID}
	- delete /available/{queue ID}/{message ID}
- acknowledge processing:
	- get /in-flight/{queue ID}/{message ID}
	- delete /in-flight/{queue ID}/{message ID}
	- put /done/{queue ID}/{message ID}
- failed processing:
	- get /in-flight/{queue ID}/{message ID}
	- delete /in-flight/{queue ID}/{message ID}
	- put /failed/{queue ID}/{message ID}
```

With a record for the state of the queue, I have the following operations:

```
/state/{queue ID}/
/done/{queue ID}/
/failed/{queue ID}/

- claim message: 
	- get /state/{queue ID}
	- put /state/{queue ID}
- acknowledge processing:
	- get /state/{queue ID}
	- put /state/{queue ID}
	- put /done/{queue ID} (<- can be a merge, since we only add messages)
- failed processing:
	- get /state/{queue ID}
	- put /state/{queue ID}
	- put /failed/{queue ID} (<- can maybe be a merge depending on how we clean-up/handle failed messages)
```

To me it seemed, that the second approach has slightly less write amplification and is more appropriate for batching. A consumer could claim a batch of messages and only need one get and one put. I did not count the lookups on the payload store, but they should be similar for both approaches.

queue/rfcs/0001-storage.md
cadonna: I tried to improve this part of the design. PTAL.

queue/rfcs/0001-storage.md
cadonna: I collected some information about existing protocols in the RFC. I still need to get my head around how to generalize.

queue/rfcs/0001-storage.md
cadonna: I adapted the metadata log format a bit. PTAL.

queue/rfcs/0001-storage.md
cadonna: You are right! Removed!

queue/rfcs/0001-storage.md
cadonna: After a discussion about composition that we had offline, I agree that column families would be the better option. The benefit of column families is the separation of the systems on memtable and SST level. That allows to use different data structures for each system and it also make the migrations of each system simpler.

queue/rfcs/0001-storage.md
cadonna: Done

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,623 @@
...
+```
+
+with `VarBytes` encoded as:
+
+```
+┌────────────┬───────────────────┐
+│   length   │       bytes       │
+│   (u32)    │ (variable-length) │
+└────────────┴───────────────────┘
+```
```
cadonna, line 300: This might be a candidate for the common encodings RFC.

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,630 @@
...
+The length of the payload can be computed from the length of the value.
+
+### Metadata log
+
+The metadata log stores the metadata of the published messages.
+The metadata log is an instance of [OpenData-Log](log/rfcs/0001-storage.md).
+
+#### Metadata record
+
+The key of the metadata record is `opendata-queue/{instance ID}/messages/metadata/{log ID}`.
```
apurvam, line 252: How are these IDs generated? Are they monotonically increasing? How are they assigned? Would we need them to be monotonically increasing to give us a natural sort order that makes lookups by (InstanceID, logID) pairs efficient?

queue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,630 @@
...
+The numbers of buckets start at 0 and increase by the capacity of the bucket.
+Let's assume the capacity is 10. 
+Thus, the first bucket has number 0, the second bucket has number 10, the third bucket number 20, etc. 
+
+The value of the active messages consists of:
+- the list of available messages, i.e., the messages that can be claimed by consumers
+- the list of in-flight messages, i.e., the messages that are currently claimed by consumers
+
+The encoding of the active messages value is as follows:
+```
```
apurvam, line 351: I know this was mentioned offline, but it really feels odd to have records like this in SlateDB. Is the idea to read-append-write these records with the same key to keep them up to date? Would it be better to consider a layout where the results are merged on read? I think a discussion of the tradeoffs of the two approaches would be beneficial.

apurvam: Ah. I see the discussion below in 'Alternatives considered'.

queue/rfcs/0001-storage.md
cadonna: I do not think they need to be necessarily monotonically increasing.
How they are generated is still to be determined. 

Regarding the instance ID, we have never discussed about having multiple OpenData sub-systems on the same SlateDB instance. That should be part of the system composition discussion. Since two OpenData-queues running on the same SlateDB instance are likely completely independent sub-systems, I assume the their instance IDs can also be completely independent, e.g., they do not need to be monotonic.

Regarding the log ID, its generation might depend on the strategy how to distribute message metadata across the metadata logs. For example, if the strategy is to have a metadata log per queue, then the log ID could be the queue ID. The distribution strategies are left for a future RFC. 

I added some details regarding the generation of the IDs.

cadonna:
> I think a lot of the questions can only be answered once we better define the use cases we are optimizing for. So let's cross that bridge when we come to it.

Yeah, I completely agree.

hachikuji:
@cadonna Thanks, LGTM! 

In regard to the notion of "buckets" to implement batching, I was wondering if we could make the behavior more dynamic. Rather than using a fixed size bucket, we could let the application request a specific batch size. The bucketing could be applied when moving a set of messages from "available" to "inflight" (which would require splitting these collections). Just food for thought to keep in your mind when we get to implementing this.

PR #131, opened 2026-01-30
Update log and timeseries to work in an observability pipeline

apurvam:
## Summary

Not for committing: this is a set of changes to get the OpenData-Log and OpenData-Timeseries working together in a metrics pipeline. These are combined into a flow that looks like this: 

```
Ingest Gateway → OpenData Log → Metrics Backend → OpenData Timeseries → prometheus-proxy
```

The actual pipeline is in another repo. This branch just contains the minimal changes needed to get the opendata services working in such a pipeline.

### DIFF
--- a/log/src/server/handlers.rs
+++ b/log/src/server/handlers.rs
@@ -4,7 +4,9 @@
 //! and ProtoJSON (`application/protobuf+json`) formats.
 
 use std::sync::Arc;
-use std::time::{Duration, Instant};
+use std::time::Duration;
+
+use tokio::time::Instant;
 
 use axum::body::Bytes;
 use axum::extract::{Query, State};
@@ -84,7 +86,7 @@ pub async fn handle_scan(
     let entries = scan_entries(&state, key.clone(), range.clone(), limit).await?;
 
     // If we have entries or follow is disabled, return immediately
-    if !entries.is_empty() || !params.follow {
+    if !entries.is_empty() || !params.follow.unwrap_or(false) {
         let values: Vec<Value> = entries
             .iter()
             .map(|e| Value {
@@ -347,6 +349,8 @@ mod tests {
             }
         }
 
+        impl StorageSnapshot for ConfigurableStorage {}
+
         #[async_trait]
         impl Storage for ConfigurableStorage {
             async fn apply(&self, _ops: Vec<RecordOp>) -> common::StorageResult<()> {
@@ -371,7 +375,7 @@ mod tests {
 
             async fn snapshot(&self) -> common::StorageResult<Arc<dyn StorageSnapshot>> {
                 self.check_failure()?;
-                Err(common::StorageError::Storage("not implemented".into()))
+                Ok(Arc::new(ConfigurableStorage::new()))
             }
 
             async fn flush(&self) -> common::StorageResult<()> {
--- a/log/src/server/request.rs
+++ b/log/src/server/request.rs
@@ -31,10 +31,7 @@ pub struct ScanParams {
     /// Maximum number of entries to return.
     pub limit: Option<usize>,
     /// If true, long-poll until data is available or timeout.
-    // `bool` doesn't implement `Deserialize` for missing fields like `Option<T>` does,
-    // so `#[serde(default)]` is needed to default to `false` when `?follow=` is omitted.
-    #[serde(default)]
-    pub follow: bool,
+    pub follow: Option<bool>,
     /// Timeout in milliseconds for long-polling (default: 30000).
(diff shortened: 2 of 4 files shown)


log/src/server/handlers.rs
```diff
@@ -79,7 +81,58 @@ pub async fn handle_scan(
...
+        return Ok(to_api_response(response, format));
+    }
+
+    // Long-poll: wait for new entries
+    let timeout = Duration::from_millis(params.timeout_ms.unwrap_or(30000));
+    let deadline = Instant::now() + timeout;
+    let poll_interval = Duration::from_millis(100);
+
+    loop {
+        tokio::time::sleep(poll_interval).await;
```
agavra: we should probably cap the poll interval to the deadline so we don't wait more than the timeout. (not a problem if timeout is >> poll_interval like it is here, but if we make it configurable I imagine it could get more important)

log/src/server/request.rs
```diff
@@ -30,6 +30,11 @@ pub struct ScanParams {
     pub end_seq: Option<u64>,
     /// Maximum number of entries to return.
     pub limit: Option<usize>,
+    /// If true, long-poll until data is available or timeout.
+    #[serde(default)]
```
agavra: I wonder why we need this annotation here but not for anything else 🤔

log/src/log.rs
```diff
@@ -210,6 +210,22 @@ impl LogDb {
...
 
+    /// Checks if the underlying storage is accessible.
+    ///
+    /// This performs a lightweight read operation to verify that the storage
+    /// backend is responding. Use this for health/readiness checks.
+    ///
+    /// # Returns
+    ///
+    /// Returns `Ok(())` if storage is accessible, or an error if the check fails.
+    pub async fn check_storage(&self) -> Result<()> {
```
agavra, line 221: we introduced a healthcheck `status()` method in slatedb we can use: https://github.com/slatedb/slatedb/pull/1234

log/Dockerfile
```diff
@@ -0,0 +1,59 @@
+# Use cargo-chef for dependency caching
```
agavra: let's put the dockerfile changes in a separate commit. I'd like to think about it for a second about where/how we want to do it but don't want to block this

timeseries/Dockerfile
```diff
@@ -0,0 +1,52 @@
+# Use cargo-chef for dependency caching
```
agavra: ditto

timeseries/src/promql/server.rs
```diff
@@ -292,3 +294,14 @@ async fn handle_label_values(
...
+
+/// Handle /-/healthy endpoint - returns 200 OK if service is running
+async fn handle_healthy() -> (StatusCode, &'static str) {
+    (StatusCode::OK, "OK")
+}
+
+/// Handle /-/ready endpoint - returns 200 OK if service is ready to serve requests
+async fn handle_ready(State(_state): State<AppState>) -> (StatusCode, &'static str) {
+    // Service is ready if it's running (TSDB is initialized in AppState)
+    (StatusCode::OK, "OK")
```
agavra, line 306: should we use the same mechanism as before? probably makes sense to expose some shared utility for storage in common.

agavra (approved):
LGTM, but I'd like to pull out the docker build pipelines to a different commit and probably separate it from the main build so it doesn't slow things down a bunch (and so that we have CI/CD etc...)

log/src/log.rs
apurvam: Good call — will file an issue to track integrating `Db::status()` via the Storage trait. For now the `SEQ_BLOCK_KEY` read works as a proxy, and we'll fix forward.

log/Dockerfile
apurvam: I've moved them.

log/src/server/handlers.rs
apurvam: The code has been updated to handle the case where timeout < poll interval. Tests are added too.

log/src/server/request.rs
apurvam: Good catch. It's because the `follow` field wasn't optional. I made it optional to match the rest.

log/src/log.rs
apurvam: FIled https://github.com/opendata-oss/opendata/issues/165

timeseries/src/promql/server.rs
apurvam: Yes, filed https://github.com/opendata-oss/opendata/issues/165 as a followup.

PR #132, opened 2026-01-30
add rfc for vector index rebalancing (lire)

rodesai:


vector/rfcs/0004-index-rebalancing.md
```diff
@@ -0,0 +1,865 @@
...
+.4% (4/1000) updates, and a merge for every .1% (but let’s set merge aside for
+now). Splits were configured to search a neighbourhood of 64 centroids for
+reassignment, resulting in an average of 5094 vectors (their implementation
+replicates vectors 8 ways) considered for reassignment, out of which 79 were
+actually reassigned (~1.5% reassign rate). If we project this to 10K
+updates/second, we’d need to be able to do 40 splits/second and 3160
+reassignments/second.
+
+In the pure insert case, we can expect a single split for every 10 inserts. This
+projects to 1000 splits/second and 79000 reassignments/second.
```
agavra: pure-insert case with 10k updates/second is probably only done in bootstrapping scenario, at which point we can probably delay LIRE to just rebuild the HNSW in full on some interval. I think it's OK if our benchmarks target the previous numbers for 3160 reassignments/s.

vector/rfcs/0004-index-rebalancing.md
```diff
@@ -0,0 +1,865 @@
...
+             │                        │                                                               │                                         
+             │                        │Split/Merge Tasks                                              │                                         
+             │                        │                                                               │                                         
+             │Split/Merge Commands    │                        ┌────────────────────────────┐         │read                                     
+             │                        │                        │                            │         │                                         
+             │                        │           ─────────────┤                            │         │                                         
+             │                        └──────────► Task Channel│                            │◄────────┘                                         
+             │                                    ─────────────┤      Imdex Rebalancer      │
```
agavra:
```suggestion
             │                                    ─────────────┤      Index Rebalancer      │                                                   
```

vector/rfcs/0004-index-rebalancing.md
```diff
@@ -0,0 +1,865 @@
...
+}
+```
+
+### Rebalance Task Model
+
+Rebalancing will be orchestrated by the Index Rebalancer by running the following
+tasks. Each task will iterate through it’s own state machine. The current state of
+each task is written to storage by the Write Command (see next section) that causes
+the transition.
+
```
agavra: does it simplify things if we choose not to persist ongoing LIRE commands? I feel like we should just abandon them on failure since they'll be re-triggered by noticing centroids that violate the min/max thresholds. We would need a way to make sure we atomically swap an old/new centroids after a split, which I feel like we can do in a single write batch.

vector/rfcs/0004-index-rebalancing.md
```diff
@@ -0,0 +1,868 @@
...
+   /// and adds them to V. The write for this state transitions SWEEP->REASSIGN
+   SWEEP,
+   /// Visit all vectors in C' and evaluate for reassignment using the centroid index. Reassign
+   /// all vectors requiring reassignment. The write for this state completes the split and
+   /// removes the task from storage.
+   REASSIGN
+}
+```
+
+#### Storage Format
```
agavra, line 358: as discussed offline, for v1 we should consider using a mechanism that executes a full LIRE within a single flush so we don't need to persist anything (failed LIRE rebalances would leave no intermediate state lingering)

agavra (approved):
Discussed a good chunk of this offline, but it LGTM! Very nicely done @rodesai and the explanations are clear and easy to follow :)

rodesai:
Closing in favor of #397

PR #134, opened 2026-01-31
Bump slatedb to 0.10.1

hachikuji:
Bumps slatedb


hachikuji:
I'm good with that. Updated.

Issue #127, opened 2026-01-28
[log] Add more granular metrics


apurvam:
## Summary

It would be good to add the following metrics to the log server. 

* records written
* bytes written
* records scanned
* bytes scanned

Note: records written is already an implemented COUNTER metric.

These metrics should exist as Gauges, which will be useful in measuring the throughput of the log. 

## Motivation

Such metrics are generally useful in measuring the performance and health of a logging service. 

## Proposed Solution

We already have a `log_append_records_total` metric that's a counter. We should use the same model to add the other gauge metrics outlined above.

PR454D:
@apurvam, by `bytes written` and `bytes scanned` metric, should this be only the value bytes ? or key bytes to be included as well ? Working on implementing this, just wanted to clarify this point..

I have created a draft PR for this 
#135

apurvam:
Both will do for now, as both can be considered part of the payload.

PR454D:
Is there any implementation regarding this issue open or can we close this?

apurvam:
No, we can fix this. Closing it out, thanks!

Fixed by #135

PR #135, opened 2026-02-02
Implement #127 Add more granular metrics

PR454D:
## Summary

What does this PR do?
Implements #127 . Have added a basic implementation of metrics. Would like to clear doubts in the issue and then will update accordingly.

## Related Issues

Fixes #
#127 

## Test Plan

How was this tested?
Unit test added

## Checklist

- [X] Tests added/updated - some tests to be added..
- [x] Documentation updated (if applicable) TBD..

### DIFF
--- a/log/src/server/metrics.rs
+++ b/log/src/server/metrics.rs
@@ -151,23 +151,6 @@ mod tests {
     }
 
     #[test]
-    fn should_increment_log_metrics() {
-        // given
-        let metrics = Metrics::new();
-
-        // when
-        metrics.log_append_bytes_total.inc_by(100);
-        metrics.log_records_scanned_total.inc_by(5);
-        metrics.log_bytes_scanned_total.inc_by(500);
-
-        // then
-        let encoded = metrics.encode();
-        assert!(encoded.contains("log_append_bytes_total_total 100"));
-        assert!(encoded.contains("log_records_scanned_total_total 5"));
-        assert!(encoded.contains("log_bytes_scanned_total_total 500"));
-    }
-
-    #[test]
     fn should_convert_http_method_to_label() {
         // given
         let method = Method::GET;


log/src/server/metrics.rs
```diff
@@ -72,6 +81,30 @@ impl Metrics {
             log_append_records_total.clone(),
         );
 
+        // Log append bytes counter
+        let log_append_bytes_written = Counter::default();
+        registry.register(
+            "log_append_bytes_written",
```
apurvam: we should add `_total` suffix to these metrics. It aligns with the prometheus naming scheme for counters. so I suggest `log_append_bytes_total` here.

log/src/server/handlers.rs
```diff
@@ -42,6 +42,12 @@ pub async fn handle_append(
     let request = AppendRequest::from_body(&headers, &body)?;
 
     let count = request.records.len();
+    let bytes_count: usize = request
+        .records
+        .iter()
+        .map(|rec| rec.key.len() + rec.value.len())
+        // should ^ this only consider value bytes ? or both ?
```
apurvam: Left a response on the issue, but we should count the key and and the value siszes in the bytes count. We can drop this comment and leave the code.

log/src/server/metrics.rs
```diff
@@ -72,6 +81,30 @@ impl Metrics {
...
+            "Total number of bytes written to the log",
+            log_append_bytes_written.clone(),
+        );
+
+        // Log scan records counter
+        let log_records_scanned = Counter::default();
+        registry.register(
+            "log_records_scanned",
```
apurvam: `log_records_scanned_total`

log/src/server/metrics.rs
```diff
@@ -72,6 +81,30 @@ impl Metrics {
...
+            "Total number of records scanned in the log",
+            log_records_scanned.clone(),
+        );
+
+        // Log scan bytes counter
+        let log_bytes_scanned = Counter::default();
+        registry.register(
+            "log_bytes_scanned",
```
apurvam: `log_bytes_scanned_total`

log/src/server/metrics.rs
PR454D: updated the metric names

log/src/server/metrics.rs
```diff
@@ -108,9 +144,29 @@ mod tests {
         // then
         let encoded = metrics.encode();
         assert!(encoded.contains("# HELP log_append_records_total"));
+        assert!(encoded.contains("# HELP log_append_bytes_total"));
+        assert!(encoded.contains("# HELP log_records_scanned_total"));
+        assert!(encoded.contains("# HELP log_bytes_scanned_total"));
         assert!(encoded.contains("# HELP http_requests_total"));
     }
 
+    #[test]
```
apurvam: I think we don't need this test, since it literally is reading and writing the same variable. I meant to create a log, append bytes, and then check the metric. We can add such a test later. For now, if you delete this test, I'll merge the PR.

apurvam (approved):
Thanks for the patch. Once you address the comment in this PR, I'll merge it.

log/src/server/metrics.rs
PR454D: Okay, removed the test..

PR454D:
Should I create a squash commit for this ?

apurvam:
> Should I create a squash commit for this ?

Thanks. I just did a squash and merge, so no need for a squash commit.

PR #136, opened 2026-02-02
move posting lists over to list of vectors

rodesai:
## Summary

- switch posting list from bitmap to list of vectors
- add the key for the deletions list since that's no longer modeled as a posting
- use a common type for index and deletions bitmaps

## Test Plan

- updated unit tests

## Checklist

- [x] Tests added/updated

### DIFF
--- a/vector/src/serde/posting_list.rs
+++ b/vector/src/serde/posting_list.rs
@@ -17,22 +17,22 @@
 //!
 //! ## Value Format
 //!
-//! The value is a FixedElementArray of PostingUpdate entries. Each entry contains:
+//! The value is a FixedElementArray of PostingUpdate entries, **sorted by id**.
+//! Each entry contains:
 //! - `posting_type`: 0x0 for Append, 0x1 for Delete
 //! - `id`: Internal vector ID (u64)
 //! - `vector`: The full vector data (dimensions * f32)
 //!
 //! ## Merge Operators
 //!
-//! Posting lists use SlateDB merge operators:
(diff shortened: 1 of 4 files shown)


vector/src/serde/posting_list.rs
```diff
@@ -177,10 +324,59 @@ impl Default for PostingListValue {
...
+pub(crate) fn merge_posting_list(existing: Bytes, new_value: Bytes, dimensions: usize) -> Bytes {
+    use std::collections::HashMap;
+
+    let vector_size = dimensions * 4;
+    let append_size = 1 + 8 + vector_size; // type + id + vector
+    let delete_size = 1 + 8; // type + id
+
+    // Map from id -> (buffer_index, offset, entry_size)
+    // buffer_index: 0 = existing, 1 = new_value
+    let mut locations: HashMap<u64, (usize, usize, usize)> = HashMap::new();
```
agavra: why not keep them sorted by id so when we merge it can be more efficient? we could also then choose better encoding algorithms at least for the ids (e.g. store just deltas as varints for ids)

given that pulling these into memory is probably the bottleneck for serving queries it's nice to get it down as small as possible and pack as many centroid posting into a single block as we can.

agavra (approved):
Skimmed, mostly just focused on the new posting list design.

vector/src/serde/posting_list.rs
rodesai: Was just trying to do the dumbest possible thing for now, but sort-merging makes sense. I'll update.

PR #137, opened 2026-02-02
add support for usearch as centroid graph

rodesai:
## Summary

- adds a trait for the centroid graph
- moves the current hnsw rust impl to an impl of the trait
- adds a usearch based implementation.

## Test Plan

unit tests added

## Checklist

- [x] Tests added/updated


vector/src/hnsw/hnsw.rs
```diff
@@ -1,41 +1,47 @@
...
 
-/// HNSW graph wrapper for centroid search.
-///
-/// The graph stores centroids and enables fast approximate nearest neighbor search
-/// to find relevant clusters during query execution.
-pub struct CentroidGraph {
+use super::CentroidGraph;
+
+/// HNSW graph implementation using the hnsw_rs library.
+pub struct HnswRsCentroidGraph {
```
agavra: is there any reason to use this? can we just remove this and the dependency? feels a bit early to start supporting multiple impls of the trait

agavra (approved):
LGTM but I think we should just axe the old depdnency for now

PR #138, opened 2026-02-02
unify vectordata and vectormeta

rodesai:
## Summary

Combines vectormeta and vectordata into a single forward index type now that postings store the actual vector data.

## Checklist

- [x] Tests added/updated


agavra (approved):
Skimmed, but the approach makes sense to me. Before long we probably want to turn VectorData into something we can deserialize partially if we know the schema (i.e. to apply post-filers / only deserialize what we need to return based on the query input). I'll create a ticket for it

PR #139, opened 2026-02-02
Point to slatedb 0.10.1 instead of main

hachikuji:
I pointed us to slatedb/main, but it's a little annoying to pull each time and it causes churn for Cargo.lock. I think it's better to stick to the latest version for now.


hachikuji:
Closing. I guess it's less of an issue with churn as I thought since the dependency only updates with an explicit update call.

PR #142, opened 2026-02-02
Add initial keyvalue RFC

hachikuji:
Initial storage draft for "keyvalue," which is a thin wrapper on top of SlateDb. API specs would follow in subsequent RFCs.

### DIFF
--- a/rfcs/0004-write-coordination.md
+++ b/rfcs/0004-write-coordination.md
@@ -1,0 +1,486 @@
+# RFC 0004: Write Coordination
+
+**Status**: Draft
+
+**Authors**:
+
+- [Almog Gavra](https://github.com/agavra)
+- [Jason Gustafson](https://github.com/hachikuji)
+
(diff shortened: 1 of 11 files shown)


keyvalue/rfcs/0001-storage.md
```diff
@@ -0,0 +1,90 @@
...
+SlateDB Key: | 0x01 | 0x10 | m | y | - | k | e | y |
+```
+
+### Values
+
+Values are passed through to SlateDB without modification. KeyValue does not impose any encoding or structure on values.
+
+## Alternatives
+
+### Direct SlateDB Usage
```
agavra, line 66: I think the biggest motivation IMO is that SlateDB doesn't have (and never will have) a server component to it. It's only embedded.

hachikuji: Good point. I'll add a note.

PR #144, opened 2026-02-03
write coordinator implementation of rfc-0004

agavra:
## Summary

Implements RFC-0004 for the write coordinator. I asked Claude to break up the commit into individually reviewable commits but then at the end to fix clippy I had to move `coordinator.rs` -> `mod.rs` which might break commenting on the most interesting of the commits... let me know if you have any problems reviewing it and I can restructure the PR again.

I recommend you primarily focus on the tests in `mod.rs` to see how the e2e flow would look like using the API.

## Related Issues

Relates to #95 and #82 

## Test Plan

How was this tested?

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/bencher/src/lib.rs
+++ b/bencher/src/lib.rs
@@ -102,7 +102,7 @@ impl Bencher {
                     storage: reporter_config.to_storage_config(),
                     ..Default::default()
                 };
-                let ts = timeseries::TimeSeries::open(ts_config).await?;
+                let ts = timeseries::TimeSeriesDb::open(ts_config).await?;
                 Some(Arc::new(TimeSeriesReporter::new(ts)))
             }
             None => None,
--- a/bencher/src/reporter.rs
+++ b/bencher/src/reporter.rs
@@ -1,6 +1,6 @@
 //! Reporter trait and implementations for benchmark metrics.
 
-use timeseries::{Series, TimeSeries};
+use timeseries::{Series, TimeSeriesDb};
 
 /// Reporter for benchmark metrics.
 ///
@@ -30,13 +30,13 @@ pub(crate) trait Reporter: Send + Sync {
     async fn flush(&self) -> anyhow::Result<()>;
 }
 
-/// Reporter that writes metrics to a TimeSeries database.
+/// Reporter that writes metrics to a TimeSeriesDb database.
 pub(crate) struct TimeSeriesReporter {
-    ts: TimeSeries,
+    ts: TimeSeriesDb,
 }
 
 impl TimeSeriesReporter {
-    pub(crate) fn new(ts: TimeSeries) -> Self {
+    pub(crate) fn new(ts: TimeSeriesDb) -> Self {
         Self { ts }
     }
 }
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -13,6 +13,7 @@ pub use sequence::{
 };
 pub use serde::seq_block::SeqBlock;
 pub use storage::config::StorageConfig;
+pub use storage::factory::{StorageRuntime, StorageSemantics, create_storage};
 pub use storage::loader::{LoadMetadata, LoadResult, LoadSpec, Loadable, Loader};
 pub use storage::{
     Record, Storage, StorageError, StorageIterator, StorageRead, StorageResult, WriteOptions,
--- a/common/src/storage/factory.rs
+++ b/common/src/storage/factory.rs
@@ -1,18 +1,110 @@
 //! Storage factory for creating storage instances from configuration.
 //!
-//! This module provides a factory function to create storage backends based on
-//! configuration, supporting both InMemory and SlateDB backends.
+//! This module provides factory functions for creating storage backends
+//! based on configuration, supporting both InMemory and SlateDB backends.
 
 use std::sync::Arc;
 
 use slatedb::DbBuilder;
 use slatedb::config::Settings;
 use slatedb::object_store::{self, ObjectStore};
+use tokio::runtime::Handle;
 
-use super::config::{ObjectStoreConfig, StorageConfig};
+use super::config::{ObjectStoreConfig, SlateDbStorageConfig, StorageConfig};
 use super::in_memory::InMemoryStorage;
 use super::slate::SlateDbStorage;
 use super::{MergeOperator, Storage, StorageError, StorageResult};
+
+/// Runtime options for storage that cannot be serialized.
+///
+/// This struct holds non-serializable runtime configuration like tokio
+/// runtime handles. Users can configure these options and pass them to
+/// system builders (e.g., `LogDbBuilder`, `TsdbBuilder`).
+///
+/// # Example
+///
+/// ```rust,ignore
+/// use common::StorageRuntime;
+///
+/// // Create a separate runtime for compaction
+/// let compaction_runtime = tokio::runtime::Builder::new_multi_thread()
+///     .worker_threads(2)
+///     .enable_all()
+///     .build()
+///     .unwrap();
+///
+/// let runtime = StorageRuntime::new()
+///     .with_compaction_runtime(compaction_runtime.handle().clone());
+///
+/// // Pass to a system builder
+/// let mut builder = LogDbBuilder::new(config);
+/// *builder.storage_mut() = runtime;
+/// let log = builder.build().await?;
+/// ```
+#[derive(Default)]
+pub struct StorageRuntime {
+    pub(crate) compaction_runtime: Option<Handle>,
+}
+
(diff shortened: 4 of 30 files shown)


common/src/coordinator/traits.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// The range of epochs contained in this flush (exclusive end).
+    /// Start is the first epoch in the flush, end is one past the last epoch.
+    pub epoch_range: Range<u64>,
+}
+
+/// A delta accumulates writes and can produce a snapshot image.
+pub trait Delta: Default + Clone + Send + Sync + 'static {
+    type Image: Send + Sync + 'static;
+    type Write: Send + 'static;
+
```
rodesai, line 50: wdyt about having a parameter for an `ImmutableDelta` that represents a frozen delta that is ready to be flushed? Then `fork_image` returns `(Self::ImmutableDelta, Self::Image)`

common/src/coordinator/traits.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// Initialize the delta from a snapshot image.
+    fn init(&mut self, image: &Self::Image);
```
rodesai, line 52: you can just make this an associated function (no self) parameter and have it return `Self` - e.g. `fn init(image: Self::Image) -> Self`. Then you don't need to rely on `Default` to construct the delta and `init` to initialize it.

rodesai: Also wdyt about giving the delta ownership of the image rather than passing it by reference so that the delta is free to mutate it. I think eventually we'll need to make `Image` internally mutable and pass it around with an `Arc` to allow readers to see it, but for now it seems simplest to just let the delta own it.

common/src/coordinator/traits.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// is complete, the [`Flusher::flush`] happens on a background thread.
+    fn fork_image(&self) -> Self::Image;
+}
+
+/// A flusher persists flush events to durable storage.
+#[async_trait]
+pub trait Flusher<D: Delta>: Default + Send + Sync + 'static {
+    /// Flush the given event to durable storage and returns a storage
+    /// snapshot for readers to use
+    async fn flush(&self, event: FlushEvent<D>) -> Result<Arc<D::Image>, String>;
```
rodesai: I think this needs to take a `Delta` and return a snapshot?

common/src/coordinator/handle.rs
```diff
@@ -0,0 +1,277 @@
...
+    pub async fn wait(&mut self, durability: Durability) -> WriteResult<()> {
+        let epoch = self.epoch().await?;
+
+        let recv = match durability {
+            Durability::Applied => &mut self.watchers.applied_rx,
+            Durability::Flushed => &mut self.watchers.flushed_rx,
+            Durability::Durable => &mut self.watchers.durable_rx,
+        };
+
+        while *recv.borrow() < epoch {
```
rodesai: nit: just use `wait_for`

common/src/coordinator/mod.rs
```diff
@@ -0,0 +1,1491 @@
...
+        // Reset the interval to start fresh from when run() is called
+        self.flush_interval.reset();
+
+        loop {
+            tokio::select! {
+                // this might starve flushes under heavy write load, we
+                // should consider alternative strategies if this becomes
+                // a problem (though heavy write loads should also trigger
+                // flushes based on size / backpressure)
+                biased;
```
rodesai: why use `biased` here?

common/src/coordinator/mod.rs
```diff
@@ -0,0 +1,1491 @@
...
+}
+
+impl<D: Delta, F: Flusher<D>> FlushTask<D, F> {
+    fn run(mut self) -> tokio::task::JoinHandle<WriteResult<()>> {
+        tokio::spawn(async move {
+            while let Some(event) = self.flush_rx.recv().await {
+                let flushed_epoch = event.epoch_range.end - 1;
+
+                self.flusher
+                    .flush(event)
```
rodesai: the output of this should be a slatedb snapshot right? what's the plan for routing that back to readers? I was thinking that should flow back to the write coordinator so it can manage the state of the current reader snapshot

common/src/coordinator/mod.rs
```diff
@@ -0,0 +1,1491 @@
...
+        // one flush is pending
+        let (flush_tx, flush_rx) = mpsc::channel(1);
+
+        let watcher = EpochWatcher {
+            applied_rx,
+            flushed_rx,
+            durable_rx,
+        };
+
+        let flusher = Arc::new(flusher);
```
rodesai: why stick this in an `Arc`?

common/src/coordinator/traits.rs
agavra: Makes sense to me, so `fork_image` would consume `self` and return the immutable version of it? That's basically how I use it in `handle_flush` anyway. Is the additional type just there to prevent us from accidentally using the mutable delta?

rodesai: > Is the additional type just there to prevent us from accidentally using the mutable delta?

Yeah. It's nice to be able to model a mutable delta differently from a frozen chunk of data that needs to be flushed.

common/src/coordinator/traits.rs
agavra: I had it as an associated function at first, now I don't remember why I changed it back to this. I think it was related to testing where there were times I didn't want to setup an `Image` but I don't think that ended up being helpful. We can make that change.

> wdyt about giving the delta ownership of the image

That makes sense to me as well.

common/src/coordinator/traits.rs
agavra: Yes, `FlushEvent` wraps the `Delta` and also includes an `epoch` (and an `image`, but with your other comment that will be part of the `Delta`)

The return I was debating whether to make it `Image` and then its up to the implementers to decide whether that's a `StorageRead` or something else that readers can use. Does that make sense or do you prefer to just require `StorageRead`? Seems nice to avoid some duplicate initialization work if we already have a materialized `Image`

common/src/coordinator/mod.rs
agavra: I imagined that we'd want the writes to take precedence over interval flushes but I guess there's not a totally good reason for that. I can remove it

common/src/coordinator/mod.rs
agavra: Yeah that was my plan as well but was leaving that for the next PR since it just added a bunch more coordinating code. The other alternative I was considering was passing the notification channel into the FlushTask so it can just directly send it out to watchers. One fewer channel to follow (reverse from FlushTask back to the coordinator)

common/src/coordinator/mod.rs
agavra: Ah originally `WriteCoordinator` held a reference to this but I see we don't need that anymore. Will fix.

common/src/coordinator/traits.rs
agavra: 👍 I'll name it `freeze` like in your implementation.

common/src/coordinator/traits.rs
agavra: Actually giving Delta the ownership of the Image makes things a little awkward. I imagine that an `Image` is immutable (a materialized snapshot) and then an `Image + Delta` make up the current state. I like the idea of keeping them separate.

common/src/coordinator/traits.rs
agavra: Actually this requires an extra clone of the delta or other rust awkwardness, I'm not sure the type safety is worth it since the frozen deltas don't live very long anyway (they just get sent straight to the flusher).

This is because this is my current flush:
```rs
let (new_delta, new_image) = self.delta.fork(&self.image);
let delta = std::mem::replace(&mut self.delta, new_delta);
let image = std::mem::replace(&mut self.image, new_image);
```

If I want to have `delta::fork` consume `self` then I need to either set self.delta to be `None` in the meantime (awkward) or clone it (not happy about that).

agavra: Ok wait let me double check that. I'm really struggling with rust today 😨

agavra: Yeah I think I'm right. In your code the clone happens implicitly when you create the `ImmutableDelta`

common/src/coordinator/traits.rs
agavra: OK after looking through this a bit I agree `StorageRead` makes more sense. Let's just do that and the readers can init it how they'd like.

PR #149, opened 2026-02-04
[log] Add logcat RFC

hachikuji:
It's useful to have simple cli tool to read from a log.

### DIFF
--- a/log/rfcs/0005-logcat-cli.md
+++ b/log/rfcs/0005-logcat-cli.md
@@ -1,0 +1,111 @@
+# RFC 0005: logcat CLI
+
+**Status**: Draft
+
+**Authors**:
+- Jason Gustafson <12502538+hachikuji@users.noreply.github.com>
+
+## Summary
+
+This RFC proposes `logcat`, a minimal command-line tool for reading entries from an OpenData-Log. Similar to `kafkacat` for Kafka, it provides a simple way to consume log entries from the command line for debugging, monitoring, and scripting.
+
+## Motivation
+
+OpenData-Log lacks a lightweight CLI tool for quick inspection and tailing of log data. Developers need to inspect entries during development, tail a key in real-time, and script log consumption in shell pipelines. A simple CLI fills this gap without requiring HTTP server setup.
+
+## Goals
+
+- Read entries from a log by key
+- Support tailing/following with configurable poll interval
+- Support resuming from a specific sequence number
+
+## Non-Goals
+
+- Writing/appending to the log
+- Complex filtering or transformations (pipe to `jq`, `grep`, etc.)
+- Multiple keys in a single invocation
+
+## Design
+
+### CLI Interface
+
+```
+logcat --key <KEY> [OPTIONS]
+
+Options:
+  -k, --key <KEY>           Key to read from (required)
+  -c, --config <PATH>       Path to config file (TOML)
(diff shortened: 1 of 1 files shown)


log/rfcs/0005-logcat-cli.md
```diff
@@ -0,0 +1,111 @@
...
+- Writing/appending to the log
+- Complex filtering or transformations (pipe to `jq`, `grep`, etc.)
+- Multiple keys in a single invocation
+
+## Design
+
+### CLI Interface
+
+```
+logcat --key <KEY> [OPTIONS]
```
agavra, line 33: honestly I never really liked Kafka's approach to this. I totally agree we need a utility but having a million different tools to interact with Kafka was always a pain.

I much prefer the traditional RDBMS approach where you have one REPL interface that you can use to SELECT/INSERT/DELETE etc... (and configurations etc...). It's a nicer UX.

hachikuji: I'm fine with that. I agree we don't want a jillion separate utilities.

apurvam: One tool would be better, and perhaps this tool can evolve into that. I can see the case for non-repl mode, if you want to pipe stuff into or out of the tool, etc. 

So perhaps we should just make it a more neutral name, something like `logcli` or something. Then it can add a repl mode and more functions.

hachikuji: I think something repl-like is good enough for what I need right now. The non-repl stuff is useful too, but could come later.

log/rfcs/0005-logcat-cli.md
```diff
@@ -0,0 +1,111 @@
...
+logcat -k events -s 100 -f -i 100
+```
+
+## Alternatives
+
+**HTTP API with curl**: Requires running the server and base64 encoding. `logcat` provides direct, simpler access.
+
+## Open Questions
+
+- Should the tool be named `logcat` or `logdbcat` to avoid confusion with Android's logcat?
```
samsond, line 105: Tiny naming suggestion: odlog might be worth considering here

log/rfcs/0005-logcat-cli.md
```diff
@@ -0,0 +1,111 @@
...
+This RFC proposes `logcat`, a minimal command-line tool for reading entries from an OpenData-Log. Similar to `kafkacat` for Kafka, it provides a simple way to consume log entries from the command line for debugging, monitoring, and scripting.
+
+## Motivation
+
+OpenData-Log lacks a lightweight CLI tool for quick inspection and tailing of log data. Developers need to inspect entries during development, tail a key in real-time, and script log consumption in shell pipelines. A simple CLI fills this gap without requiring HTTP server setup.
+
+## Goals
+
+- Read entries from a log by key
+- Support tailing/following with configurable poll interval
```
samsond, line 19: I noticed the roadmap mentions tail following “without polling,” while this RFC proposes polling with an interval. Are these intended to describe the same capability, with polling as an initial implementation, or is the roadmap pointing to a different future mechanism?

samsond (commented):
Looks good overall. I left two small review comments.

log/rfcs/0005-logcat-cli.md
PR454D: TUI would be a good use case for this cli, without args can act as a TUI, with specific --key arg, can show values to stdout.

PR #150, opened 2026-02-04
[log] Add segment discovery logic to reader

hachikuji:
The openmessaging benchmark I was trying to run didn't work because the reader cached segments at startup time and failed to discover segments created afterwards. In this patch, I've added a refresh loop for the segment cache which is configured by a `refresh_interval` config. This is also used to set the manifest poll interval in SlateDb.

### DIFF
--- a/log/tests/reader_writer.rs
+++ b/log/tests/reader_writer.rs
@@ -6,8 +6,8 @@
 use std::time::Duration;
 
 use bytes::Bytes;
-use common::storage::config::{LocalObjectStoreConfig, ObjectStoreConfig, SlateDbStorageConfig};
 use common::StorageConfig;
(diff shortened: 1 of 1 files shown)


log/Cargo.toml
```diff
@@ -38,3 +38,4 @@ prometheus-client = { version = "0.22", optional = true }
 base64.workspace = true
 proptest.workspace = true
 reqwest.workspace = true
+tempfile = "3"
```
apurvam, line 41: Is this meant to be cleaned up?

apurvam (approved):
LGTM, thanks!

log/Cargo.toml
hachikuji: I use it in the integration test. We need something other than the in-memory store to replicate the bug.

PR #151, opened 2026-02-04
replace "Image" with "Context" in write coordinator

agavra:


common/src/coordinator/traits.rs
```diff
@@ -44,12 +44,15 @@ pub struct FlushEvent<D: Delta> {
 
 /// A delta accumulates writes and can produce a snapshot image.
 pub trait Delta: Default + Sized + Send + Sync + 'static {
-    type Image: Send + Sync + 'static;
+    /// The Context is data owned only while the Delta is mutable. After
+    /// freezing the delta the context is returned to the write coordinator
+    type Context: Send + Sync + 'static;
```
rodesai, line 49: can you drop the `Sync` bound here?

agavra: If Context is not `Sync` then we can't have `Delta` both hold the `Context` and remain `Sync`, which we want so readers can eventually observe the in-progress `Delta`

PR #155, opened 2026-02-05
switch seq allocation to use deferred writes

rodesai:
## Summary

switches sequence allocation to a deferred writes model.
- SequenceAllocator now returns both a sequence number and optionally a record that needs to be written to storage to reserve the block if the seq num is in a new block
- Drops the SequenceBlockAllocator since the interaction with storage now happens outside

This allows sequence allocation to include sequence block puts in the deltas built up by write coordinator deltas

## Checklist

- [x] Tests added/updated

### DIFF
--- a/common/src/sequence.rs
+++ b/common/src/sequence.rs
@@ -1,18 +1,22 @@
 //! Sequence number allocation and persistence.
 //!
 //! This module provides components for allocating monotonically increasing
-//! sequence numbers with crash recovery. It consists of two layers:
+//! sequence numbers with crash recovery. It provides:
 //!
-//! 1. **[`SeqBlockStore`]**: Lower-level component for persisting sequence block
-//!    allocations to storage.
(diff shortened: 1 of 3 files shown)


common/src/sequence.rs
```diff
@@ -251,56 +168,89 @@ impl SequenceAllocator {
     /// # Errors
     ///
     /// Returns an error if block allocation fails (e.g., storage error).
-    pub async fn allocate(&self, count: u64) -> SequenceResult<u64> {
-        let mut block = self.block.lock().await;
-
-        let remaining = block.remaining();
+    pub fn allocate(&mut self, count: u64) -> (u64, Option<Record>) {
```
agavra, line 169: let's update the docs for this one. I'm imagining that these Records need to be tracked somewhere else and then applied to storage.

agavra (approved):
LGTM! Just a quick docs update

PR #157, opened 2026-02-05
fix write coordinator error handling + explicit lifecycle

rodesai:
## Summary

- propagate apply errors to write handle rather than failing coordinator
- make lifecycle management of the coordinator more explicit. Previously it would shutdown when there were no more write handles. After this change it's decoupled from the handles and stops on an explicit stop call.

## Test Plan

unit tests

## Checklist

- [x] Tests added/updated

### DIFF
--- a/common/src/coordinator/mod.rs
+++ b/common/src/coordinator/mod.rs
@@ -133,7 +133,7 @@ impl<D: Delta, F: Flusher<D>> WriteCoordinator<D, F> {
         let Some(write_task_jh) = self.write_task_jh.take() else {
             return Ok(());
         };
-        let _ = self.stop_tok.cancel();
+        self.stop_tok.cancel();
         write_task_jh.await.map_err(|e| e.to_string())?
     }
 }


common/src/coordinator/mod.rs
```diff
@@ -52,28 +52,18 @@ pub(crate) enum WriteCommand<D: Delta> {
...
-    flush_tx: mpsc::Sender<FlushEvent<D>>,
-    cmd_rx: mpsc::Receiver<WriteCommand<D>>,
-    applied_tx: watch::Sender<u64>,
-    #[allow(dead_code)]
-    durable_tx: watch::Sender<u64>,
-    epoch: u64,
-    last_flush_epoch: u64,
-    flush_interval: Interval,
+    handle: WriteCoordinatorHandle<D>,
+    stop_tx: oneshot::Sender<()>,
```
agavra: wondering why not use `CancellationToken`? we can reuse the same cancellation token across multiple tasks (including the flush task)

common/src/coordinator/mod.rs
```diff
@@ -96,49 +86,119 @@ impl<D: Delta, F: Flusher<D>> WriteCoordinator<D, F> {
...
+    }
+
+    pub fn start(&mut self) {
+        let Some((write_task, flush_task)) = self.tasks.take() else {
+            // already started
+            return;
+        };
+        let flush_task_jh = flush_task.run();
+        let write_task_jh = write_task.run(flush_task_jh);
+        self.write_task_jh = Some(write_task_jh);
```
agavra, line 129: did you mean to set `self.tasks` here? where does that get set?

common/src/coordinator/mod.rs
```diff
@@ -96,49 +86,119 @@ impl<D: Delta, F: Flusher<D>> WriteCoordinator<D, F> {
...
+    }
+
+    pub async fn stop(mut self) -> Result<(), String> {
+        let Some(write_task_jh) = self.write_task_jh.take() else {
+            return Err(String::from("coordinator already stopped"));
```
agavra: why not make stop idempotent?

common/src/coordinator/mod.rs
```diff
@@ -96,49 +86,119 @@ impl<D: Delta, F: Flusher<D>> WriteCoordinator<D, F> {
...
+        };
+        let _ = self.stop_tx.send(());
+        write_task_jh.await.map_err(|e| e.to_string())?
```
agavra, line 137: should we await the flush_task_jh as well?

common/src/coordinator/mod.rs
```diff
@@ -173,24 +238,26 @@ impl<D: Delta, F: Flusher<D>> WriteCoordinator<D, F> {
...
-        epoch_tx: oneshot::Sender<u64>,
+        epoch_tx: oneshot::Sender<Result<u64, String>>,
     ) -> Result<(), String> {
         let write_epoch = self.epoch;
         self.epoch += 1;
 
+        let result = self.delta.apply(write);
         // Ignore error if receiver was dropped (fire-and-forget write)
-        let _ = epoch_tx.send(write_epoch);
+        let _ = epoch_tx.send(result.map(|_| write_epoch));
```
agavra: I'm wondering if it makes sense to return both an epoch and the error somehow. that way some writers could choose to "ignore and continue"

common/src/coordinator/mod.rs
rodesai: ah that's what it's called 😄 . I was writing this on bart with no network access. Will do

common/src/coordinator/mod.rs
rodesai: good call. will do

common/src/coordinator/mod.rs
rodesai: The write task takes ownership of it, so there's no need. That way the write task can errror out early if the flush task fails

common/src/coordinator/mod.rs
rodesai: Not sure what you mean here. Why would you need the epoch if you wanted to ignore? There's nothing to wait on.

common/src/coordinator/mod.rs
agavra: > I was writing this on bart with no network access

🤠 woah, coding with no network access? I didn't realize that's still possible 🤣

common/src/coordinator/mod.rs
rodesai: Oh you mean for a partially applied write? I suppose we could, though I expect the usual pattern would be that once a write is enqueued its atomic.

common/src/coordinator/mod.rs
rodesai: It's set in `new` and consumed by `start` so the tasks can actually be spawned.

agavra: ah got it! i misread the `else` branch above

PR #158, opened 2026-02-05
vector write coordinator

rodesai:
## Summary

Moves vector write path over to the write coordinator. Writes are now simply enqueued on the write coordinator queue using the coordinator's handle. Writes are processed by `VectorDbDelta` and flushed by `VectorDbFlusher`.

## Test Plan

unit/integration tests

## Checklist

- [x] Tests added/updated

### DIFF
--- a/vector/src/delta.rs
+++ b/vector/src/delta.rs
@@ -105,6 +105,8 @@ impl Delta for VectorDbWriteDelta {
     type Context = VectorDbDeltaContext;
     type Write = Vec<VectorWrite>;
     type Frozen = VectorDbImmutableDelta;
+    type ApplyResult = ();
+    type Broadcast = ();
 
     fn init(context: VectorDbDeltaContext) -> Self {
         Self {
--- a/vector/src/flusher.rs
+++ b/vector/src/flusher.rs
@@ -3,14 +3,15 @@
 //! This module contains the `VectorDbFlusher` which applies accumulated
 //! RecordOps to storage atomically and updates the shared snapshot for queries.
 
+use std::ops::Range;
 use std::sync::{Arc, Mutex};
 
 use anyhow::Result;
 use async_trait::async_trait;
-use common::coordinator::{FlushEvent, Flusher};
-use common::storage::{Storage, StorageRead, StorageSnapshot};
+use common::coordinator::{BroadcastDelta, Flusher};
+use common::storage::{Storage, StorageSnapshot};
 
-use crate::delta::VectorDbWriteDelta;
+use crate::delta::{VectorDbImmutableDelta, VectorDbWriteDelta};
 
 /// Flusher implementation for VectorDb.
 ///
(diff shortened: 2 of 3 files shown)


vector/src/delta.rs
```diff
@@ -1,407 +1,604 @@
...
-//! timeseries delta, but adapted for vector data with external ID tracking.
+//! in memory before they are atomically flushed to storage.
 //!
 //! ## Key Design
 //!
-//! - Delta is keyed by **external_id** (String), not internal ID
-//! - ID dictionary lookups happen at flush time, not during write
-//! - No ID allocation during write() - deferred to flush time
-//! - Later writes to the same external_id override earlier ones
+//! - The `VectorDbImage` contains shared state: the in-memory ID dictionary
```
agavra: nit: I think we renamed this (might be worth using rustdoc links so it auto-updates on refactor)

vector/src/delta.rs
```diff
@@ -1,407 +1,604 @@
...
+/// This is passed to `Delta::init()` when creating a fresh delta. The image
+/// contains references to shared in-memory structures that persist across
+/// delta lifecycles.
+pub struct VectorDbDeltaContext {
+    /// Vector dimensions for encoding.
+    pub dimensions: usize,
+    /// In-memory ID dictionary mapping external_id -> internal_id.
+    /// Updated by the delta during apply().
+    pub dictionary: Arc<DashMap<String, u64>>,
+    /// In-memory centroid graph for assignment (immutable after initialization).
```
agavra, line 62: doesn't this need to change on LIRE tasks?

vector/src/delta.rs
```diff
@@ -1,407 +1,604 @@
...
-    config: &'a Config,
-    pending_vectors: HashMap<String, PendingVector>,
+/// Implements the `Delta` trait for use with WriteCoordinator.
+pub struct VectorDbWriteDelta {
+    /// Reference to the shared image.
+    ctx: VectorDbDeltaContext,
+    /// Accumulated RecordOps (ID dictionary, vector data).
+    ops: Vec<RecordOp>,
+    /// Posting list updates grouped by centroid (merged in freeze).
+    posting_updates: HashMap<u32, Vec<PostingUpdate>>,
```
agavra, line 87: wouldn't we also want to maintain attribute posting updates for the field inverted indexes? (same comment goes to actually building the delta)

vector/src/flusher.rs
```diff
@@ -0,0 +1,48 @@
...
+/// Flusher implementation for VectorDb.
+///
+/// Applies the accumulated RecordOps to storage atomically and updates
+/// the shared snapshot for queries.
+pub struct VectorDbFlusher {
+    /// Storage backend.
+    pub storage: Arc<dyn Storage>,
+    /// Shared snapshot for queries (updated after each flush).
+    /// Uses std::sync::Mutex - callers should lock briefly and clone.
+    pub snapshot: Arc<Mutex<Arc<dyn StorageSnapshot>>>,
```
agavra, line 25: I'm guessing this is going to be subsumed by #160?

vector/src/delta.rs
rodesai: yep, that'll change once we put the lire stuff in

vector/src/flusher.rs
rodesai: yep, once I integrate 160

vector/src/delta.rs
rodesai: yeah, that's just not implemented yet so it doesn't show up in this change. It'll come in a later pr

PR #159, opened 2026-02-06
[log] Use common write coordinator

hachikuji:
A couple changes in the coordinator. 
- Allow the delta to return an apply result. This is used in the log to obtain the sequence number corresponding to the write.
- Allow flushing without a delta. This is used in LogDb::flush
- Pass await_durable in flush so that we can indicate either Durability::Flushed or Durability::Durable
 
And in log:
- Update append semantics to use Durability::Applied by default. We lose "read your writes," but there's a path to get it back once we've taken over memtable accumulation I guess.
- Use common sequence allocator and get rid of log sequence allocator
- Factor out common LogDbRead which is shared between LogDb and LogDbReader. the only difference is how segment changes propagate to the cache. In LogDb, we subscribe to delta changes.

### DIFF
--- a/common/src/coordinator/traits.rs
+++ b/common/src/coordinator/traits.rs
@@ -40,8 +40,8 @@ pub enum Durability {
 ///   applied here until the delta is frozen.
 /// - **`Frozen`** — an immutable snapshot of the delta, sent to the
 ///   [`Flusher`] to be persisted to storage.
-/// - **`Broadcast`** — a lightweight payload extracted during the flush,
-///   sent to readers so they can update their read image.
+/// - **`Broadcast`** — a minimal representation of the flushed state
+///   that readers need to update their read image.
 pub trait Delta: Sized + Send + Sync + 'static {
     /// Mutable state owned by the delta while it accumulates writes.
     /// Returned to the write coordinator on [`freeze`](Delta::freeze) so the
@@ -52,8 +52,8 @@ pub trait Delta: Sized + Send + Sync + 'static {
     /// Immutable snapshot produced by [`freeze`](Delta::freeze), consumed by
     /// the [`Flusher`] to persist the batch to storage.
     type Frozen: Send + Sync + 'static;
-    /// Lightweight payload broadcast to subscribers after a flush, carrying
-    /// only the information readers need to update their read image.
+    /// Minimal representation of flushed state, broadcast to subscribers
+    /// so they can update their read image.
     type Broadcast: Clone + Send + Sync + 'static;
     /// Metadata returned from [`apply`](Delta::apply), delivered to the caller
     /// through [`WriteHandle::wait`](super::WriteHandle::wait).


common/src/coordinator/handle.rs
```diff
@@ -15,42 +15,64 @@ pub(crate) struct EpochWatcher {
...
+        Ok(self.recv().await?.epoch)
     }
 
     /// Wait for the write to reach the specified durability level.
-    pub async fn wait(&mut self, durability: Durability) -> WriteResult<()> {
-        let epoch = self.epoch().await?;
+    ///
+    /// Returns the apply result produced by [`Delta::apply`] once the
+    /// requested durability level has been reached.
+    pub async fn wait(&mut self, durability: Durability) -> WriteResult<M> {
```
agavra, line 74: this is a bit of an odd API to me. I feel like we should have three methods: `epoch()`, `result()` and `wait()` and then the first one that consumes the oneshot should cache the result so we can read it from another method (I think this may have been a bug in the original code where you couldn't both call epoch/wait).

common/src/coordinator/handle.rs
```diff
@@ -100,37 +122,42 @@ impl<D: Delta> WriteCoordinatorHandle<D> {
...
 
     /// Request a flush of the current delta.
     ///
     /// This will trigger a flush even if the flush threshold has not been reached.
+    /// When `await_durable` is true, the flush will also call `storage.flush()`
+    /// to guarantee durability, and the durable watermark will be advanced.
     /// Returns a handle that can be used to wait for the flush to complete.
-    pub async fn flush(&self) -> WriteResult<WriteHandle> {
-        let (epoch_tx, epoch_rx) = oneshot::channel();
+    pub async fn flush(&self, await_durable: bool) -> WriteResult<WriteHandle> {
```
agavra: nit: I feel like this should be `flush_durable` or something like that because this method will return as soon as the flush command is sent either way

common/src/coordinator/mod.rs
```diff
@@ -311,14 +332,23 @@ impl<D: Delta, F: Flusher<D>> FlushTask<D, F> {
                     .map_err(|e| WriteError::FlushError(e.to_string()))?;
```
agavra: don't we need to add an await durable API to `Flusher`? otherwise there's no guarantee that the data is actually durable

common/src/coordinator/mod.rs
```diff
@@ -250,44 +254,60 @@ impl<D: Delta> WriteCoordinatorTask<D> {
...
+        let has_writes = self.epoch != self.delta_start_epoch;
+        if !has_writes && !await_durable {
             return;
         }
 
         let epoch_range = self.delta_start_epoch..self.epoch;
-        self.delta_start_epoch = self.epoch;
+        if has_writes {
+            self.delta_start_epoch = self.epoch;
+        }
```
agavra, line 319: 😕 can't say I love all these conditionals and added Optionals piggy-backing on the flush path. I wonder if it's better to just add another top level `force_durable` API to the write coordinator that just coordinates directly with the `Flusher` to force a storage durable flush.

common/src/coordinator/handle.rs
rodesai: Starting to sound like a state machine 😄

common/src/coordinator/mod.rs
rodesai: I think it would be pretty clean to model it like this:
- Add a different message variant `FlushStorage` that the write coordinator can send to the Flusher to indicate that it should flush storage (e.g. call `Db#flush`).
- Change the flag in the `WriteCommand#Flush` to be called `flush_storage`
- If `flush_storage` is true, then the write coordinator sends this message after doing its normal flush handling.

I think this captures the intended flows pretty well. The periodic flush is just meant to clear data from the write coordinators buffers to keep memory usage in check and start the process of persisting the data. The client call to flush via the write handle may or may not want to request that data be forced to durable storage. In that case the coordinator can just send along a `FlushStorage` message after the normal flush handling. We probably also want to send a final `FlushStorage` message during clean shutdown of the coordinator.

hachikuji: Yeah, agree it is not ideal. Let me revisit and consider the suggestions.

common/src/coordinator/handle.rs
hachikuji: Trying to get what is odd about it. I don't feel too strongly about it, but it felt natural to have a single stop to wait for the durability I needed and get the result.

agavra: not sure why I felt weird about it last night, it makes more sense to me now.

common/src/coordinator/traits.rs
```diff
@@ -35,25 +35,15 @@ pub enum Durability {
...
-
 /// A delta accumulates writes and can produce a snapshot image.
 pub trait Delta: Sized + Send + Sync + 'static {
     /// The Context is data owned only while the Delta is mutable. After
     /// freezing the delta the context is returned to the write coordinator
     type Context: Send + Sync + 'static;
     type Write: Send + 'static;
     type Frozen: Clone + Send + Sync + 'static;
+    /// The payload broadcast to subscribers after a flush.
+    type Broadcast: Clone + Send + Sync + 'static;
```
hachikuji, line 57: Good discussion point here. The problem I noticed in the log implementation was that I had to clone the whole record set so that I could move it to storage _and_ send it out to subscribers. But the read path does not really need the individual records. It just needs to see the segments. So I introduced a separate type to represent the "broadcast delta". This allows systems to choose the portions of the delta to send to subscribers.

agavra: So we have:

1. `Delta`: the mutable, in progress delta
2. `Frozen`: what gets sent to the flusher to flush
3. `Broadcast`: what gets sent to readers to update their read image

I think that makes sense to me. Probably worth putting this in a comment at the top of `Delta` just so its obvious.

Though the abstraction makes sense to me, I'm wondering why we wouldn't want the reader to get the records themselves. isn't it likely that it's going to want to serve the most recently written records to subscribers that are at the tail of the log?

common/src/coordinator/traits.rs
```diff
@@ -70,10 +68,27 @@ pub trait Delta: Sized + Send + Sync + 'static {
...
+    /// Consumes the frozen delta by value and returns a snapshot for readers
+    /// along with a broadcast payload for subscribers.
+    async fn flush_delta(
+        &self,
+        frozen: D::Frozen,
+        epoch_range: &Range<u64>,
+    ) -> Result<BroadcastDelta<D>, String>;
+
+    /// Ensure storage durability (e.g. call storage.flush()).
+    async fn flush_storage(&self) -> Result<(), String>;
```
agavra, line 113: [not to address in this PR] just to make sure we're on the same page, eventually `flush_delta` and `flush_storage` will be merged if we no longer use a slate memtable but instead have our own WAL and direct L0 writes in a `flush`. does that make sense in your mental model? `flush_delta` would write to the WAL and make things durable, and you would use #160 to read from the current/frozen deltas.

log/src/delta.rs
```diff
@@ -0,0 +1,367 @@
...
+        Self {
+            context,
+            records: Vec::new(),
+            new_segments: Vec::new(),
+        }
+    }
+
+    /// Apply a write to the delta. Returns the [`AppendResult`] with the
+    /// base sequence number assigned to the batch.
+    fn apply(&mut self, write: Self::Write) -> Result<AppendResult, String> {
```
agavra, line 88: curious why you decided to do all this work in `apply` instead of doing some of it in `flush`? does it hurt to batch some writes together if we have a very quick flush interval?

log/src/log.rs
```diff
@@ -60,188 +51,50 @@ struct LogInner {
...
     storage: LogStorage,
     clock: Arc<dyn Clock>,
-    inner: RwLock<LogInner>,
+    read_inner: Arc<RwLock<crate::reader::LogReadInner>>,
+    flush_subscriber_task: JoinHandle<()>,
 }
 
 impl LogDb {
     /// Opens or creates a log with the given configuration.
-    ///
```
agavra, line 94: any reason why all the rustdoc went away? was this some weird Claude side-effect?

agavra (approved):
LGTM! Some comments and I think there's going to be a race with @rodesai's #160 but otherwise I think it's all good.

log/src/delta.rs
hachikuji: Some of the work could be deferred to flush I guess, but we probably need to do sequence allocation here since it's needed in the return type for append. The alternative is to force append to always await flush().

common/src/coordinator/traits.rs
hachikuji: Yeah, that probably makes sense once we have active subscription. I guess we might still want the ability to choose selectively though since a user might not be subscribed to all keys.

log/src/log.rs
hachikuji: Yeah, weird. Let me add it back.

common/src/coordinator/traits.rs
agavra: yeah I like the abstraction and it's like 0 overhead because you can just declare the broadcast type to be the same as the frozen type if you want. was just curious 😄

common/src/coordinator/traits.rs
hachikuji: Yeah, that makes sense to me.

common/src/coordinator/traits.rs
hachikuji: That's what I was thinking too. It is nice to have the option.

PR #160, opened 2026-02-06
include current+frozen deltas in read view

rodesai:
## Summary

Extends the reader view published by the write coordinator to include the current delta and frozen (unflushed) deltas:
- Changes the type of the broadcast event to View. View provides the current db snapshot + last flushed delta as before, but now also includes the frozen delta and a reader for the current delta.
- The frozen deltas are stored in FrozenDelta which includes both the Frozen associated type as well as the delta's epoch range
- The current delta is exposed to readers over a new associated type called Reader. Delta implementations provide a Reader type with the db-specific read apis

Also maintains a current view in the WriteCoordinator for readers that just want to observe the current state on each read.

For context, we'll need this for vector's LIRE tasks to be able to observe the current postings for centroids being split/merged

## Checklist

- [x] Tests added/updated

### DIFF
--- a/common/src/coordinator/mod.rs
+++ b/common/src/coordinator/mod.rs
@@ -837,7 +837,7 @@ mod tests {
             .collect();
         actual.sort_by(|a, b| a.0.cmp(&b.0));
         let mut expected: Vec<(&str, u64, u64)> = expected.to_vec();
-        expected.sort_by(|a, b| a.0.cmp(&b.0));
+        expected.sort_by(|a, b| a.0.cmp(b.0));
         for (actual, expected) in actual.iter().zip(expected.iter()) {
             assert_eq!(
                 actual.0, expected.0,
--- a/vector/src/db.rs
+++ b/vector/src/db.rs
@@ -34,8 +34,6 @@ use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Duration;
 
-type SnapshotLock = std::sync::Mutex<Arc<dyn StorageSnapshot>>;
-
 /// Vector database for storing and querying embedding vectors.
 ///
 /// `VectorDb` provides a high-level API for ingesting vectors with metadata.
@@ -45,10 +43,6 @@ pub struct VectorDb {
     config: Config,
     #[allow(dead_code)]
     storage: Arc<dyn Storage>,
-
-    /// Storage snapshot used for queries (updated by flusher).
-    /// Protected by std::sync::Mutex - callers should clone immediately and release.
-    snapshot: Arc<SnapshotLock>,
 
     /// The WriteCoordinator itself (stored to keep it alive).
     write_coordinator: WriteCoordinator<VectorDbWriteDelta, VectorDbFlusher>,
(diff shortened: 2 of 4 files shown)


common/src/coordinator/mod.rs
```diff
@@ -78,40 +86,42 @@ impl<D: Delta, F: Flusher<D>> WriteCoordinator<D, F> {
         // one flush is pending
         let (flush_tx, flush_rx) = mpsc::channel(1);
 
-        // Broadcast channel for flush results (buffer size 16 should be plenty)
-        let (flush_result_tx, _) = broadcast::channel(16);
+        // Channel for flush task to send completion results back to coordinator
+        let (flush_complete_tx, flush_complete_rx) = mpsc::unbounded_channel();
```
agavra: curious why not have the flush task handle this instead of sending it over two channels? in other words, I think it's OK for the flush task to maintain the `View` with a reference to the "current delta" Mutex. Makes the dataflow a little easier to follow IMO.

I _think_ this also might help simplify the:
```rs
while let Some(complete) = self.flush_complete_rx.recv()
```
loops we have in the main coordinator method

common/src/coordinator/mod.rs
```diff
@@ -78,40 +86,42 @@ impl<D: Delta, F: Flusher<D>> WriteCoordinator<D, F> {
...
 
-        let watcher = EpochWatcher {
-            applied_rx,
-            flushed_rx,
-            durable_rx,
-        };
+        // Broadcast channel for read state updates (buffer size 16 should be plenty)
+        let (view_tx, _) = broadcast::channel(16);
 
+        let flush_stop_tok = CancellationToken::new();
```
agavra, line 96: nit: I think the point of a cancellation token is that you can share it. we only need one `stop_tok` for all of the related tasks

common/src/coordinator/mod.rs
```diff
@@ -278,55 +327,105 @@ impl<D: Delta> WriteCoordinatorTask<D> {
...
+    fn handle_flush_complete(&mut self, complete: FlushComplete<D>) {
+        let flushed_epoch = complete.delta.epoch_range.end - 1;
+        let reader = self.delta.reader();
+        let view = {
+            let mut view = self.view.lock().expect("lock poisoned");
+            let mut new_frozen = view.frozen.clone();
+            let last = new_frozen
+                .pop()
+                .expect("frozen should not be empty when flush completes");
+            assert_eq!(last.epoch_range, complete.delta.epoch_range);
```
agavra: let's add a msg to this, but just making sure I got it this is preventing the case where `complete` comes out of order for whatever reason? I think this might have some problems with the way @hachikuji implemented #159 to send empty deltas and reserve epochs.

common/src/coordinator/mod.rs
```diff
@@ -1686,22 +1944,29 @@ mod tests {
...
-        // We can verify it exists and is usable (InMemoryStorage in tests)
         assert!(Arc::strong_count(&result.snapshot) >= 1);
 
         // cleanup
         coordinator.stop().await;
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-    async fn should_include_delta_in_flush_result() {
+    async fn should_include_delta_in_view_after_flush() {
```
agavra, line 2023: can we add a test to include reading from teh current, unflushed delta?

common/src/coordinator/traits.rs
```diff
@@ -68,6 +52,13 @@ pub trait Delta: Sized + Send + Sync + 'static {
     /// copy-on-write or reference counting) since it blocks writes. After this
     /// is complete, the [`Flusher::flush`] happens on a background thread.
     fn freeze(self) -> (Self::Frozen, Self::Context);
+
+    fn reader(&self) -> Self::Reader;
```
agavra: nit: rustdoc (ideally covering the sync conditions and when data is made available/reader is updated)

common/src/coordinator/traits.rs
```diff
@@ -68,6 +52,13 @@ pub trait Delta: Sized + Send + Sync + 'static {
...
+}
+
+pub struct FrozenDelta<D: Delta> {
```
agavra: _super_ nit: wdyt about `StampedDelta` or `DeltaAndEpoch` so we don't have two names for `Frozen` and `FrozenDelta` which is a little confusing.

If we just create a `Stamped` it could be:
```rs
pub struct Stamped<T> {
  pub val: T,
  pub epoch_range: Range<u64>
}
```

Making it generic to anything that has an epoch range associated with it. Not sure if this is helpful or not.

rodesai:
> LGTM, I do feel like when we're done with all these coordinator changes we should take a step back and see if we can simplify the loop. All these channels floating around make me a little uncomfortable since it's so hard to debug message-passing.

Totally agree. Once this is merged I have a PR that adds another channel for internal writes 😅

common/src/coordinator/mod.rs
rodesai: ah yeah we could do that. I asked claude to implement it so that the coordinator task owns the current view but I had to add a mutex anyway to make it observable by readers.

common/src/coordinator/mod.rs
rodesai: This is intentional. The shutdown needs to propagate gracefully down the pipeline. So first the write task stops, then it sends its final flush, and then once that flush is done, it stops the flush task.

common/src/coordinator/mod.rs
rodesai: yeah it's just a sanity check to make sure we see the flush result that we expect.

common/src/coordinator/mod.rs
rodesai: Will do. I was thinking there isn't much to test there since the interesting part is the actual reader. But its worth testing that the reader is set correctly to point to the new delta after the current one is frozen.

PR #161, opened 2026-02-07
Implement offset and @ modifiers for vector/matrix selectors

### LINKED ISSUE #91: [tsdb] evaluate vector selectors with respect to the specified offset and at modifiers
## Summary

PromQL allows specifying `offset` and `@` modifiers when selecting vectors. The evaluator currently just ignores these. We need to extend vector selection to adjust the start/end times of the vector lookups based on these modifiers.

## Motivation

Full PromQL support

## Proposed Solution

See above

## Additional Context

https://prometheus.io/docs/prometheus/latest/querying/basics/#offset-modifier

samsond:
## Summary
Adds support for PromQL time modifiers to achieve Prometheus compatibility:
- `offset <duration>`: shifts evaluation time backward (positive) or forward (negative)
- `@ <timestamp>`: sets absolute evaluation time
- `@ start()` / `@ end()`: uses query envelope boundaries

## Implementation details:
- Added `apply_time_modifiers()` helper that applies `@` first, then `offset`
- Updated `evaluate_vector_selector()` and `evaluate_matrix_selector()` to accept three time parameters: `evaluation_ts` (per-step time), `query_start`, and `query_end` (query envelope for `@ start()`/`@ end()`)
- Clarified time parameter roles by introducing `evaluation_ts` for per-step evaluation time (avoiding overloading `end`)
- For instant queries, all three time parameters equal the query time

## Related Issues

Fixes #91 

## Test Plan

Added 5 new tests covering:

* offset
* @ <timestamp> 
* @ start() / @ end()
* combined @ + offset semantics

All existing and new tests pass (401 total)

## Checklist

- [X] Tests added/updated

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -672,8 +672,11 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
     /// - `@ start()`: Uses query start time
     /// - `@ end()`: Uses query end time
     ///
-    /// When both modifiers are present, `@` is applied first, then `offset` is applied
-    /// relative to the `@` time (order-independent).
+    /// When both modifiers are present, `offset` is applied relative to the `@`
+    /// modifier time. Although PromQL defines the result as order-independent
+    /// (e.g. `@ t offset d` == `offset d @ t`), we normalize the implementation
+    /// by applying `@` first and then applying `offset`. This keeps the logic
+    /// simple and matches Prometheus semantics.
     ///
     /// See: <https://prometheus.io/docs/prometheus/latest/querying/basics/#offset-modifier>
     fn apply_time_modifiers(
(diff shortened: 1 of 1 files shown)


timeseries/src/promql/evaluator.rs
```diff
@@ -2294,4 +2440,413 @@ mod tests {
             );
         }
     }
+
+    #[tokio::test]
+    async fn should_evaluate_vector_selector_with_offset_modifier() {
```
agavra: can we add tests for evaluating with positive and negative offsets? and it would be good to have tests for timestamps that don't align exactly with the evaluation timestamp.

timeseries/src/promql/evaluator.rs
```diff
@@ -611,6 +664,59 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
+                AtModifier::Start => query_start,
+                AtModifier::End => query_end,
+            };
+        }
+
+        // Apply offset modifier (relative adjustment)
+        if let Some(offset) = &vector_selector.offset {
+            adjusted_time = match offset {
+                Offset::Pos(duration) => {
+                    // Positive offset: look back in time
```
agavra: nit: since this seems counterintuitive let's just make it extra clear that this is intentional by giving an example and noting this is how prometheus works (I double checked)

timeseries/src/promql/evaluator.rs
samsond: Agreed, I will add coverage.

timeseries/src/promql/evaluator.rs
samsond: Good point. I will update.

agavra:
Merging. Thanks again @samsond for the contribution 😄

PR #164, opened 2026-02-09
migrate timeseries to use write coordinator

### LINKED ISSUE #95: [tsdb] serialize all ingestion to a given time bucket

agavra:
## Summary

This change is 100% Claude's attempt to oneshot a migration to the write coordinator. I reviewed it to make sure its sane and it looks right to me, and I've also split it into commits that are easily reviewable.

I think the caching mechanisms might be all sorts of messed up, but I'm going to fix forward for that so don't focus too much on that.

## Related Issues

Fixes #95
Relates to #82 

## Test Plan

Unit tests

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/timeseries/src/minitsdb.rs
+++ b/timeseries/src/minitsdb.rs
@@ -182,9 +182,9 @@ impl MiniTsdb {
 
         let handle = self.write_coordinator.handle(WRITE_CHANNEL);
         let mut write_handle = handle
-            .write(series_list.to_vec())
+            .try_write(series_list.to_vec())
             .await
-            .map_err(map_write_error)?;
+            .map_err(|e| map_write_error(e.discard_inner()))?;
 
         write_handle
             .wait(Durability::Applied)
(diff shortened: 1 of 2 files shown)


hachikuji:
I was considering the complexity of the coordinator-per-bucket model. It's nice that it simplifies the delta logic. The downside though is that it makes it hard to reason about the state across all buckets. The epochs from each coordinator would not have any relation to each other. I don't know if there is an immediate downside. Perhaps it causes strange write anomalies where we flush a newer bucket before an older one for example. Might that lead to temporary gaps in the data? My gut feeling is that a single coordinator would make things easier for us in the long term.

agavra:
That makes sense to me, discussed offline. Primary motivation of having it this way was that it was faster/easier to implement (changes the least structure). I agree that long term its probably better to centralize the write coordinator.

PR #172, opened 2026-02-10
Enable write coordinator blocking and retry

hachikuji:
There are backpressure scenarios where both blocking and non-blocking semantics are useful. In the embedded Log API, I think we want want to have both options available. We could have a blocking `append` and a non-blocking `try_append` for example. To allow both options, I've added a `write_timeout` method to `WriteCoordinatorHandle`. I've also modified `WriteError` so that it returns the `Write`. This enables retry loops without cloning.

### DIFF
--- a/log/src/error.rs
+++ b/log/src/error.rs
@@ -83,7 +83,8 @@ impl From<SequenceError> for Error {
 impl From<WriteError> for Error {
     fn from(err: WriteError) -> Self {
         match err {
-            WriteError::Backpressure => Error::Internal("write queue full".into()),
+            WriteError::Backpressure(()) => Error::Internal("write queue full".into()),
+            WriteError::TimeoutError(()) => Error::Internal("write queue timeout".into()),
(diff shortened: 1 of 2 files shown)


common/src/coordinator/handle.rs
```diff
@@ -128,20 +129,69 @@ impl<D: Delta> WriteCoordinatorHandle<D> {
...
+
     /// Submit a write to the coordinator.
     ///
     /// Returns a handle that can be used to retrieve the apply result
-    /// and wait for the write to reach a desired durability level.
-    pub async fn write(&self, write: D::Write) -> WriteResult<WriteHandle<D::ApplyResult>> {
+    /// and wait for the write to reach a desired durability level. On
+    /// failure the original write is returned inside the error so it
+    /// can be retried without cloning.
+    pub async fn write(
```
agavra: let's also rename this `try_write` to be in line with `try_send` if we're going to mirror `try_timeout`

PR #173, opened 2026-02-10
add docker build, request stat metrics, shutdown hooks

apurvam:
## Summary

3 separate changes to get closer to a production quality system: 

1. docker builds
2. request stat histograms (requests pending, etc)
3. shutdown hooks.

I can split this into 3 PRs. The main discussion is when and how to trigger the docker builds. Once we have that figured out I can implement the changes.


.github/workflows/docker.yml
```diff
@@ -0,0 +1,126 @@
+name: Build and Push Docker Images
+
+on:
+  push:
+    branches: [main, opendata-heracles, opendata-heracles-2]
```
apurvam: I only added opendata-heracle and opendata-heracles-2 because I needed the images from these branches. Once the current CI run finishes, I'll delete these before merging.

agavra: after this CI merges let's just delete this full file. I'll merge it with https://github.com/opendata-oss/opendata/blob/main/.github/workflows/build-binaries.yml manually.

We'll probably want to have some kind of "nightly/dev" build that we can use for testing, but we can probably set that up later and just run it on a release for now.

log/src/server/metrics.rs
```diff
@@ -113,13 +128,34 @@ impl Metrics {
             http_requests_total.clone(),
         );
 
+        // HTTP request duration histogram (buckets from 1ms to ~8s)
```
agavra, line 131: is 8s enough when `follow` that can take 30s?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,126 @@
...
+    tags: ["v*"]
+  pull_request:
+    branches: [main]
```
agavra: also why are we running the docker build on each pull request?

apurvam: Because I was iterating on a branch, and wanted outputs from each push to my PR to test with.

.github/workflows/docker.yml
apurvam: sounds good. The whole point of this file was to get CI images to test with. When the current build succeeds, we would have validated each dockerfile. So I'll remove this.

log/src/server/metrics.rs
apurvam: The `follow` _timeout_ is 30 seconds. Currently, it's really naive. If there is no data when the scan comes it, it will block for 30 seconds, regardless of whether new data arrives in the meantime. The only purpose the current fetch api serves is to avoid the client from spinning.

apurvam:
> LGTM. we might want to quickly discuss what we want to do with the release images

WDYM? Like when should we generate images? or something else?

agavra:
> WDYM? Like when should we generate images? or something else?

Ah nevermind that, I think I misunderstood the intention of the current docker build workflow in this PR.

Issue #130, opened 2026-01-29
Enable OpenMessaging Benchmark for Log


hachikuji:
# Enable OpenMessaging Benchmark for Log

## Summary

Enable the Log to run the [OpenMessaging Benchmark (OMB)](https://openmessaging.cloud/docs/benchmarks/) framework to provide standardized, comparable performance metrics. [WarpStream](https://www.warpstream.com/blog/warpstream-benchmarks-and-tco) is the closest architectural comparison (object-storage backend, stateless agents) and uses OMB with published results, making it an ideal reference point.

## Motivation

- **Standardized benchmarking**: OMB is the industry standard used by Kafka, Pulsar, Redpanda, WarpStream, and others
- **Comparable results**: Enables direct comparison with WarpStream's published benchmarks. WarpStream is architecturally similar to Log—both use object storage as the primary data plane (rather than local disks), resulting in stateless compute nodes and a similar latency/cost trade-off profile
- **Production-realistic workloads**: OMB supports configurable fan-out, batch sizes, and client counts
- **KPI coverage**: Captures throughput (msg/s, MB/s), latency percentiles (P50-P99.99), and backlog metrics

### 1. Agree on OpenMessaging as the benchmark

- [ ] Confirm OMB as the target benchmark framework
- [ ] Review OMB driver API requirements:
  - `BenchmarkDriver`: initialize, createTopic, createProducer, createConsumer
  - `BenchmarkProducer`: `sendAsync(key, payload)` → `CompletableFuture<Void>`
  - `BenchmarkConsumer`: push-based via `ConsumerCallback.messageReceived(payload, publishTimestamp)`

### 2. Design fair comparison with WarpStream

WarpStream's benchmark configuration:
- 1 million messages/second, 1 KiB messages
- 288 partitions (single topic), 64 producer/consumer clients
- LZ4 compression, `randomBytesRatio: 0.5` (realistic compression)
- Fan-out scenarios: 1x → 3x consumer fan-out

**Questions to resolve:**
- [ ] How do Log keys map to OMB topics/partitions? (288 keys? single key with implicit partitioning?)
- [ ] What client configuration achieves comparable batching behavior?
- [ ] What are acceptable latency bounds? (WarpStream: ~600ms P99 E2E with object storage)
- [ ] How do we handle fan-out scenarios without consumer groups?

**Reference metrics from WarpStream (6x m6in.4xl nodes):**
| Metric | Value |
|--------|-------|
[...]

PR #176, opened 2026-02-11
fix: use BTreeMap for Params for consistent benchmark output

ryandielhenn:
## Summary

What does this PR do?

Use BTreeMap instead of Hashmap for Params to ensure benchmark output parameters are displayed in consistent order across all parameter sets for readability

## Test Plan

How was this tested?

Ran cargo run --bin log-bench and verified params appear in consistent order across all benchmark runs.

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/bencher/src/params.rs
+++ b/bencher/src/params.rs
@@ -1,6 +1,6 @@
 //! Benchmark parameter collection.
 
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 
 use timeseries::Label;
 
@@ -11,14 +11,14 @@ use timeseries::Label;
 /// to/from `Vec<Label>` for use with the metrics system.
 #[derive(Debug, Clone, Default)]
 pub struct Params {
-    inner: HashMap<String, String>,
+    inner: BTreeMap<String, String>,
 }
 
 impl Params {
     /// Create an empty Params collection.
     pub fn new() -> Self {
         Self {
-            inner: HashMap::new(),
+            inner: BTreeMap::new(),
         }
     }
 


agavra (approved):
LGTM, thanks @ryandielhenn - I'm curious what you were working on to have noticed this.

ryandielhenn:
@agavra Just exploring the codebase, noticed the output was hard to scan when comparing parameter sets across runs. Planning to work on TSDB and vector benchmarks.

agavra:
Makes sense, probably best to start with `tsdb`. Vector is still very early, we're still implementing the index maintenance mechanism (cc @rodesai)

ryandielhenn:
I also read both [RFC 0003](https://github.com/opendata-oss/opendata/blob/main/rfcs/0003-benchmark-framework.md) and [issue #130](https://github.com/opendata-oss/opendata/issues/130) and my understanding is OpenMessaging is used for competitive benchmarking against Warpstream, Kafka, etc, and the framework mentioned in the RFC and this PR is for internal regression testing is that correct?

agavra:
Yup, that's correct (for `log`). You can see the implementation of that here: https://github.com/opendata-oss/openmessaging-benchmark

For external testing of `timeseries` we use a fork of https://github.com/VictoriaMetrics/prometheus-benchmark -- you can see https://github.com/responsivedev/p8s-bench for a terraform'd deployment of that.

hachikuji:
My hope was to eventually have minimalistic workloads for nightly regression testing on all systems. Sometimes external benchmarks are a pain to hook in. Openmessaging, for example, brings in the whole jvm. For log/queue/keyvalue, it should be easy to define realistic, simple workloads.

ryandielhenn:
@agavra I started working on the `tsdb` benchmark and noticed that `TimeSeriesDb::open` is not yet implemented, so I'm putting it down but happy to finish once that lands.

agavra:
😓 @ryandielhenn you caught us in the middle of a rename. right now the main method is `Tsdb::new`, we'll eventually migrate it to `TimeseriesDb` but you can use the former for now.

PR #177, opened 2026-02-11
feat(promql): introduce promqltest DSL framework

### LINKED ISSUE #93: Add a specification-by-example-driven functional test suite covering PromQL queries
## Summary

We should add an easy-to-update suite of functional tests that exercise the full PromQL query language. The idea would be to have each test case driven by a specification. I'm imagining this could look something like this. Lets say the specification is in json, then it could have the form:
```json
{
    "inputs": [
        {
            "labels": {...label kv pairs...},
            "points": [{"timestamp": <timestamp>, "value": <value>}]
        },
    ],
    "query": "<promql query string>"
    "result": "<expected result, including data type e.g. scalar/instant-vector/range-vector and data points>"
}
```

The test would then first start a timeseries server, then iterate over these and for each, seed the test with the specified data, execute the query, and validate the results.

It should also be easy to have an LLM generate these test cases given the PromQL specification documentation.

## Motivation

Makes sure we exercise the full PromQL query language.

## Additional Context

This idea is inspired by the KSQL query validation tests: https://github.com/confluentinc/ksql/tree/master/ksqldb-functional-tests#ksql-query-validation-tests

samsond:
## Summary
Implement a Prometheus-inspired promqltest DSL for functional testing of PromQL queries via .test files.

What does this PR do?

Core components:
- dsl.rs: parses load, eval instant, and clear commands
- runner.rs: auto-discovers testdata/*.test and executes commands
- testdata/: directory-based test organization

Key design decisions:
- Sequential block semantics for multiple value expressions to guarantee strictly increasing step indices at parse time
- Unitless timestamps interpreted as Unix epoch seconds (matches PromQL)
- Deterministic result sorting before comparison (PromQL does not guarantee order)
- Label matching checks only labels present in expected samples (__name__ optional)

## Related Issues

Fixes #93 

## Test Plan

- 13 DSL parser tests (given/when/then pattern)
- 7 runner tests including integration tests
- Tests for absolute step offsets and invalid values
- Tests for brace handling, label parsing, timestamp calculation
- Includes initial test suite for @ and selectors.
## Checklist

- [x] Tests added/updated

Inspired by: https://github.com/prometheus/prometheus/tree/main/promql/promqltest


samsond:
This turns out to be a huge lift, and we need to start somewhere. I tried to capture the inspiration from [Prometheus's promqltest](
https://github.com/prometheus/prometheus/tree/main/promql/promqltest) while adapting it to our architecture.

## Current Structure

I came up with two files (dsl.rs and runner.rs) to minimize the initial code review, even though it's kind of big. The intention is to split it further if we get early feedback and like the direction.

Right now runner.rs does everything:
- Parses test files
- Loads data into TSDB
- Evaluates queries
- Compares results

I've added commented demarcation lines to show the logical separation points.

## Recommended Split (if we like the direction)
```sh
promqltest/
├── dsl.rs        # Parsing only
├── loader.rs     # load -> tsdb
├── evaluator.rs  # query -> result
├── assert.rs     # compare results
├── runner.rs     # orchestration
```

This keeps each module focused and minimal while maintaining the same functionality.

If we don't like this direction, I'm happy to pivot. Just wanted to get something working first to validate the approach before over-engineering the structure.

timeseries/src/promql/promqltest/dsl.rs
```diff
@@ -0,0 +1,731 @@
...
+
+#[derive(Debug, Clone)]
+pub enum Command {
+    Load(LoadCmd),
+    EvalInstant(EvalInstantCmd),
+    Clear(ClearCmd),
+}
+
+// ============================================================================
+// Data Structures
```
agavra, line 40: I'm wondering how many of these we can replace with our existing data structures in model.rs for example:

```rs
pub struct Sample {
    /// Timestamp in milliseconds since Unix epoch.
    ///
    /// Uses `i64` (following chrono/protobuf conventions) to support pre-1970 dates.
    pub timestamp_ms: i64,

    /// The sample value.
    ///
    /// May be NaN or ±Inf for special cases.
    pub value: f64,
}
```

and
```rs
pub struct Series {
    /// Labels identifying this series, including `__name__` for the metric name.
    pub labels: Vec<Label>,

    // --- Metadata (last-write-wins) ---
    /// The type of metric (gauge or counter).
    pub metric_type: Option<MetricType>,

    /// Unit of measurement (e.g., "bytes", "seconds").
    pub unit: Option<String>,

    /// Human-readable description of the metric.
    pub description: Option<String>,

    // --- Data ---
    /// One or more samples to write.
    pub samples: Vec<Sample>,
}
```

timeseries/src/promql/promqltest/dsl.rs
```diff
@@ -0,0 +1,731 @@
...
+// ============================================================================
+
+#[derive(Debug, Clone)]
+pub struct SeriesLoad {
+    pub metric: String,
```
agavra: note that we typically treat `metric` as just another label (`__name__`)

timeseries/src/promql/promqltest/dsl.rs
```diff
@@ -0,0 +1,731 @@
...
+pub struct EvalResult {
+    pub labels: HashMap<String, String>,
+    pub value: f64,
+}
+
+// ============================================================================
+// Parser
+// ============================================================================
+
+pub struct Parser;
```
agavra, line 70: I didn't review this in depth, the tests are proof enough IMO that it works and it's not used in prod so that's fine. LMK if there's something in particular you wanted my attention on

timeseries/src/promql/promqltest/README.md
```diff
@@ -0,0 +1,152 @@
...
+cargo test -p timeseries run_all_promql_tests
+
+# Show individual test file results
+cargo test -p timeseries run_all_promql_tests -- --nocapture
+
+# Run a specific test file (requires test function in mod.rs)
+cargo test -p timeseries promql::promqltest::tests::at_modifier
+```
+
+## Test File Format
```
agavra, line 54: is there a way to annotate a certain test as "ignore"? it would be nice for us to add a `ignore` directive that ignores statements until the next clear. that way we can copy files over directly from prometheus' test suite and just ignore tests that fail. this will help us identify our % coverage

timeseries/src/promql/promqltest/README.md
```diff
@@ -0,0 +1,152 @@
...
+   }
+   ```
+
+The test is automatically discovered and executed by `run_all_promql_tests()`.
+
+## Running Tests
+
+```bash
+# Run all discovered tests
+cargo test -p timeseries run_all_promql_tests
```
agavra: nit: for all of these we need to use `-p opendata-timeseries` (we had to rename the crates to publish them)

timeseries/src/promql/promqltest/runner.rs
```diff
@@ -0,0 +1,503 @@
...
+    files.sort();
+    Ok(files)
+}
+
+// ============================================================================
+// Test Runner (Orchestration)
+// ============================================================================
+
+/// Run all embedded test files (matches Prometheus RunBuiltinTests)
+pub async fn run_builtin_tests() -> Result<(), String> {
```
agavra, line 38: we can do this in a follow-up, but it would be nice if we could have the test runner list the individual tests instead of having it be on big test result:
```
running 1 test
test promql::promqltest::tests::run_all_promql_tests ... ok
```
Ideally it would list individually that it ran both `at_modifier` and `selectors`

timeseries/src/promql/promqltest/runner.rs
```diff
@@ -0,0 +1,503 @@
...
+{
+    let commands = parse_test_file(content)?;
+    let mut tsdb = storage_factory();
+    let mut eval_count = 0;
+
+    for cmd in commands {
+        match cmd {
+            Command::Clear(_) => {
+                // Create fresh TSDB instance - clears all state including caches and snapshots
+                tsdb = storage_factory();
```
agavra, line 94: we should make sure to close the old one, slatedb leaks some resources otherwise (I think) and we should do the same at the end of the for loop

timeseries/src/promql/promqltest/runner.rs
```diff
@@ -0,0 +1,503 @@
...
+
+    let data = resp.data.ok_or_else(|| {
+        format!(
+            "No data in response (query: {}, status: {:?}, error: {:?})",
+            query, resp.status, resp.error
+        )
+    })?;
+    let result = data.result.as_array().ok_or("Result is not a vector")?;
+
+    // Convert JSON to structured EvalResult
```
agavra, line 238: unrelated to this PR (I'll create an issue to track this) but we should probably have a native, non-JSON return value for the embedded result of the query and only convert it to JSON at the router level

timeseries/src/promql/promqltest/runner.rs
agavra: Here's a patch that accomplishes that: 

[many-tests.patch](https://github.com/user-attachments/files/25242085/many-tests.patch)

When you run all the tests you see individual lines for each one

```
  test promql::promqltest::tests::should_pass_at_modifier ... ok
  test promql::promqltest::tests::should_pass_selectors ... ok
```

timeseries/src/promql/promqltest/README.md
```diff
@@ -0,0 +1,152 @@
...
+   async fn my_feature() {
+       run_test("my_feature", include_str!("testdata/my_feature.test"))
+           .await
+           .unwrap();
+   }
+   ```
+
+The test is automatically discovered and executed by `run_all_promql_tests()`.
+
+## Running Tests
```
agavra, line 44: if anyone else reviews this I confirmed test failures are clear to understand by intentionally adding a failure:

```
---- promql::promqltest::tests::should_pass_at_modifier stdout ----

thread 'promql::promqltest::tests::should_pass_at_modifier' (10751265) panicked at /Users/agavra/dev/opendata/target/debug/build/opendata-timeseries-7adab3c01f734cab/out/promql_tests_generated.rs:6:10:
called `Result::unwrap()` on an `Err` value: "at_modifier eval #1 (query: metric @ 100): Value mismatch: expected 30, got 20"
```

it would be extra nice if we could output a line number, but for now `eval #1` is sufficiently descriptive

timeseries/src/promql/promqltest/README.md
```diff
@@ -0,0 +1,152 @@
...
+cargo test -p timeseries run_all_promql_tests -- --nocapture
+
+# Run a specific test file (requires test function in mod.rs)
+cargo test -p timeseries promql::promqltest::tests::at_modifier
+```
+
+## Test File Format
+
+```
+# Load time series data
```
rodesai: Is the format here different from the prometheus one? If not, shall we just point to that specification?

rodesai (approved):
Thanks @samsond! This looks great and is super valuable

timeseries/src/promql/promqltest/README.md
samsond: I really like this idea. It would give us a clean way to track unimplemented features without commenting out or modifying upstream tests.

timeseries/src/promql/promqltest/README.md
samsond: Addressed.

timeseries/src/promql/promqltest/runner.rs
samsond: That makes sense, I will use the patch.

timeseries/src/promql/promqltest/README.md
samsond: Good call, I'll link to Prometheus's spec and only document our ignore directive extension.

timeseries/src/promql/promqltest/dsl.rs
samsond: Good call, it will help us be consistent.

timeseries/src/promql/promqltest/dsl.rs
samsond: @agavra I looked into reusing model::Sample and model::Series, but there are semantic differences that make the DSL types distinct.

SeriesLoad vs model::Series

- SeriesLoad stores (step_index, value) — step indices are test DSL concepts that later get converted into timestamps based on the load interval.
- model::Series stores (timestamp_ms, value) absolute timestamps.
- The conversion happens during ingestion (timestamp = base_time + step_index × interval), so reusing model::Series would force premature timestamp materialization.

ExpectedSample / EvalResult vs model types

- Test types use HashMap<String, String> to allow flexible label matching. For example, {job="test"} matches any metric with that label, ignoring __name__.
- model::Series uses Vec<Label> and enforces canonical ordering and exact structure.
- The HashMap approach keeps test assertions ergonomic and intentionally looser than storage semantics.

keeping DSL types separate:
- Preserves clear separation between test representation and storage model
- Prevents coupling the test framework to future production model changes
- Makes test semantics explicit (step-based vs timestamp-based)

My preference is to keep the DSL representation isolated and convert at the ingestion boundary. That said, I can add lightweight conversion helpers if tighter integration is preferred.

timeseries/src/promql/promqltest/runner.rs
samsond: @agavra I see Storage has a close() method, but Tsdb doesn't expose one. 

Should I:
1. Add pub(crate) async fn close(&self) to Tsdb that calls self.storage.close().await?
2. Or is there a different cleanup pattern you prefer?

Once confirmed, I'll add the close calls before creating a new instance and at the end of the test loop.

agavra: Ah let's do that as a follow-up then. We should defintiely have a public close method. I'll create a ticket for that.

agavra: https://github.com/opendata-oss/opendata/issues/182

timeseries/src/promql/promqltest/dsl.rs
agavra: That is reasonable. Let's keep them separate since the data model here is pretty well contained.

timeseries/src/promql/promqltest/runner.rs
samsond: perfect.

timeseries/src/promql/promqltest/README.md
samsond: I tried implementing ignore and using clear to stop ignoring, but that surfaced an interesting issue.

Right now, ignore can only be stopped by clear, and clear also wipes all loaded data. That makes it easy to forget that data must be reloaded afterward, which feels error-prone and slightly confusing.

So I’m thinking of separating concerns more explicitly:

* `ignore`: start skipping commands
* `resume`: stop skipping, keep loaded data
* `clear`: Wipe all loaded data (doesn't affect ignore state)

This keeps control flow (ignore/resume) separate from storage state (clear) and avoids accidental data loss. I’ll document this behavior clearly in the README.

samsond:
Cool I will create a separate ticket to split runner.rs and also flesh out the test suite.

PR #178, opened 2026-02-11
[log] Expose append APIs to enable backpressure semantics

hachikuji:
We modify the log API to expose two methods which mirror the write APIs on the coordinator:

- `try_append`: returns immediately if the queue is full
- `append_timeout`: blocks for up to the timeout while waiting to add to the queue.

I've added an `AppendError` which returns the records in the case of a timeout or a full queue so that the write can be retried.

I've also modified the append semantics to separate durability as a concern. A successful append means that the records have been accepted into the system. However, the records are not guaranteed durable until a call to `flush()`. This begs the question of how a flush failure will be handled. How will the user know which records remained in the queue? What I was considering for a subsequent patch is to add a `FlushError` which returns an iterator over whatever records could not be confirmed durable before the system failed. The user can then reopen the log and retry these records if they wish, or use some other recovery approach.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2384,7 +2384,7 @@ checksum = "d6790f58c7ff633d8771f42965289203411a5e5c68388703c06e14f24770b41e"
 
 [[package]]
 name = "opendata-common"
-version = "0.1.3"
+version = "0.1.4"
 dependencies = [
  "async-trait",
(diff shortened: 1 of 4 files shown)


agavra (approved):
Clever returning the vec of records with the error type. LGTM

PR #180, opened 2026-02-11
Add tsdb regression benchmark

ryandielhenn:
## Summary

What does this PR do?

Adds ingestion throughput benchmark for `tsdb` using the `bencher` framework. Measures samples/sec and series/sec across varying batch sizes and number of labels per series. 

## Test Plan

How was this tested?

cargo run --bin timeseries-bench

## Checklist

- [ ] Tests added/updated (benchmark serves as test)

### DIFF
--- a/timeseries/bench/Cargo.toml
+++ b/timeseries/bench/Cargo.toml
@@ -6,7 +6,7 @@ publish = false
 
 [dependencies]
 bencher.workspace = true
-timeseries = { path = "..", package = "opendata-timeseries", features = ["testing"] }
+timeseries = { path = "..", package = "opendata-timeseries"}
 anyhow.workspace = true
 async-trait.workspace = true
 tokio.workspace = true


ryandielhenn:
I'm seeing some performance results that don't make much sense - most likely my ingestion process has errors in this benchmark and will keep looking into it but thought I'd share in case anyone has ideas.

Performance results are erratic - no consistent pattern between number of series ingested and ingested samples/sec.

- num_series=10:   720K samples/sec
- num_series=100:  1.23K samples/sec (500x slower!)
- num_series=1000: 571K samples/sec

Will investigate further.

ryandielhenn:
@agavra 
@hachikuji 

The benchmark runs successfully but shows platform-specific performance issues for specific configurations across multiple runs:

**On Linux:**
Slow configurations:
- num_series=100, num_samples=100, num_labels=5: ~200 samples/sec (47-55s per run)
- num_series=100, num_samples=1000, num_labels=5: ~1K-2K samples/sec (48-100s per run)

Fast configurations (for comparison):
- num_series=100, num_samples=100, num_labels=2: ~600K samples/sec
- num_series=100, num_samples=100, num_labels=10: ~300K samples/sec
- num_series=100, num_samples=100, num_labels=20: ~250K samples/sec

**On macOS:**
The same num_labels=5 configurations are slower than other label counts but not as slow as on my Linux machine (81K vs 200 samples/sec on Linux).

The Linux slowdown is reproducible across multiple runs on my machine - specifically num_labels=5 with num_series=100 and num_samples>=100. Not sure if this is specific to my environment or a broader issue - would appreciate any insights.

timeseries/src/tsdb.rs
```diff
@@ -18,7 +18,7 @@ use crate::util::Result;
 ///
 /// Tsdb manages multiple MiniTsdb instances (one per time bucket) and provides
 /// a unified QueryReader interface that merges results across buckets.
-pub(crate) struct Tsdb {
+pub struct Tsdb {
```
hachikuji: We probably don't want to expose this (similar for the other ones). I think the plan is to migrate over to `TimeSeries` which will be public. I have been meaning to get to this, just not quite there yet.

ryandielhenn: I can wait for the name change and update this PR once it's done. @agavra mentioned to use `Tsdb`  for now, so I did it as a quick way to get the benchmark working but was planning on it being temporary. I can help with that name change if needed.

In regards to the storage configuration I'm using `InMemoryStorage` with the `OpenTsdbMergeOperator`, a pattern I've seen elsewhere in the codebase.

```rust
async fn create_test_storage() -> Arc<dyn Storage> {
    Arc::new(InMemoryStorage::with_merge_operator(Arc::new(
        OpenTsdbMergeOperator,
    )))
}
````

I just built the benchmark again with `cargo build --release --bin timeseries-bench` and ran the benchmark with `perf record -g ./target/release/timeseries-bench`. I checked the `perf report`  and it looks like >80% of time is being spent in `bytes::shared_drop` (~40%) & `bytes::shared_clone` (~44%). I know the `OpenTsdbMergeOperator` uses bytes quite a bit but not sure the exact problem. 

On the upside the other parameter sets that weren't problematic sped up quite a bit when compiling with the `--release` flag.

```
Running benchmark: ingest
  [benchmark=ingest, commit=f5b0ce3422ecc1aff8ab57a8a3749c298bbf47d8, branch=bench/timeseries, num_labels=5, num_samples=100, num_series=10]
    samples_per_sec  3.42M
    series_per_sec   34.15K
    elapsed_ms       5.00K
```

**Comment by ryandielhenn** [comment]
I changed the Tsdb visibility and all of it's dependencies back to crate since I noticed @apurvam  added `testing.rs`. I also realized that the slowdown was being caused because I was generating new labels for each series inside of the benchmark loop. I've changed the code so that the labels are pre-generated before the benchmark starts. This way just ingestion is being tested not label generation. 

I would appreciate any insight on how to improve further, but these results seem much better - 

## Timeseries Benchmark Results

**Commit:** `c55c288` | **Branch:** `bench/timeseries`

### Ingest Performance

| Series | Labels | Samples | Samples/sec | Series/sec | Duration |
|--------|--------|---------|-------------|------------|----------|
| 10     | 5      | 100     | 3.96M       | 39.56K     | 5.00s    |
| 100    | 5      | 100     | 3.43M       | 34.27K     | 5.00s    |
| 1000   | 5      | 100     | 4.19M       | 41.88K     | 5.01s    |
| 100    | 2      | 100     | 5.45M       | 54.46K     | 5.00s    |
| 100    | 10     | 100     | 2.47M       | 24.69K     | 5.00s    |
| 100    | 20     | 100     | 1.47M       | 14.71K     | 5.00s    |
| 100    | 5      | 10      | 2.64M       | 263.62K    | 5.00s    |
| 100    | 5      | 1000    | 4.34M       | 4.34K      | 5.00s    |

File: timeseries/bench/src/ingest.rs [hunk 2]
```diff
@@ -0,0 +1,137 @@
...
+        // Bucket at minute 60 covers ms [3_600_000, 7_200_000), a 3_600_000 ms range.
+        // We keep all sample timestamps inside this single bucket so the benchmark
+        // measures steady-state ingest rather than bucket-creation overhead.
+        let bucket_start_ms: i64 = 3_600_000;
+        let bucket_range_ms: i64 = 3_600_000;
+
+        // Start the timed benchmark
+        let runner = bench.start();
+
+        let tsdb: TestTsdb = create_test_tsdb().await;
```
apurvam: Hi @ryandielhenn , this method only creates an inmemory slateDB since the point was to support integration tests on the HTTP API:

```rust
    let config = StorageConfig::SlateDb(SlateDbStorageConfig {
        path: "test-data".to_string(),
        object_store: ObjectStoreConfig::InMemory,
        settings_path: None,
    });
```

For a benchmark, we may want to parameterize it so that it it mimics the config for production.

ryandielhenn:
Hey @apurvam, Thanks for the feedback! The benchmark now uses `LocalObjectStoreConfig` with a temp directory so we get real disk I/O instead of in-memory. Would you also want the option to run benchmarks against an S3 bucket? I don't have access to AWS credentials currently but I could write up what I think will work. I will be back home around 8pm today and will work on it tomorrow if I don't get to it tonight.

Still getting good performance with the change.

timeseries/src/testing.rs
```diff
@@ -66,6 +70,29 @@ pub async fn create_test_tsdb() -> TestTsdb {
     }
 }
 
+/// Create a [`TestTsdb`] with a caller-provided object store config.
+///
+/// This allows benchmarks to use production-like storage backends
+/// such as local filesystem or S3 instead of in-memory.
+pub async fn create_test_tsdb_with_config(object_store: ObjectStoreConfig) -> TestTsdb {
```
apurvam, line 62: Nit: in the spirit of DRY, the `create_test_tsdb` function should call this one with the inmemory config as a parameter.

apurvam (commented):
Thanks @ryandielhenn . I'd like to let @hachikuji take a look, but this looks good to me. The main thing I'd like his feedback on is how to parameterize the storage configs. It looks to me like the log benchmarks use an inmemory config by default, but I also know that they were have been run against S3.

If we can emulate that pattern with the tsdb benchmarks, then I think this is good to merge for now and we can address any follow up comments later.

ryandielhenn:
Thanks @apurvam, you should be able to use `cargo run --bin timeseries-bench -- --config s3config.toml` to use S3 and the benchmark will default `LocalObjectStoreConfig` if no config is specified.

ryandielhenn:
Just got an AWS account after finding out they have 6 months of free services! I spotted a bug after testing against S3 but the bug is fixed now and the benchmarks against it work. Not only do they work but getting similar performance to local disk IO. @apurvam @hachikuji

PR #186, opened 2026-02-12
add a skeleton for index rebalancer which executes lire rebalancing

rodesai:
## Summary

Adds a skeleton for the lire-based index rebalancer:
- Add an enum for lire rebalance ops (split/merge)
- Add an enum for rebalance step commands as described in the rfc (RebalanceCommand)
- change the vector write message to an enum of either a vector write batch or a rebalance step command (RebalanceCommand)
- add a write channel to the vector coordinator for rebalance commands
- add IndexRebalancer which accepts rebalance ops and executes them

bootstrap the db with a single centroid by default

add a readable view for the current delta

## Checklist

- [x] Tests added/updated

### DIFF
--- a/common/src/coordinator/handle.rs
+++ b/common/src/coordinator/handle.rs
@@ -1,4 +1,4 @@
-use super::WriteCommand;
+use super::{BroadcastedView, WriteCommand};
 use super::{Delta, Durability, WriteError, WriteResult};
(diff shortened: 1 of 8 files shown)


rodesai:
I'm going to merge this. I'll look out for late-arriving feedback

PR #197, opened 2026-02-13
[log] Add view subscriber and move read sync to the read path

hachikuji:
This patch hooks the read semantics provided by the write coordinator into `LogDb`. Current semantics are a bit limited (no support for durable reads), but it is straightforward to implement Durability::Flushed. This guarantees that all data written prior to an explicit flush() will be visible to readers. To enable this, I've added a common::ViewSubscriber which encapsulates the epoch bookkeeping (basically through EpochWatermarks). In `LogDb`, we use a separate task to poll the subscriber and update local state. Our reader methods show the pattern for achieving Flushed visibility. We grab the current flushed epoch from the coordinator and then wait until the reader is caught up to it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2400,6 +2400,7 @@ dependencies = [
  "bytes",
  "criterion",
  "futures",
+ "prometheus-client",
  "proptest",
  "serde",
  "serde_yaml",
@@ -2501,6 +2502,8 @@ dependencies = [
  "roaring 0.10.12",
  "rstest",
  "tokio",
+ "tokio-util",
+ "tracing",
  "usearch",
 ]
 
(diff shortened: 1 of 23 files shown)


common/src/coordinator/subscriber.rs
```diff
@@ -0,0 +1,262 @@
+//! View subscription and progress monitoring.
+//!
+//! When a caller subscribes to a [`WriteCoordinator`](super::WriteCoordinator),
```
apurvam, line 3: I'm not as plugged into the WriteCoordinator design, but these Views are new concepts that were not described in the original RFC right? I just want to make sure I didn't miss anything.

rodesai: Yeah, I don't think the original RFC detailed exactly how reads of the in-flight deltas would work.

rodesai (approved):
LGTM!

Issue #193, opened 2026-02-12
Collect basic SlateDB  metrics


apurvam:
## Summary

We need to collect and export SlateDB metrics from opendata services. Without visibility into what SlateDB is doing, It's hard to maintain opendata services in production.

## Proposed Solution

We should add a library that collects metrics for slatedB, and each opendata database can export these metrics via their existing prometheus scrape endpoints. In the future, a metrics registry can be a common component that includes the salient SlateDB metrics by default.

PR #198, opened 2026-02-13
Add SlateDB Metrics

apurvam:
## Summary

Collect SlateDB metrics and make them available in the `/metrics` api of `Log` and `Timeseries`. To achieve this, we create a Prometheus Gauge wrapper around SlateDB's `ReadableStat`, and register these wrapped gauges in the Prometheus registry of the Log and TimeSeries. Every scrape reads directly from SlateDB for its metrics. 

## Related Issues

Fixes https://github.com/opendata-oss/opendata/issues/193

## Test Plan

Apart from unit tests, added an integration test for the Log which checks that:

1. SlateDB metrics are returned in the /metrics response.
2. SlateDB Metrics are as expected after a write to the log.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/common/src/storage/slate.rs
+++ b/common/src/storage/slate.rs
@@ -12,6 +12,28 @@ use slatedb::{
     MergeOperatorError, WriteBatch, config::WriteOptions as SlateDbWriteOptions,
 };
 
+/// Thin wrapper that exposes a SlateDB [`ReadableStat`] as a Prometheus gauge.
+///
+/// Instead of snapshotting stats into an intermediate representation and
+/// refreshing gauges, this reads the live atomic value on each encode/scrape.
+#[cfg(feature = "metrics")]
+#[derive(Debug)]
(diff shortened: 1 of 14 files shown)


common/src/storage/slate.rs
```diff
@@ -12,6 +12,28 @@ use slatedb::{
...
+impl prometheus_client::encoding::EncodeMetric for ReadableStatGauge {
+    fn encode(
+        &self,
+        mut encoder: prometheus_client::encoding::MetricEncoder,
+    ) -> Result<(), std::fmt::Error> {
+        encoder.encode_gauge(&self.0.get())
+    }
+
+    fn metric_type(&self) -> prometheus_client::metrics::MetricType {
+        prometheus_client::metrics::MetricType::Gauge
```
agavra, line 33: nit: let's create a ticket to update this after SlateDB with https://github.com/slatedb/slatedb/pull/1283 is released

apurvam: Done. #201

PR #199, opened 2026-02-13
add the ability to pause/unpause write channels

rodesai:
## Summary

adds the ability to pause/unpause write channels consumed by the write coordinator. Callers can get a pause handle for any named channel read by the coordinator, and use the returned handle to pause further consumption of that channel, or unpause to resume consumption. This is useful for dbs that need to implement more sophisticated backpressure mechanisms. For example, in vector we need to apply backpressure when indexing is not able to keep up with the rate of ingestion

## Test Plan

I'm testing it when running vector with lire. The existing unit tests do a good job of asserting there isn't a regression.

## Checklist

- [n/a] Tests added/updated

### DIFF
--- a/common/src/coordinator/mod.rs
+++ b/common/src/coordinator/mod.rs
@@ -2774,10 +2774,7 @@ mod tests {
         // then
         // make sure the data only contains keys from b (so a was never processed)
         let data = coordinator.view().current.data.lock().unwrap().clone();
-        let mut expected = (0..1000)
-            .into_iter()
-            .map(|i| format!("b{}", i))
-            .collect::<HashSet<_>>();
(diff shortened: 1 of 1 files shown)


common/src/coordinator/mod.rs
```diff
@@ -600,6 +612,55 @@ impl<D: Delta> BroadcastedViewInner<D> {
     }
 }
 
+struct PausableReceiver<D: Delta> {
```
agavra, line 615: mind adding a quick multi-threaded unit test for this that makes sure we don't receive something after calling pause (and then resuming)? just triple check that there aren't any deadlock potentials. Don't see why there would be given they're two different channels.

PR #203, opened 2026-02-13
(timeseries) add integration tests

### LINKED ISSUE #14: tsdb p8s server match[] parameters cannot be parsed over curl

apurvam:
## Summary

fixes #14 

* Add integration tests for the timeseries HTTP server using Axum's oneshot() infrastructure with a real SlateDB-backed TSDB (in-memory object store)
* Fix a latent bug where `match[]` query parameters on `/api/v1/series`, `/api/v1/labels`, `/api/v1/label/{name}/values`, and ` /federate` were completely broken — `serde_urlencoded` cannot deserialize repeated keys into `Vec<String>`, so any  request with `match[]` would fail with a 400. The newly added integration tests exposed this.
* Replace axum's built-in Query/Form extractors with axum-extra's versions, which use serde_html_form and natively handle repeated keys like `match[]=a&match[]=b`

## Test Plan

This is patch only adds test cases.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/server.rs
+++ b/timeseries/src/promql/server.rs
@@ -1,13 +1,13 @@
 use std::net::SocketAddr;
 use std::sync::Arc;
 
+use axum::extract::Request;
 use axum::extract::{FromRequest, Path, State};
 use axum::http::{Method, StatusCode, Uri, header};
 use axum::response::{IntoResponse, Redirect, Response};
 use axum::routing::get;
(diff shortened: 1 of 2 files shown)


agavra (approved):
Nice! This is a much cleaner fix to #14 than hand-coding something, didn't realize that existed in axum. 

Skimmed testing code and LGTM

PR #204, opened 2026-02-13
fix flaky pause test

agavra:
see: https://github.com/opendata-oss/opendata/actions/runs/22004762183/job/63585799111

I was able to reproduce this by adding a sleep() after coordinator.start() but before pause_handle.stop()


apurvam (approved):
LGTM to fix the flaky build. We can find better solutions later.

agavra:
@apurvam I think this is the right solution, the problem (IMO) was in the test not the code

PR #205, opened 2026-02-14
feat(promql): implement range vector functions with Prometheus parity

### LINKED ISSUE #87: [tsdb] support the full suite of range vector fns
## Summary

PromQL supports a large set of range vector fns (see https://prometheus.io/docs/prometheus/latest/querying/functions/). We should support them all from our expression evaluator.

## Motivation

Full PromQL support.

## Proposed Solution

We need to add the functions to the function registry.

samsond:
Add 5 core range vector functions (sum, avg, min, max, count over time) with exact Prometheus semantics including IEEE-754 bit-level compatibility.

- sum_over_time: Kahan summation with Neumaier improvement
- avg_over_time: Hybrid direct/incremental mean with overflow protection
- min_over_time: Manual loop for correct NaN handling
- max_over_time: Manual loop for correct NaN handling
- count_over_time: Simple sample counting

**Numerical Stability:**
- Kahan summation with Neumaier improvement for sum/avg
- Proper parentheses to prevent compiler reordering
- #[inline(never)] to lock IEEE-754 semantics
- Bitwise exact match with Prometheus (verified by test)

**NaN Handling:**
- Real numbers replace NaN in min/max
- All-NaN input returns NaN (not ±inf)
- Manual loops avoid fold() edge cases

**Overflow Protection:**
- avg_over_time switches to incremental mean on overflow
- sum_over_time returns inf directly (skips compensation)
- Matches Prometheus hybrid strategy

**Time Range Handling:**
- Clamps negative timestamps to UNIX_EPOCH
- Handles @ modifier + offset combinations
- Prevents panic on queries before UNIX_EPOCH
- Note: Workaround for SystemTime limitation; Prometheus uses int64 ms

**Generic aggr_over_time helper:**
- Applies aggregation function to each series
- Skips empty series (Prometheus behavior)
- Returns single sample per series at eval_timestamp_ms

**Pure aggregation functions:**
- avg_kahan: Hybrid mean calculation
- Closures for simple operations (min, max, count)

**Unit tests (17 total):**
- Basic functionality for all 5 functions
- Catastrophic cancellation (Kahan correctness)
- Bitwise exact match with Prometheus
- NaN replacement in min/max
- All-NaN edge cases

**Integration tests:**
- Upstream Prometheus test suite (at_modifier.test)
- Tests with @ modifier and offset combinations
- All passing 

- TODO: Histogram support (marked in all functions)
- TODO: Additional range functions (stddev, quantile, etc.)
- TODO: Iterator-based evaluation for memory efficiency
- TODO: Replace SystemTime with int64 milliseconds (like Prometheus) to natively support negative timestamps instead of clamping to UNIX_EPOCH

Closes #87 (partial)

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/functions.rs
+++ b/timeseries/src/promql/functions.rs
@@ -3,22 +3,25 @@ use std::collections::HashMap;
 use super::evaluator::{EvalResult, EvalSample, EvalSamples};
 use crate::model::Sample;
 
-/// Kahan summation increment with Neumaier improvement
-/// Performs addition of a value to a running sum with compensation
+/// Kahan summation increment with Neumaier improvement (1974)
+///
+/// Performs compensated summation to minimize floating-point rounding errors.
+/// The Neumaier variant handles the case where the next term is larger than
+/// the running sum, which the original Kahan algorithm (1965) did not address.
+///
 /// Returns (new_sum, new_compensation)
 #[inline(never)]
 // Important: do NOT inline.
-// Prometheus observed precision regressions when the compiler reordered
-// floating-point operations during inlining. We lock behavior to match
-// Prometheus' IEEE-754 semantics exactly.
-// See: https://github.com/prometheus/prometheus/issues/16714
+// Compiler reordering of floating-point operations can cause precision loss.
+// This was observed in Prometheus (issue #16714) and we lock the behavior
+// to maintain IEEE-754 semantics exactly.
 fn kahan_inc(inc: f64, sum: f64, c: f64) -> (f64, f64) {
     let t = sum + inc;
 
     let new_c = if t.is_infinite() {
         0.0
     } else if sum.abs() >= inc.abs() {
-        // Neumaier improvement: swap if next term larger than sum
+        // Neumaier improvement: swap roles when next term is larger
         c + ((sum - t) + inc)
     } else {
         c + ((inc - t) + sum)
@@ -483,6 +486,7 @@ impl RangeFunction for CountOverTimeFunction {
 mod tests {
     use super::*;
     use crate::model::Sample;
+    use rstest::rstest;
     use std::collections::HashMap;
 
     fn create_sample(value: f64) -> EvalSample {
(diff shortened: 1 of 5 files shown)


timeseries/src/promql/evaluator.rs
```diff
@@ -505,15 +505,32 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
+        // See: https://github.com/prometheus/prometheus/blob/main/model/timestamp/timestamp.go
+        //
+        // Rust's SystemTime cannot represent times before UNIX_EPOCH. Subtracting
+        // a range may therefore underflow. To avoid panic, we clamp to UNIX_EPOCH.
+        //
+        // This is a semantic deviation from Prometheus. In practice it is harmless
+        // unless data exists before UNIX_EPOCH.
+        //
+        // TODO: Replace SystemTime with i64 millisecond timestamps across the
+        // evaluator to achieve full PromQL time semantics and remove this clamp.
```
agavra, line 524: I'm totally OK with this limitation

timeseries/src/promql/functions.rs
```diff
@@ -1,6 +1,107 @@
...
+/// Kahan summation increment with Neumaier improvement
+/// Performs addition of a value to a running sum with compensation
+/// Returns (new_sum, new_compensation)
+#[inline(never)]
+// Important: do NOT inline.
+// Prometheus observed precision regressions when the compiler reordered
+// floating-point operations during inlining. We lock behavior to match
+// Prometheus' IEEE-754 semantics exactly.
+// See: https://github.com/prometheus/prometheus/issues/16714
+fn kahan_inc(inc: f64, sum: f64, c: f64) -> (f64, f64) {
```
agavra, line 18: huh pretty cool, I hadn't heard of Kahan sums before today!

timeseries/src/promql/functions.rs
```diff
@@ -346,6 +584,101 @@ mod tests {
         assert_eq!(result[0].labels, labels);
     }
 
+    #[test]
+    fn should_apply_sum_over_time_function() {
```
agavra, line 592: since we're implementing some pretty nuanced and easy to get wrong I'd feel more comfortable if these tests had a more comprehensive testing mechanism. 

One way we can do this is to use proptest (https://github.com/proptest-rs/proptest) to generate a bunch of test cases and then a high-precision oracle to get close-to accurate estimations. It seems like the Kahan summation has a proven error bound we can leverage (https://en.wikipedia.org/wiki/Kahan_summation_algorithm). 

According to Claude the error bound is 

```|computed_sum - true_sum| ≤ 2ε · Σ|xᵢ| + O(nε²)```

Where ε is `f64::EPSILON`. A good way to test is to just write the proptest and see if any of the generated sequences fails the error bound.

This would apply for sum/avg. The others are easier to compute.

File: timeseries/src/promql/functions.rs [hunk 4]
```diff
@@ -382,4 +715,183 @@ mod tests {
...
+        // package main
+        //
+        // import (
+        //     "fmt"
+        //     "math"
+        //     "github.com/prometheus/prometheus/util/kahansum"
+        // )
+        //
+        // func main() {
+        //     values := []float64{1e16, 1.0, -1e16}
```
agavra, line 775: it would be good to tests various different values, not sure what a good amount is but having ~10 test cases would make me feel a little more comfortable hand-rolling this type of stuff

timeseries/src/promql/promqltest/testdata/at_modifier.test
apurvam, line 113: Could some of these be unignored as well?

timeseries/src/promql/functions.rs
```diff
@@ -1,6 +1,107 @@
...
+
+/// Kahan summation increment with Neumaier improvement
+/// Performs addition of a value to a running sum with compensation
+/// Returns (new_sum, new_compensation)
+#[inline(never)]
+// Important: do NOT inline.
+// Prometheus observed precision regressions when the compiler reordered
+// floating-point operations during inlining. We lock behavior to match
+// Prometheus' IEEE-754 semantics exactly.
+// See: https://github.com/prometheus/prometheus/issues/16714
```
apurvam: Given that we are handrolling a bunch of very nuanced code here, I would love to undertand why we need to do it. Is it performance, precision, or just copying the method prometheus uses to maintain compatibility, or something else?

samsond: Yes @apurvam its mainly for precision. Floating-point math is subtle, and preventing inlining keeps the operation order deterministic to avoid rounding regressions.

timeseries/src/promql/functions.rs
samsond: Yes I had to learn it as well, and it helps us with floating-point precision.

timeseries/src/promql/promqltest/testdata/at_modifier.test
samsond: Those tests are still ignored because they use label_replace(), which is blocked on two things we haven't implemented  yet:

1. Currently function call with string literal arguments is explicitly rejected as noted here [lines 783-790 in evaluator.rs](https://github.com/opendata-oss/opendata/blob/main/timeseries/src/promql/evaluator.rs#L766-L775). 
2. The `label_replace` function itself even after string argument support is added, we'd still need to implement the actual label_replace logic

timeseries/src/promql/functions.rs
samsond: Good idea it will definitely give us more confidence in the numerical correctness.

timeseries/src/promql/functions.rs
samsond: cool, I will use a parameterized test and cover multiple value combinations(similar to what I did in SlateDB).

timeseries/src/promql/functions.rs
samsond: Since we already use proptest, I've extended it to generate randomized sequences for sum_over_time and avg_over_time. I've also introduced a high-precision oracle using [rug](https://docs.rs/rug/latest/rug/) (arbitrary-precision arithmetic via GMP) to validate the Kahan error bound against ground truth.

PR #206, opened 2026-02-14
[log] Progress toward reading writes 

hachikuji:
Putting this up as a draft. I was working on integrating the coordinator view API further so that we could get read-your-writes in `LogDb`. The main progress in this patch is giving the read side access to the frozen delta state. I did some refactoring along the way that I'll probably split up into separate PRs.

Summary                                                                                                                               
                                                                                                                                        
  - Introduce delta_reader module — new DeltaReaderView that merges storage snapshots with frozen (unflushed) deltas to provide read
  visibility into writes that have been applied but not yet flushed. Includes FrozenViewLogIterator for iterating entries from frozen   
  deltas and FrozenLogDeltaView for broadcasting segment/entry metadata to subscribers. Supports scan_entries, list_keys (with   
  deduplication across storage + frozen deltas), and list_segments (merging cached + frozen segments).                                  
  - Rename delta.rs → delta_writer.rs — separates the write-side delta logic (LogDelta, LogFlusher, LogWrite, FrozenLogDelta) from the
  new read-side module. The Delta::FrozenView associated type now produces Arc<FrozenLogDeltaView> to enable cheap sharing with read
  subscribers.                                                                                                                          
  - Simplify LogDb read path — LogDb now holds a DeltaReaderView instead of LogReadView. The flush subscriber passes the full    
  coordinator View to DeltaReaderView::update_view() rather than manually patching snapshots and segments. Read operations (scan,  
  list_keys, list_segments) delegate to DeltaReaderView.
  - Remove ScanOptions / scan_with_options — LogRead::scan is now the only scan method (the options indirection wasn't pulling its
  weight).
  - Eliminate LogStorage wrapper struct — LogDb and LogFlusher now store Arc<dyn Storage> directly. Storage methods (close, snapshot,
  flush, put_with_options) are called directly with ? since From<StorageError> for Error already exists. LogStorageRead becomes a
  blanket extension trait on StorageRead instead of a concrete wrapper struct. Test-only write helpers (write_segment, write_entry,
  write_seq_block) live in a #[cfg(test)] LogStorageWrite extension trait.
  - Refactor SegmentCache / SegmentSnapshot — SegmentSnapshot is extracted as an immutable point-in-time view of segments for readers
  (with find_covering range queries), separate from SegmentCache which remains the mutable write-path cache. SegmentCache::open now
  takes &dyn StorageRead instead of &LogStorageRead.
  - Simplify LogKeyIterator / listing — list_keys on LogStorageRead now returns BTreeSet<Bytes> directly; LogKeyIterator wraps that set.
   ListingCache moves to delta_writer concerns only.

### DIFF
--- a/log/src/delta_writer.rs
+++ b/log/src/delta_writer.rs
@@ -8,9 +8,9 @@ use crate::delta_reader::FrozenLogDeltaView;
 use crate::listing::ListingCache;
 use crate::model::{AppendOutput, LogEntry, Record as UserRecord, SegmentId};
 use crate::segment::{LogSegment, SegmentCache};
-use common::Storage;
 use async_trait::async_trait;
 use bytes::Bytes;
+use common::Storage;
 use common::coordinator::{Delta, Flusher};
 use common::storage::StorageSnapshot;
 use common::{Record, WriteOptions};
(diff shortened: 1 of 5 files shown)


hachikuji:
Closing this as it is out of date and I'm separating out the independent changes which ended up lumped together here. The first is here: https://github.com/opendata-oss/opendata/pull/230.

PR #214, opened 2026-02-17
misc storage changes we need for lire

rodesai:
## Summary

Changes the posting updates constructor to use last-write-wins for dedup instead of error on duplicates. This allows multiple reassignments in a single flush (e.g. if a vector is assigned and then reassigned by a split).

Add a new fn merge_decoded_posting_lists (priority k-way merge, earlier operand wins). This will be used for the read-applied-writes view so we can merge the in-memory posting lists. Split/merge need this to figure out the set of vectors that need to move around after a split/merge without blocking the split/merge on a flush after each step.

Adds a CentroidChunk merge operator which appends new centroid entries. This allows us to add centroids to chunks as they get created.

Adds CentroidScanResult struct that returns the last centroid chunk and its size along with all the centroids. We use this to initialize the delta with the current chunk metadata so it knows how to write new centroids.

## Test Plan

unit tests

## Checklist

- [x] Tests added/updated

### DIFF
--- a/vector/src/db.rs
+++ b/vector/src/db.rs
@@ -193,9 +193,9 @@ impl VectorDb {
             .scan_all_centroids(config.dimensions as usize)
             .await?;
 
-        if !existing_centroids.is_empty() {
+        if !existing_centroids.entries.is_empty() {
             // Use existing centroids from storage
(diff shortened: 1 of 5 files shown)


rodesai:
I'm going to merge this to unblock the actual lire PR

PR #216, opened 2026-02-17
Fix http_requests_in_flight gauge leak on future cancellation

apurvam:
## Summary

The in-flight gauge was incremented in the middleware's call() but only decremented after the inner future completed successfully. When the future was cancelled (client disconnect during long-poll) or the inner service returned an error (triggering early return via ?), the dec() was skipped and the gauge leaked +1 permanently.

This can cause the inflight request gauge to grow unbounded if there the future is cancelled, as is often the case with the log service and the the fetch-with-timeout call. See:

Screenshot 2026-02-16 at 6 05 26 PM

The fix is to replace the manual dec() with an RAII InFlightGuard that calls dec() in its Drop impl. The guard is moved into the async block so it is dropped whenever the future terminates — whether by success, error propagation, or cancellation.

## How was this tested?

Added integration tests to timeseries and log to verify the at the inflight request gauge is decremented in all cases. These tests fail without the patch. 

## Checklist

- [x] Tests added/updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2467,7 +2467,7 @@ checksum = "d6790f58c7ff633d8771f42965289203411a5e5c68388703c06e14f24770b41e"
 
 [[package]]
 name = "opendata-common"
-version = "0.1.5"
+version = "0.1.6"
 dependencies = [
  "async-trait",
  "bytes",
@@ -2522,7 +2522,7 @@ dependencies = [
 
 [[package]]
 name = "opendata-timeseries"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "async-trait",
  "axum",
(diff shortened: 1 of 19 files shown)


log/src/server/middleware.rs
```diff
@@ -192,6 +206,85 @@ where
...
+        };
+
+        assert_eq!(metrics.http_requests_in_flight.get(), 0);
+
+        // when
+        let response = service.call(test_request("/test")).await.unwrap();
+
+        // then
+        assert_eq!(response.status().as_u16(), 200);
+        assert_eq!(metrics.http_requests_in_flight.get(), 0);
```
cadonna, line 241: Since you test for 0 `http_requests_in_flight` here and also on line 232, you are effectively not testing that the metrics is decremented. Is there a way to verify that `http_requests_in_flight` is incremented to 1 and then decremented to 0 when the future completes?
Similar holds for the other tests.

apurvam: Good call. Updated the tests.

PR #217, opened 2026-02-17
add ViewReader for reading the current/frozen deltas + snapshot

rodesai:
## Summary

Adds a ViewReader which allows reading the latest writes. Currently, its just intended to be used by the rebalancer to read posting lists. In the future we can extend it to support reading the other data so it can be used by queries.

Couple other minor changes that got pulled into this diff:
- Extract writer options into VectorDbDeltaOpts
- Add chunk tracking to context (current_chunk_id, current_chunk_count).
- rename delted_vectors to deleted_centroids
- support sending back results from the delta apply fn. This will be used by the rebalance commands

This patch also has changes to schedule splits from the current delta. This will be removed in a later patch and that logic will actually go into a new `IndexRebalancer` struct. I'm leaving it in to make that patch easier to rebase)

## Test Plan

Unit tests

## Checklist

- [x] Tests added/updated

### DIFF
--- a/vector/src/db.rs
+++ b/vector/src/db.rs
@@ -11,7 +11,9 @@
 //! - Delta handles dictionary lookup, centroid assignment, and builds RecordOps
 //! - Flusher applies ops atomically to storage
 
-use crate::delta::{VectorDbDeltaContext, VectorDbWrite, VectorDbWriteDelta, VectorWrite};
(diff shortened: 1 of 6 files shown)


rodesai:
merging to unblock the actual lire implementation pr

PR #219, opened 2026-02-17
vector lire split/merge rebalancing implementation

rodesai:
## Summary

Implements the core split/merge rebalancing pipeline for LIRE:

IndexRebalancer is switched to a "synchronous" split/merge execution model.
- IndexRebalancer now executes rebalance tasks directly, running each rebalance op in its own tokio tasks. It tracks the running tasks in an in-memory map.
- IndexRebalancer now also owns scheduling of the rebalance ops. It owns the in-memory centroid counts and keeps track of which centroids need a rebalance, running up to some maximum number of tasks concurrently to rebalance these. When a task finishes, or when a centroid's count changes, the rebalancer tries to schedule new rebalance tasks - adding the involved centroid to the set of pending centroids if at the max task limit.
- Finally, this patch fills in the split/merge op implementations for the rebalancer

RebalanceCommand processing through the write coordinator. Each rebalance op works by issuing a series of RebalanceCommands through the write coordinator to split, merge, and reassign vectors. This patch includes the implementations of these operations.

I also pulled in a couple somewhat orthogonal changes into this PR:

centroid management:
- When loading a db with centroids we no longer provide full entries with ids - just the vectors. This aligns with LIRE where the db assigns vector ids
- When loading the centroids from storage, we fetch the last centroid and its size and use it to populate the delta ctx so it knows where to add new centroids

The search path no longer uses the deletion bitmap. It just prunes deleted vectors using the forward index. We'll address efficient handling of updates in a later patch.

## Guide for Reviewers

1. Review `handle_split` and `handle_merge`. These are the entry points for split/merge execution and each fn synchronously executes a single split/merge of a centroid. Each op works by sending rebalance commands to the write coordinator. The commands are defined in `commands.rs`. Each command has a docstring that explains its behaviour.
2. Review the command implementations in `commands.rs` 
3. Review the scheduling logic in `IndexRebalancer`. The entry points are `update_counts` and `finish_split`/`finish_merge`. `update_counts` notifies the rebalancer about a change to cluster sizes and might trigger a split/merge. `finish_task` notifies the rebalancer that a task as completed and might schedule new splits/merges if there are any pending or the split/merge itself caused a downstream split/merge.

## Test Plan

- Unit tests for rebalance ops
- End-to-end test that evaluates recall against sift1m/cohere1m (see #221):
```bash
rohan@Rohans-MacBook-Air opendata % RUST_BACKTRACE=1 RUST_LOG=info cargo test -p opendata-vector --test sift1m sift1m_recall -- --ignored --nocapture
   Compiling opendata-vector v0.1.0 (/Users/rohan/responsive/opendata/vector)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 4.10s
     Running tests/sift1m.rs (target/debug/deps/sift1m-ca51429af0824cdb)

running 1 test
Loaded 1000000 base vectors
Written batch 10000/100000
Written batch 20000/100000
Written batch 30000/100000
Written batch 40000/100000
Written batch 50000/100000
Written batch 60000/100000
Written batch 70000/100000
Written batch 80000/100000
Written batch 90000/100000
Written batch 100000/100000
Ingested all base vectors
Loaded 10000 queries and 10000 ground truth entries
Num centroids: 628

query 1000: hnsw_recall@10 = 0.9106, exact_recall@10 = 0.9106
```

## Checklist

- [x] Tests added/updated

### DIFF
--- a/vector/src/lire/commands.rs
+++ b/vector/src/lire/commands.rs
@@ -293,11 +293,11 @@ impl VectorDbWriteDelta {
     ) -> Result<Arc<dyn Any + Send + Sync + 'static>, String> {
         let (c0_id, seq_alloc_put) = self.ctx.id_allocator.allocate_one();
         if let Some(seq_alloc_put) = seq_alloc_put {
-            self.ops.push(RecordOp::Put(seq_alloc_put));
+            self.ops.push(RecordOp::Put(seq_alloc_put.into()));
         }
         let (c1_id, seq_alloc_put) = self.ctx.id_allocator.allocate_one();
         if let Some(seq_alloc_put) = seq_alloc_put {
-            self.ops.push(RecordOp::Put(seq_alloc_put));
+            self.ops.push(RecordOp::Put(seq_alloc_put.into()));
         }
         let c0 = CentroidEntry::new(c0_id, cmd.c0.centroid_vec);
         let c0_postings = cmd.c0.postings;
--- a/vector/src/lire/rebalancer.rs
+++ b/vector/src/lire/rebalancer.rs
@@ -883,7 +883,7 @@ mod tests {
             for centroid_vec in &centroid_vecs {
                 let (centroid_id, seq_put) = allocator.allocate_one();
                 if let Some(put) = seq_put {
-                    initial_ops.push(RecordOp::Put(put));
+                    initial_ops.push(RecordOp::Put(put.into()));
                 }
                 entries.push(CentroidEntry::new(centroid_id, centroid_vec.clone()));
             }
--- a/vector/src/db.rs
+++ b/vector/src/db.rs
@@ -250,7 +250,7 @@ impl VectorDb {
         for vector in centroids {
             let (centroid_id, seq_alloc_put) = id_allocator.allocate_one();
             if let Some(seq_alloc_put) = seq_alloc_put {
-                ops.push(common::storage::RecordOp::Put(seq_alloc_put));
+                ops.push(common::storage::RecordOp::Put(seq_alloc_put.into()));
             }
             entries.push(CentroidEntry::new(centroid_id, vector));
         }
--- a/vector/src/delta.rs
+++ b/vector/src/delta.rs
@@ -886,7 +886,7 @@ mod tests {
         for (centroid_id, expected_count) in [(1u64, 2i32), (2, 1), (3, 1)] {
             let stats_key = CentroidStatsKey::new(centroid_id).encode();
             let stats_merge = frozen.ops.iter().find(|op| match op {
-                RecordOp::Merge(record) => record.key == stats_key,
+                RecordOp::Merge(record) => record.record.key == stats_key,
                 _ => false,
             });
             assert!(
@@ -896,7 +896,7 @@ mod tests {
             );
             if let Some(RecordOp::Merge(record)) = stats_merge {
                 let value = crate::serde::centroid_stats::CentroidStatsValue::decode_from_bytes(
-                    &record.value,
+                    &record.record.value,
                 )
                 .unwrap();
                 assert_eq!(


vector/src/hnsw/usearch.rs
```diff
@@ -85,9 +85,9 @@ impl UsearchCentroidGraph {
             dimensions,
             metric,
             quantization: ScalarKind::F32,
-            connectivity: 16,     // M parameter
-            expansion_add: 200,   // ef_construction
-            expansion_search: 30, // ef_search default
+            connectivity: 16,      // M parameter
```
rodesai, line 88: these are just small tweaks to the centroid graph parameters to improve recall

vector/src/lire/rebalancer.rs
```diff
@@ -1,86 +1,1609 @@
...
+            .map_err(|e| e.to_string())?;
+
+        let posting_list: PostingList = posting_list_value;
+        let vectors: Vec<(u64, Vec<f32>)> = posting_list
+            .iter()
+            .map(|p| (p.id(), p.vector().to_vec()))
+            .collect();
+
+        // If < 2 vectors, return early (can't split)
+        if vectors.len() < 2 {
```
agavra, line 424: this condition makes sense to me but I'm curious about what causes it - is there some condition that causes us to enqueue many splits for the same centroid that pushes that centroid down below the split size?

vector/src/lire/rebalancer.rs
```diff
@@ -1,86 +1,1609 @@
...
+
+        // 1. Find nearest `split_search_neighbourhood` centroids of the split region
+        let neighbours = self
+            .centroid_graph
+            .search(c_vector, self.opts.split_search_neighbourhood + 2);
+
+        // 2. Get fresh view for reading posting lists
+        let reader = ViewReader::new(self.coordinator_handle.view());
+
+        // 3. For each neighbor (excluding c0, c1), check if any of their vectors should move
```
agavra, line 645: do we need to exclude the original `c` or is it gone by now?

vector/src/lire/rebalancer.rs
```diff
@@ -1,86 +1,1609 @@
...
+            .get_posting_list(c_from, self.opts.dimensions)
+            .await
+            .map_err(|e| e.to_string())?;
+
+        // Safety check: if c has < 1 vector (e.g. drained by concurrent ops), abort.
+        if posting_list.is_empty() {
+            warn!(
+                c_from,
+                num_vectors = posting_list.len(),
+                "cannot merge centroid with fewer than 2 vectors"
```
agavra:
```suggestion
                "cannot merge empty centroid"
```

vector/src/lire/rebalancer.rs
```diff
@@ -1,86 +1,1609 @@
...
+            .write(VectorDbWrite::Rebalance(sweep_cmd))
+            .await
+            .map_err(|e| e.to_string())?;
+        write_handle
+            .wait(Durability::Applied)
+            .await
+            .map_err(|e| e.to_string())?;
+
+        // =========================================================================
+        // Phase 3: REASSIGN
```
agavra, line 785: now that I'm thiking about it, why do we separate the the merge/reassign steps? i.e. why don't we just find the nearest new centroid for each when removing c_from instead of initially assigning them all to c_to?

vector/src/lire/rebalancer.rs
```diff
@@ -1,86 +1,1609 @@
...
+        // =========================================================================
+
+        // Re-read c_to's posting list (all vectors from c_from are now in c_to)
```
agavra: why are we doing this instead of collecting them from the first two steps (merge/sweep). we should have them already right?

vector/src/lire/commands.rs
```diff
@@ -15,27 +21,79 @@ pub(crate) enum RebalanceCommand {
     Merge(MergeCommand),
     MergeSweep(MergeSweepCommand),
     MergeReassign(MergeReassignCommand),
+    FinishSplit(FinishSplitCommand),
```
agavra: nit: `SplitFinish` and `MergeFinish` 😆 because that'll sort them nicely

vector/src/lire/rebalancer.rs
```diff
@@ -1,86 +1,1609 @@
...
+        );
+    }
+
+    pub(crate) fn total_ops_pending_and_running(&self) -> usize {
+        self.tasks.num_tasks() + self.pending_rebalance.len()
+    }
+
+    pub(crate) fn maybe_schedule_split(&mut self, candidate_centroid_id: u64) {
+        if self.tasks.participants().contains(&candidate_centroid_id) {
+            trace!("centroid is already actively participating in rebalance. skip");
```
agavra: nit for all of these logs we should add the centorid id

vector/src/lire/commands.rs
```diff
@@ -117,48 +282,256 @@ impl VectorDbWriteDelta {
...
+    ) -> Result<Arc<dyn Any + Send + Sync + 'static>, String> {
+        let mut view = self.view.write().expect("lock poisoned");
+
+        for reassignment in reassignments {
+            let source_id = self
+                .ctx
+                .centroid_graph
+                .get_centroid_vector(reassignment.source_centroid_id);
+            if source_id.is_none() {
+                // vector was from a centroid involved in a racing split/merge. skip
```
agavra, line 428: I think having racing split/merges answers one of my other questions but it makes me wonder if we should support it or not. The writer is single-threaded anyway, so I feel like we should consider queuing up `Split` and `Merge` commands and only picking the next one off the list once we see `FinishSplit` or `FinishMerge` so that racing rebalances aren't a problem

vector/src/lire/rebalancer.rs
```diff
@@ -1,86 +1,1609 @@
...
+
+    pub(crate) fn register_participants(&mut self, id: u64, participants: &[u64]) {
+        self.tasks.register_participants(id, participants);
+    }
+
+    pub(crate) fn finish_split(&mut self, task_id: u64, c0: Option<u64>, c1: Option<u64>) {
+        self.finish_task(task_id, vec![c0, c1].into_iter().flatten().collect());
+    }
+
+    pub(crate) fn finish_merge(&mut self, task_id: u64, c_to: u64) {
```
agavra, line 179: (putting this comment at a random location) I remember discussing in person "blocking" a flush until all the in-memory tracked Split/Merges are complete. Where does that happen?

vector/src/lire/rebalancer.rs
rodesai: It shouldn't really happen in practice. Its not possible to schedule multiple splits for the same centroid concurrently. The "queue" is really a set of centroid ids that need maintenance, and the rebalancer makes sure not to schedule work for a centroid that's already being split/merged. There is a very unlikely race here though where a split is scheduled for centroid C at the same time as some other centroid C' that's nearby C. As part of the split of C' you could theoretically move a bunch of vectors out of C as part of its neighborhood reassignment step.

vector/src/lire/rebalancer.rs
rodesai: It's gone by now. The original split removes it from the graph.

vector/src/lire/rebalancer.rs
rodesai: Yeah would save a read.

vector/src/lire/commands.rs
rodesai: I was thinking this originally, but we need it to drive enough parallelism to ingest at a reasonable rate. When loading up a db there are _a lot_ of splits happening.

vector/src/lire/rebalancer.rs
rodesai: It's not implemented in this PR. If we don't do it we'll just leave the index in a somewhat suboptimal state. Leaving fixing that for a follow-up.

vector/src/lire/rebalancer.rs
rodesai: We need to in one step delete the centroid so no new vectors can be placed into it _and_ move the vectors somewhere reasonable so they can be returned by search. I suppose we could compute their new assignments inline, but I wanted to have the old centroid be deleted before doing the search in the hnsw.

PR #223, opened 2026-02-18
add support for put/merge options to storage i/f + minimal log ttl

rodesai:
## Summary

This patch adds support for specifying per-record put/merge options when writing atomic batches to storage. Currently the options only have a single field - ttl.

Ttl can be set to Default, NoExpiry, or ExpireAfter. Default writes the row using the underlying storage system's default ttl. NoExpiry writes the row with no expiration. ExpireAfter writes the row with an explicit lifetime.

To allow specifying ttl on a per-row basis the write interfaces to storage are changed to encapsulate the written row in a PutRecordOp and MergeRecordOp for put/merge, respectively. These structs contain the row plus an options struct with per-row options.

For log, we write some metadata that must always be retained even if the rest of the data is written with expiry. Specifically, the sequence block and per-segment metadata need to be retained so that we don't lose the sequence allocation and the current segment/next segment id, respectively. After this patch we should be able to set default ttl for the slatedb storage backing log while still allowing new data to be written/read.

Note that this is mostly a workaround to allow us to use log without blowing up storage costs. We should still do a proper retention design for log that treats retention of segments atomically - either a segment exists or its expired. In the implementation here we leave the segment metadata around and just let slatedb clean up the log entries.

## Test Plan

Unit tests

## Checklist

- [x] Tests added/updated

### DIFF
--- a/log/src/server/handlers.rs
+++ b/log/src/server/handlers.rs
@@ -254,7 +254,7 @@ mod tests {
     use super::*;
     use crate::Config;
     use axum::http::StatusCode;
-    use common::StorageConfig;
+    use common::{MergeRecordOp, PutRecordOp, StorageConfig};
 
     fn test_config() -> Config {
         Config {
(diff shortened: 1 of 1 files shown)


agavra (approved):
LGTM

Issue #92, opened 2026-01-17
[tsdb] support the `/metadata` Prometheus API


rodesai:
## Summary

Support the metric metadata prometheus API. This is an experimental API, but is used by many tools like Grafana. The metadata endpoint returns metadata about the metrics that are currently being scraped from targets. We can probably start by just maintaining this catalogue in-memory for now and updating it as metrics are scraped.

## Motivation

Prometheus API support

## Additional Context

See https://prometheus.io/docs/prometheus/latest/querying/api/#querying-metric-metadata

ryandielhenn:
I’ll pick this up.

ryandielhenn:
Here's a PR: https://github.com/opendata-oss/opendata/pull/227

agavra:
Fixed in #227

PR #227, opened 2026-02-20
Support the `/metadata` Prometheus API

ryandielhenn:
## Summary

What does this PR do?

Adds support for the `/metadata` Prometheus API. 

## Related Issues

[Issue #92](https://github.com/opendata-oss/opendata/issues/92#event-22878466073)

## Test Plan

How was this tested?

Adds test coverage to test that the metadata handler handles the `metric`, `limit`, and `limit_per_metric` fields of the metadata request.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/tsdb_router.rs
+++ b/timeseries/src/promql/tsdb_router.rs
@@ -1997,7 +1997,7 @@ mod tests {
         .await
         .unwrap();
 
-        // when: query for a label that doesn't exist in any bucket
+        // when: query for all metric metadata
         let request = MetadataRequest {
             metric: None,
             limit: None,
@@ -2005,7 +2005,7 @@ mod tests {
         };
         let response = tsdb.metadata(request).await;
 
-        // then: should return empty result
+        // then: should return all metric metadata
         assert_eq!(response.status, "success");
         let data = response.data.unwrap();
         assert_eq!(data.len(), 2);
@@ -2030,7 +2030,7 @@ mod tests {
         .await
         .unwrap();
 
-        // when: query for a label that doesn't exist in any bucket
+        // when: query metadata filtered for cpu_usage
         let request = MetadataRequest {
             metric: Some("cpu_usage".to_string()),
             limit: None,
@@ -2064,7 +2064,7 @@ mod tests {
         .await
         .unwrap();
 
-        // when: query for a label that doesn't exist in any bucket
+        // when: query for metadata with limit=2
         let request = MetadataRequest {
             metric: None,
             limit: Some(2),
@@ -2117,7 +2117,7 @@ mod tests {
             .await
             .unwrap();
 
-        // when: query for a label that doesn't exist in any bucket
+        // when: query metadata using limit_per_metric=4
         let request = MetadataRequest {
             metric: None,
             limit: None,


timeseries/src/tsdb.rs
```diff
@@ -140,6 +146,27 @@ impl Tsdb {
...
+                        .metric_type
+                        .map(|t| t.as_str().to_string())
+                        .unwrap_or_default(),
+                    help: series.description.clone().unwrap_or_default(),
+                    unit: series.unit.clone().unwrap_or_default(),
+                };
+                let mut catalog = self.metadata_catalog.write().await;
+                let entries = catalog.entry(metric_name).or_default();
+                if !entries.contains(&entry) {
+                    entries.push(entry);
```
rodesai, line 166: I guess there's technically no bound on how large this can grow. I think that's ok though since we're just caching metric names rather than timeseries. Can you file an issue to deal with this by purging entries that are older than some threshold?

ryandielhenn: https://github.com/opendata-oss/opendata/issues/228

ryandielhenn: Filed the issue and tested the endpoint against a live instance using the quickstart config. Metadata is being returned correctly with types and help text from scraped targets.

samsond:
I was originally expecting metadata lifecycle to be scraper-scoped (updated during scrape + iteration-based pruning at end of scrape, similar to Prometheus). With this change, metadata appears to be owned by TSDB and updated during ingestion instead. Is that the intended ownership model going forward?

ryandielhenn:
I put the metadata catalog in `Tsdb` because the `metadata` handler goes through `PromqlRouter` which is implemented for `Tsdb`. It was a natural place to store and access the catalog, but i hadn't considered scraper-scoped ownership with pruning. I think keeping the catalog in `Tsdb` for storage/reads but moving the update and pruning logic to the scraper would be the right approach. Happy to refactor if you agree. @samsond

samsond:
Thanks @ryandielhenn  that clarifies it. I mainly wanted to confirm the intended ownership model rather than suggest a refactor. The current approach makes sense given how the router is structured.

timeseries/src/promql/tsdb_router.rs
```diff
@@ -713,8 +714,35 @@ impl PromqlRouter for Tsdb {
...
 
-    async fn metadata(&self, _request: MetadataRequest) -> MetadataResponse {
-        todo!()
+    async fn metadata(&self, request: MetadataRequest) -> MetadataResponse {
+        let catalog = self.metadata_catalog.read().await;
+        let mut data: HashMap<String, Vec<MetricMetadata>> =
+            if let Some(ref metric) = request.metric {
+                // Filter for specific metric
+                catalog
+                    .get(metric)
```
agavra, line 723: from an API perspective I'd expect to be able to pass in the `limit_per_metric` here so that we can retrieve fewer metadata rows

timeseries/src/tsdb.rs
```diff
@@ -30,6 +32,9 @@ pub(crate) struct Tsdb {
     /// LRU cache (50 max) for read-only query buckets.
     /// Only populated for buckets NOT in ingest_cache.
     query_cache: Cache<TimeBucket, Arc<MiniTsdb>>,
+
+    // Metadata catalog
+    pub(crate) metadata_catalog: RwLock<HashMap<String, Vec<MetricMetadata>>>,
```
agavra, line 37: is this fully in memory? do we repopulate it on a restart?

agavra (commented):
I may have missed something in a previous, but I'm not sure this approach works. The `/metadata` endpoint should probably work across restarts. I also think it should leverage the inverted/forward indexes instead of maintaining an additional in-memory structure that is updated independently (that would also solve hte restart problem)

rodesai:
> I may have missed something in a previous, but I'm not sure this approach works. The /metadata endpoint should probably work across restarts.

I don't think we need to support this. The endpoint is meant to return metadata about the metrics that are currently being scraped by the server. On a restart it will be repopulated by the scraper. This is consistent with the current prometheus behavior (let me dig up where I saw this)

agavra:
Ah, I confirmed the behavior locally:
```
> curl localhost:9090/api/v1/metadata
{"status":"success","data":{"mock_api_calls":[{"type":"counter","unit":"","help":"API calls by endpoint and method"}],"mock_cpu_usage_percent":[{"type":"gauge","unit":"","help":"Simulated CPU usage"}],"mock_errors_total":[{"type":"counter","unit":"","help":"Total number of errors"}],"mock_http_requests_by_status":[{"type":"counter","unit":"","help":"HTTP requests by status code"}],"mock_memory_usage_bytes":[{"type":"gauge","unit":"","help":"Simulated memory usage"}],"mock_request_duration_seconds":[{"type":"histogram","unit":"","help":"Request duration histogram"}],"mock_requests_total":[{"type":"counter","unit":"","help":"Total number of requests received"}],"mock_uptime_seconds":[{"type":"gauge","unit":"","help":"Time since server started"}]}}%

docker restart 7d3395aa0afa

curl localhost:9090/api/v1/metadata
{"status":"success","data":{}}%

> curl "localhost:9090/api/v1/label/__name__/values"
{"status":"success","data":["go_gc_cycles_automatic_gc_cycles_total",...}
```

agavra:
@rodesai I guess this is good to merge? Were you waiting on anything?

rodesai:
@agavra nope!

Issue #84, opened 2026-01-17
[tsdb] support functions with scalar parameters


rodesai:
## Summary

Support passing scalar parameters to functions in PromQL expressions.

## Motivation

Full PromQL support

## Proposed Solution

Fill in the todo in `eval_single_argument` in `Evaluator`

ryandielhenn:
I'll take this.

Issue #185, opened 2026-02-12
Track PromQL test coverage from Prometheus test suite


samsond:
The promqltest framework is now in place with ignore/resume directives for marking unimplemented features. This is a tracking issue for expanding test coverage by importing tests from Prometheus's test suite.

## Current Coverage

- [x] at_modifier.test - Basic @ modifier support (done in #177 )
- [x] selectors.test - Label matching and selectors (done in #177 )
- [x] subquery.test - Subquery support (done in #237)
- [x] aggregators.test - Aggregation functions (sum, avg, min, max, etc.) (done in #280)
- [x] functions.test - Scalar and vector functions (done in #229)
- [x] name_label_dropping.test - Label manipulation (In review pr #310)

## Prometheus Test Suite

Source: https://github.com/prometheus/prometheus/tree/main/promql/promqltest/testdata
Destination: `timeseries/src/promql/promqltest/testdata/`
Framework docs: See `timeseries/src/promql/promqltest/README.md` for test format and `ignore/resume` usage

### Core Tests (Priority)

Create sub-issues for these:

- [ ] operators.test - Binary operators (+, -, , /, %, ^)
- [ ] literals.test - Literal values and constants
- [ ] range_queries.test - Range vector selectors
- [ ] duration_expression.test - Duration parsing
- [ ] staleness.test - Staleness handling

### Advanced Tests

- [ ] limit.test - Result limiting
- [ ] info.test - Info metrics
- [ ] type_and_unit.test - Type and unit metadata
- [ ] trig_functions.test - Trigonometric functions
- [ ] collision.test - Hash collision handling
- [ ] histograms.test - Histogram operations
- [ ] native_histograms.test - Native histogram format
- [ ] extended_vectors.test - Extended vector operations
- [ ] fill-modifier.test - Fill modifier

## Note on Test Coverage

If you're implementing a feature that needs tests from this list:
- Add the test file as part of your feature PR (don't create a separate test-only PR)
- Check off the item in this tracking issue
- Link your feature PR in a comment

## Workflow for Sub-Issues

1. Copy test file from Prometheus to testdata/
2. Run tests to identify failures
3. Add ignore directive before unimplemented sections
4. Add resume when ready to continue testing
5. Track coverage in PR: X passing / Y total tests

## Example Sub-Issue Template

```text
**Title:** Add operators.test from Prometheus

**Description:**
Import `operators.test` from Prometheus test suite and mark unimplemented operators with `ignore`.

[...]

PR #229, opened 2026-02-23
add support for passing scalars to functions in promql expressions

ryandielhenn:
## Summary

Adds support for passing scalars to functions in promql expressions.

## Related Issues
#84 

## Test Plan

Build an expression with a scalar, evaluate the expression via `eval_single_argument`, and verify that the scalar is wrapped in a single element `Vec<EvalSample>`.

## Checklist

- [X] Tests added/updated

### DIFF
--- a/timeseries/src/promql/promqltest/testdata/functions.test
+++ b/timeseries/src/promql/promqltest/testdata/functions.test
@@ -1,5 +1,4 @@
 # Tests for PromQL scalar function arguments
-
 eval instant at 0s vector(1)
 	{} 1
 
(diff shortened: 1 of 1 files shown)


timeseries/src/promql/evaluator.rs
```diff
@@ -870,10 +870,14 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
-                "unsupported scalar result".to_string(),
-            )),
+            ExprResult::Scalar(s) => Ok(vec![EvalSample {
+                timestamp_ms: evaluation_ts
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .unwrap_or_default()
+                    .as_millis() as i64,
+                value: s,
+                labels: HashMap::new(),
+            }]),
```
cadonna: It is not yet clear to me if this is the right way to represent a scalar.

In Prometheus, they specify a separate type for a scalar: https://github.com/prometheus/prometheus/blob/70bc06718dbc4b0e7f588e9c6e8363c6c41da85f/promql/value.go#L52

I would follow the Prometheus pattern and specify a separate struct for a scalar. This might also simplify type checking since function might get scalars as arguments and return scalars.

I am also not sure if this is the correct location for the change. Method `eval_single_argument()` is not called anywhere, whereas method `evaluate_call()` seems to do the same thing (and also has a "not yet supported" message for scalars) and it is called in `evaluate_expr()`. It looks like `eval_single_argument()` is kind of a unused leftover.

cadonna: If my suspicion is correct and the right location is `evaluate_expr()`, then the scalar will anyway be  represented as `ExprResult::Scalar`.

ryandielhenn:
@cadonna Thanks for clearing this up. The issue pointed to `eval_single_argument`, so that's where I started, but you're right that it's unused. I've updated the PR to remove it and support Scalar in `evaluate_call` instead.

timeseries/src/promql/evaluator.rs
```diff
@@ -836,48 +836,7 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
-        {
-            // TODO: support functions with scalar args
-            ExprResult::Scalar(_) => Err(EvaluationError::InternalError(
-                "unsupported scalar result".to_string(),
-            )),
-            ExprResult::InstantVector(v) => Ok(v),
-            ExprResult::RangeVector(_) => Err(EvaluationError::InternalError(
-                "range vector handling not yet implemented in eval_single_argument".to_string(),
-            )),
+            ExprResult::Scalar(s) => Ok(ExprResult::Scalar(s)),
```
cadonna: Where do you apply the function to the scalar?

ryandielhenn: Thanks for your patience! I've added a PromQLScalarFunction trait for functions that operate on scalars, implemented VectorFunction using it, and registered it in the FunctionRegistry. The evaluate_call scalar arm now looks up functions from the registry instead of passing through. Updated the test accordingly.

timeseries/src/promql/evaluator.rs
```diff
@@ -2381,6 +2352,46 @@ mod tests {
         }
     }
 
+    #[tokio::test]
```
agavra, line 2842: let's also add tests to `src/promql/promqltest/testdata` now that we have an integration testing framework. also see #185 we should copy over any relevant tests from the prometheus test stuie

timeseries/src/promql/functions.rs
```diff
@@ -571,6 +586,19 @@ impl RangeFunction for StdvarOverTimeFunction {
     }
 }
 
+/// Vector function: converts a scalar value to a single-element instant vector with no labels.
+struct VectorFunction;
+
+impl PromQLScalarFunction for VectorFunction {
```
agavra: I think it might be more interesting to support something like `clamp(v instant-vector, min scalar, max scalar)` https://prometheus.io/docs/prometheus/latest/querying/functions/#clamp

this takes in both a vector and two scalars. this will verify that the approach works in general

timeseries/src/promql/evaluator.rs
```diff
@@ -2381,6 +2352,46 @@ mod tests {
...
+    async fn test_evaluate_call_scalar_argument() {
+        let (reader, end_time) = setup_mock_reader(vec![]);
+        let mut evaluator = Evaluator::new(&reader);
+
+        let call = Call {
+            func: Function::new("vector", vec![ValueType::Scalar], false, ValueType::Scalar),
+            args: FunctionArgs::new_args(Expr::NumberLiteral(NumberLiteral { val: 42.0 })),
```
agavra, line 2849: can we also add a test where the argument to the function is the result of a function? e.g. `vector(1 - 1)` is a valid expression in p8s

agavra (commented):
Thanks @ryandielhenn - it took me a second to understand what was going on in this PR, but it makes sense to me. I'm still a bit confused about how it looks like in the long term so I added some comments to that effect. I think more end-to-end testing and adding a more interesting scalar function (e.g. `clamp`) would help see this PR through

timeseries/src/promql/functions.rs
cadonna: I proposed `vector` because at the moment we restrict functions to have just one argument (https://github.com/opendata-oss/opendata/blob/582c752bca3da68cf39efc5e1f3f24a28a5d599c/timeseries/src/promql/evaluator.rs#L788).
Also the traits for the functions assume one single argument.
Changing all of that seemed to me to be outside the scope of this specific issue.

If you both agree, I can create a new issue that proposes to make the number of argument for functions arbitrary. Then we can get the scalar issue solved and after that we can tackle the argument issue.

WDYT?

agavra: Got it, with those additional limitations I guess that makes sense. We can keep this scoped to the `vector` function (is that the only single-argument scalar function? if that's the case having a registry for it seems like overkill). it seems like that is the case based on a quick scan / AI chat

ryandielhenn: `vector` is the only single-argument scalar function I found. I agree that supporting multiple arguments is a larger problem and that a separate issue could be created for it. 

I also agree that adding a new trait to the registry for one function seems a bit overkill. I'll try combining `PromQLFunction` & `PromQLScalarFunction` by using an enum for the arguments. E.G. 

```rust
pub(crate) enum PromQLArg {
    InstantVector(Vec<EvalSample>),
    Scalar(f64),
}

pub(crate) trait PromQLFunction {
    fn apply(
        &self,
        arg: PromQLArg,
        eval_timestamp_ms: i64,
    ) -> EvalResult<Vec<EvalSample>>;
}
```

`PromQLArg` could be extended to `Vec<PromQLArg>` in the future to support multiple arguments of type InstantVector/Scalar.

I would be interested to know if another approach is preferable here.

timeseries/src/promql/evaluator.rs
ryandielhenn: Will look into this tomorrow.

timeseries/src/promql/functions.rs
cadonna: Here the promised issue: https://github.com/opendata-oss/opendata/issues/250

cadonna: IMO, we can either hard-code the vector function or use a registry for now. The specification and retrieval of functions need to change anyways with issue https://github.com/opendata-oss/opendata/issues/250.

In other words, let's get this PR merged and then tackle https://github.com/opendata-oss/opendata/issues/250.

agavra:
@ryandielhenn let me know when this is ready for another pass and i'll take a look

timeseries/src/promql/evaluator.rs
ryandielhenn: @agavra I added the Prometheus `functions.test` suite into `testdata/`. Most tests are wrapped in `ignore` as they depend on unimplemented functions (e.g. `time()`, `year()`, `resets()`). These can be enabled as features are added. Coverage: 7/383 tests passing.

Also added a unit test in `evaluator.rs` for evaluating scalar expressions as function arguments (e.g. `vector(1 - 1)`).

Let me know if you'd like me to go through the file more granularly to identify additional passing tests.

agavra: Just too a look, thank you! This is really helpful, we can unignore those over time. This is excellent.

PR #233, opened 2026-02-24
[log] simplified writer path

hachikuji:
The write coordinator was adding a lot of complexity in the Log implementation. Delta accumulation, which is helpful for timeseries is not really leveraged in Log. To get back read-your-writes, we needed to implement additional merge logic, some of which I did submitted and closed [previously](https://github.com/opendata-oss/opendata/pull/206). 

In fact, slatedb's memtable is already a reasonable approach for representing log data. In this patch, I've implemented something which is based on the coordinator concepts, such as epoch management and view propagation, but much more tailored to Log. This restores read-your-writes, which just leaves "read durable" as the main semantic gap.

### DIFF
--- a/log/src/log.rs
+++ b/log/src/log.rs
@@ -289,7 +289,7 @@ impl LogDb {
         let target = self.handle.flushed_epoch();
         self.epoch_watcher
             .clone()
-            .wait(target, Durability::Flushed)
+            .wait(target, Durability::Written)
             .await
             .map_err(|_| Error::Internal("writer shut down".into()))?;
         Ok(())
@@ -497,7 +497,7 @@ fn spawn_flushed_subscriber(
                 last_segments = Some(Arc::clone(&view.segments));
             }
 
-            watermarks.update_flushed(view.epoch);
+            watermarks.update_written(view.epoch);
         }
     });
     (watcher, task)
--- a/log/src/writer.rs
+++ b/log/src/writer.rs
@@ -10,7 +10,7 @@
 //! The log writer skips [`Durability::Applied`] because there is no in-memory
 //! delta stage — writes go straight to SlateDB. The watermarks advance as:
 //!
-//! - **Flushed** — after each append or seal, once the data is in SlateDB's
+//! - **Written** — after each append or seal, once the data is in SlateDB's
 //!   memtable (visible to snapshot reads, but not yet on disk).
 //! - **Durable** — after `storage.flush()`, once SlateDB has synced to disk.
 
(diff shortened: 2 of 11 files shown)


log/src/log.rs
```diff
@@ -519,75 +471,36 @@ impl LogDbBuilder {
...
+    let task = tokio::spawn(async move {
+        let mut last_segments: Option<Arc<[LogSegment]>> = None;
+        while written_rx.changed().await.is_ok() {
+            let view = written_rx.borrow_and_update().clone();
+
             let mut rv = read_view.write().await;
-            rv.update_snapshot(view.snapshot.clone() as Arc<dyn common::StorageRead>);
-            if !flushed.val.new_segments.is_empty() {
-                rv.apply_new_segments(&flushed.val.new_segments);
+            rv.update_snapshot(view.snapshot as Arc<dyn common::StorageRead>);
```
agavra, line 489: this happens on every write now, correct?

log/src/writer.rs
```diff
@@ -0,0 +1,596 @@
...
+        // 3. Assign listing entries for new keys
+        let keys: Vec<Bytes> = write.records.iter().map(|r| r.key.clone()).collect();
+        self.listing_cache
+            .assign_new_keys(assignment.segment.id(), &keys, &mut records);
+
+        // 4. Build log entry records
+        self.add_entries(&assignment.segment, base_seq, &write.records, &mut records);
+
+        // 5. Write to storage and broadcast
+        self.write_and_broadcast(records, &assignment).await?;
```
agavra, line 190: this is the other main difference, we're not "flushing" and broadcasting on every write?

log/src/writer.rs
```diff
@@ -0,0 +1,596 @@
...
+        }
+
+        let snapshot = self.storage.snapshot().await.map_err(|e| e.to_string())?;
+        self.epoch += 1;
+        self.written_tx.send_replace(WrittenView {
+            epoch: self.epoch,
+            snapshot,
+            segments: Arc::clone(&self.segments_snapshot),
+        });
+        self.watermarks.update_flushed(self.epoch);
```
agavra: I wonder if reusing the watermarks makes sense since we're using "flushed" to mean two different things. alternatively we just rename it something like

1. Written
2. Readable
3. Durable

hachikuji: Yeah, I agree the terminology is confusing applied here. How about this?
1. Applied (to in-memory data structures)
2. Written (to SlateDb)
3. Durable (in s3)

For log, we would just skip over Applied.

agavra: LGTM

log/src/log.rs
hachikuji: Yes, that's right. I had a previous patch which pulled the changes to the reader lazily, but it was a little messier. Still something we can consider if we notice the locking in profiles.

PR #234, opened 2026-02-24
[timeseries] fix vector-vector binary ops with hash-based label matching

### LINKED ISSUE #231: PromQL: vector-vector binary operations always fail with cardinality/label matching errors
##  Description:

All vector-vector binary operations fail, even trivially correct ones like metric - metric. Vector-scalar operations  work fine.

## Steps to Reproduce

` curl -s --data-urlencode 'query=up - up' 'http://localhost:9090/api/v1/query'`

Expected: Every series subtracted from itself, all values = 0.
Actual:
```json
{"status":"error","data":null,"error":"PromQL internal error: Vector-vector operations require matching labels","errorType":"execution"}
```

Aggregated version also returns the same error: `max by (service_name) (up) - max by (service_name) (up)`

Vector-scalar operations work correctly: `up + 1`  succeeds returning 12 results

### Side-by-side: Prometheus vs OpenData Timeseries

Query: up - up (instant)

Prometheus:
```json
{"status":"success","data":{"resultType":"vector","result":[{"metric":{...},"value":[1740000000,"0"]}, ...]}}
```
Returns all series with value 0.

OpenData Timeseries:
```json
{"status":"error","data":null,"error":"PromQL internal error: Vector-vector operations require matching
  labels","errorType":"execution"}
```

## Additional Context

* Fails for both instant (/api/v1/query) and range (/api/v1/query_range) queries
* Fails regardless of whether offset is used
* The error message varies: "require matching labels" for same-metric ops, "require one-to-one matching cardinality" when offset is involved
* "data": null in the error response also breaks Grafana parsing (see #232)

apurvam:
## Summary

Replace broken positional zip() matching with hash-based matching:
- Drop `__name__` before comparison per Prometheus semantics
- Support on()/ignoring() label modifiers
- Silently drop unmatched series instead of erroring
- Enforce one-to-one cardinality with duplicate-key checks
- Handle bool modifier: retain false results, drop `__name__`
- Drop `__name__` for bool comparisons in vector-scalar paths

## Related Issues

Fixes #231 

## Test Plan

Added extesive promql compatibility tests. 

## Checklist

- [x] Tests added/updated


timeseries/src/promql/evaluator.rs
```diff
@@ -963,51 +972,71 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
                                 continue;
                             }
-                            sample.value = value;
-                            result.push(sample);
+                            let mut result_labels = left_sample.labels;
+                            // Drop __name__ per Prometheus semantics:
+                            // - Arithmetic ops: always drop
+                            // - Comparison ops: drop if on() (Include) or bool modifier;
+                            //   ignoring() alone does NOT cause __name__ drop
+                            let drop_name = if is_comparison {
```
agavra: can probably do something like 
```rs
let drop_name = !is_comparison 
 || matches!(matching, Some(LableModifier::Include(_)) 
 || return_bool
```

timeseries/src/promql/evaluator.rs
```diff
@@ -1660,6 +1722,175 @@ mod tests {
...
+        "cpu_usage + memory_bytes",
+        vec![
+            ("cpu_usage", vec![("env", "prod"), ("instance", "i1")], 0, 50.0),
+            ("cpu_usage", vec![("env", "prod"), ("instance", "i2")], 1, 60.0),
+            ("memory_bytes", vec![("env", "prod")], 2, 100.0),
+        ],
+        vec![] // No matching label sets (different labels), all dropped
+    )]
+    #[case(
+        "binary_vector_vector_comparison",
```
agavra, line 1842: can we add a test case for matched arithmetic with different metric names? (e.g. only label that doesn't match is `__name__`)

timeseries/src/promql/evaluator.rs
```diff
@@ -926,11 +929,14 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
                             Ok(value) => {
-                                // For comparison operations, filter out false results (0.0)
-                                if is_comparison && value == 0.0 {
+                                // For comparison ops without bool, filter out false results
+                                if is_comparison && !return_bool && value == 0.0 {
                                     None
                                 } else {
                                     sample.value = value;
+                                    if is_comparison && return_bool {
+                                        sample.labels.remove(METRIC_NAME);
```
agavra: 😵‍💫 all these scattered operations that remove `__name__` from labels feels very brittle. It seems to me like there should be a more principled approach to it.

I asked Claude to examine prometheus' go code and it came up with this:

----

Great find. The key insight from Prometheus is:

**The rule is much simpler than what the current code implements.** Prometheus uses:

```go
drop_name = changesMetricSchema(op) || returnBool
```

Where `changesMetricSchema` is true for arithmetic ops (`+`, `-`, `*`, `/`, `%`, `^`, `atan2`). This same rule applies uniformly to both vector-scalar and vector-vector paths. There's no special-casing for `on()` vs `ignoring()`.

This means the current code has a bug on line 1028 — it special-cases `on()` to drop `__name__` for comparisons, but Prometheus doesn't do that. The only comparison case that drops `__name__` is `bool`.

**Prometheus also has a "delayed name removal" approach** worth noting: instead of calling `remove(__name__)` inline, they set a `DropName: true` flag on each sample, and then a single `cleanupMetricLabels()` pass at the top level strips it at the very end. This is exactly the "single place" pattern you're looking for.

So the principled approach, matching Prometheus, would be:

1. Add a `drop_name: bool` field to `EvalSample`
2. In each binary op path, just set `sample.drop_name = changesMetricSchema(op) || return_bool` — no `remove()` calls anywhere
3. At the top of the evaluation (where the final result is returned), do one pass to strip `__name__` from any sample with `drop_name == true`

The condition simplifies to just:

```rust
fn changes_metric_schema(op: &TokenType) -> bool {
    matches!(op.id(), T_ADD | T_SUB | T_MUL | T_DIV | T_MOD | T_POW)
}

// In each binary op arm:
sample.drop_name = changes_metric_schema(op) || return_bool;
```

~No `on()` / `ignoring()` check needed — that was incorrect in the current implementation.~ <-- this is wrong

agavra: not sure if this is helpful, but here are supposed references (i'm about to go through them and check if they're real):


Here are the key reference points in `prometheus/prometheus` at commit `5d3f9ee`:

**The core rule** — `changesMetricSchema`:
https://github.com/prometheus/prometheus/blob/5d3f9ee39b1f1414e6fb06a534d1ae0b21745d2a/promql/engine.go#L4474-L4481

**Delayed cleanup** — single-pass `cleanupMetricLabels` at the top level:
https://github.com/prometheus/prometheus/blob/5d3f9ee39b1f1414e6fb06a534d1ae0b21745d2a/promql/engine.go#L4291-L4313

**Vector-scalar** — sets `DropName` when `changesMetricSchema(op) || returnBool`:
https://github.com/prometheus/prometheus/blob/5d3f9ee39b1f1414e6fb06a534d1ae0b21745d2a/promql/engine.go#L3142-L3147

**Vector-vector** — two mechanisms:
- `resultMetric` drops name eagerly for arithmetic via `changesMetricSchema`: https://github.com/prometheus/prometheus/blob/5d3f9ee39b1f1414e6fb06a534d1ae0b21745d2a/promql/engine.go#L3081-L3084
- `DropName: returnBool` set on the sample for bool comparisons: https://github.com/prometheus/prometheus/blob/5d3f9ee39b1f1414e6fb06a534d1ae0b21745d2a/promql/engine.go#L3014-L3019

One nuance worth noting: even in Prometheus, the vector-vector path is slightly split — `resultMetric` handles arithmetic eagerly while `DropName` handles `bool` comparisons via deferred cleanup. But the underlying rule is the same everywhere: `changesMetricSchema(op) || returnBool`.

agavra: re: on/ignoring they say:

>   So the on() / ignoring() difference for __name__ in comparisons is real behavior, but in Prometheus it's a natural side effect of on() filtering result labels down to only the listed labels — not a special __name__

  For a comparison like a > on(env) b:
  - changesMetricSchema(>) is false → __name__ is NOT explicitly dropped
  - But matching.On is true → lb.Keep(["env"]) — only env is kept, so __name__ is implicitly dropped because it's not in the on() list

  For a > ignoring(instance) b:
  - changesMetricSchema(>) is false → __name__ not dropped
  - matching.On is false → lb.Del(["instance"]) — only instance is deleted, __name__ survives

apurvam: Great call to have Claude look at the existing prometheus code. I agree that the current code is a bit of a mess which is really hard to understand. 

I'll have it take a second pass with your suggestions.

timeseries/src/promql/evaluator.rs
apurvam: Added some cases: https://github.com/opendata-oss/opendata/pull/234/changes/2fd1d0b35e3577d5df0677717cd3ca13a66bae22#diff-5d841901db73ef0d8f844e4b5926153e934423526fc5883f71495ae3ac506f13R1907

timeseries/src/promql/evaluator.rs
apurvam: Updated the code to match the suggested structure. It's much cleaner IMO.

PR #236, opened 2026-02-24
read neighbour postings in parallel during split

rodesai:
## Summary

When splitting a centroid the rebalancer searches its neighbours for vectors that need to be reassigned to preserve NPA. This patch does these reads concurrently to help speed up splits by doing the i/o concurrently.

## Test Plan

- all unit tests pass
- vector recall is preserved (e2e recall test passes)

## Checklist

- [x] Tests added/updated

### DIFF
--- a/vector/src/lire/rebalancer.rs
+++ b/vector/src/lire/rebalancer.rs
@@ -708,8 +708,7 @@ impl IndexRebalancerTask {
 
         // 4. Process each neighbour's postings to find reassignments
         for (neighbour_id, posting_result) in neighbour_ids.iter().zip(posting_results) {
-            let neighbour_postings: PostingList =
-                posting_result.map_err(|e| e.to_string())?;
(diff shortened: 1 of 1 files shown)


vector/src/lire/rebalancer.rs
```diff
@@ -693,18 +693,23 @@ impl IndexRebalancerTask {
...
+            .filter(|&id| id != c0_id && id != c1_id)
+            .collect();
 
-            let neighbour_postings_value = reader
-                .get_posting_list(neighbour_id, self.opts.dimensions)
-                .await
-                .map_err(|e| e.to_string())?;
+        let posting_futures: Vec<_> = neighbour_ids
+            .iter()
+            .map(|&id| reader.get_posting_list(id, self.opts.dimensions))
```
agavra, line 705: I feel like we should support `reader.get_posting_list_batch([ids], dims)` - there may be some smarts we can do in the storage layer to reduce lookups when looking up multiple IDs. Can always be done as a followup.

PR #237, opened 2026-02-25
feat(timeseries): add subquery support to PromQL evaluator

### LINKED ISSUE #81: support promql subqueries from tsdb
## Summary

Add support for subqueries to the timeseries promql evaluator. This will allow running queries with expressions of the form `<instant_query> '[' <range> ':' [ <resolution> ] ']' [ offset <duration> ]`. Subqueries are explained in detail in the prometheus docs and in the subquery proposal:
- https://prometheus.io/blog/2019/01/28/subquery-support/
- https://prometheus.io/docs/prometheus/latest/querying/basics/#subquery

## Motivation

Needed to support the full PromQL spec.

## Proposed Solution

The evaluator already has a branch for handling subqueries, it's just stubbed with `!todo()`. We need to fill that branch in by executing the subquery the way that `query_range` executes - e.g. by iterating over all the evaluation times and calling back into the evaluator with the specified instant query.

samsond:
Changes:
- Add Timestamp type for clean i64 millisecond arithmetic
- Implement evaluate_subquery with step alignment and time modifiers
- Refactor evaluator to use Timestamp internally (SystemTime only at entry)
- Support @ modifier and offset on subqueries
- Handle negative timestamps for subqueries before UNIX_EPOCH

Tests:
- Add subquery.test with 30+ test cases
- Update at_modifier.test to enable subquery tests
- Ignore 2 edge cases: counter-to-zero extrapolation, native histograms

## Related Issues

Fixes #81 

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -671,9 +671,8 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
 
         // Evaluate the inner expression at each step within the subquery range
         let mut series_map: HashMap<Vec<Label>, Vec<Sample>> = HashMap::new();
-        let mut current_time_ms = aligned_start_ms;
 
-        while current_time_ms <= subquery_end_ms {
+        for current_time_ms in (aligned_start_ms..=subquery_end_ms).step_by(step_ms as usize) {
             let current_time = Timestamp::from_millis(current_time_ms);
 
             let result = self
@@ -712,8 +711,6 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
                     value: sample.value,
                 });
             }
-
-            current_time_ms += step_ms;
         }
 
         let mut range_vector = Vec::new();
--- a/timeseries/src/promql/promqltest/testdata/subquery.test
+++ b/timeseries/src/promql/promqltest/testdata/subquery.test
@@ -124,7 +124,7 @@ load 7s
 eval instant at 80s rate(metric_total[1m])
   {} 2.517857143
 
-# TODO: Counter extrapolation toward zero not yet fully implemented.
+# TODO(OpenData): Counter extrapolation toward zero not yet fully implemented.
 # Expected: 2.360655737704918 (extrapolated toward zero/counter origin: (144 - 0) / 61)
 # Got: 2.3666666666666667 (extrapolated to range start: (144 - 2) / 60)
 # Prometheus extrapolates counters toward zero (counter origin) for better accuracy,
@@ -153,7 +153,7 @@ eval instant at 20s min_over_time(metric_total[10s])
 eval instant at 20s min_over_time(metric_total[15s:10s])
   {} 1
 
-# TODO: Nested subquery with rate() shows ~5x difference due to extrapolation issues.
+# TODO(OpenData): Nested subquery with rate() shows ~5x difference due to extrapolation issues.
 # Expected: 0.12119047619047618, Got: 0.5892857142857143
 # This is caused by incomplete counter boundary extrapolation in rate(), which becomes
 # amplified when nested inside a subquery. The Fibonacci sequence data (exponentially


samsond:
### apply_time_modifiers:

Originally, apply_time_modifiers was implemented specifically for VectorSelector, since that was the only expression type using @ and offset.

While adding subquery support, it became clear that time modifiers are shared semantics across multiple expression types, not just VectorSelector.

I refactored the function to accept only the minimal required fields (@ and offset) instead of coupling it to a specific expression type. This keeps the logic reusable, reduces coupling, and makes it easier to extend to other expressions that support time modifiers.

### dsl.rs:
Previously, when the parser encountered ignore, it would still attempt to parse commands inside the ignored section (e.g., load) and fail, even though that block should be skipped entirely.

The issue was that command parsing was happening before checking the ignore/resume state.

The fix ensures that the ignore/resume logic is evaluated first. Everything between ignore and resume is now treated as raw text and skipped, regardless of what commands appear inside. This aligns the behavior with the intended DSL semantics.

### Subquery Evaluation (Review Focus)

Most of the change affects subquery evaluation and how time modifiers interact with it.

Nested subqueries were tested and are passing. Time modifier semantics remain unchanged, but please review:

- Evaluation timestamp propagation
- Step alignment (`div_euclid`)
- Interaction between `@`, `offset`, and subquery range

samsond:
Note to self: update the tracking issues for subquery tests after this PR is approved — https://github.com/opendata-oss/opendata/issues/185

timeseries/src/promql/promqltest/testdata/subquery.test
```diff
@@ -0,0 +1,198 @@
...
+
+# Only one value between 10s and 20s, 2@14
+eval instant at 20s min_over_time(metric_total[10s])
+  {} 2
+
+# min(2@20)
+eval instant at 20s min_over_time(metric_total[15s:10s])
+  {} 1
+
+# TODO: Nested subquery with rate() shows ~5x difference due to extrapolation issues.
```
agavra: is this our TODO or a prometheus TODO?

timeseries/src/promql/evaluator.rs
```diff
@@ -596,23 +600,136 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
+                labels_key.sort();
+
+                let values = series_map.entry(labels_key).or_default();
+                values.push(Sample {
+                    timestamp_ms: current_time_ms,
+                    value: sample.value,
+                });
+            }
+
+            current_time_ms += step_ms;
```
agavra: nit: can this be done as a for loop? a bit easier to read

timeseries/src/promql/evaluator.rs
```diff
@@ -596,23 +600,136 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
+
+            for sample in samples {
+                let mut labels_key: Vec<Label> = sample
+                    .labels
+                    .iter()
+                    .map(|(k, v)| Label {
+                        name: k.clone(),
+                        value: v.clone(),
+                    })
+                    .collect();
```
agavra, line 705: this seems like it might be a really hot loop to be doing so much label cloning. probably worth a ticket to figure out how to reduce the cloning overhead here (especially given the labels are probably going to be the same for lots of samples). Maybe maintaining a first map from Unordered Labels -> Ordered Labels so we only create it once (and most of them will probably have the same order of labels).

timeseries/src/promql/promqltest/testdata/subquery.test
samsond: Good catch, I'll update the comment to clarify that these gaps in our current implementation

timeseries/src/promql/evaluator.rs
samsond: cool, I will update this as well.

timeseries/src/promql/evaluator.rs
samsond: Makes sense. Since labels are stable per series, we can maintain a small cache from unordered label sets to their canonical sorted representation and avoid repeated sorting in the subquery step loop. 

I've opened a follow-up ticket to track this optimization: https://github.com/opendata-oss/opendata/issues/242

agavra:
Changes looks good, do you want to hit the button now that you've got write access? 😆 🚀

PR #240, opened 2026-02-25
apply centroid pruning optimization from SPANN

rodesai:
## Summary

This patch applies the query-aware dynamic pruning optimization from spann. The optimization is enabled by setting the config query_pruning_factor. If this is set, then query will only consider centroids that are a similar distance to the query vector as the closest centroid, where similar is defined as being within that distance multiplied by the pruning factor

## Test Plan

unit tests added
ran the sift100k integration test

## Checklist

- [x] Tests added/updated

### DIFF
--- a/vector/src/db.rs
+++ b/vector/src/db.rs
@@ -685,8 +685,14 @@ impl VectorDb {
         }
 
         // 3. Dynamic pruning: skip posting lists whose centroids are far from query
+        let original_ncentroids = centroid_ids.len();
         let centroid_ids = self.prune_centroids(&centroid_ids, query);
-        debug!("after dynamic pruning: {} centroids", centroid_ids.len());
+        debug!(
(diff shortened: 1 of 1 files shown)


vector/src/db.rs
```diff
@@ -681,7 +684,11 @@ impl VectorDb {
             return Ok(Vec::new());
         }
 
-        // 3. Load posting lists and score candidates
+        // 3. Dynamic pruning: skip posting lists whose centroids are far from query
+        let centroid_ids = self.prune_centroids(&centroid_ids, query);
+        debug!("after dynamic pruning: {} centroids", centroid_ids.len());
```
agavra: nit: either remove this or add information that identifies the query and other context to make it more meaningful

rodesai: good call. will update

PR #244, opened 2026-02-25
Failing storage tests

hachikuji:
I was looking at test gaps and noticed that we do not test storage failures. I considered using failpoints, but it is a bit heavy for simple error propagation testing, so I introduced a `FailingStorage` wrapper which allows test cases to enable basic failures.


common/src/storage/in_memory.rs
```diff
@@ -371,6 +371,187 @@ impl Storage for InMemoryStorage {
...
+    }
+
+    /// Makes `put_with_options` return the given error on every subsequent call.
+    pub fn fail_put(&self, err: super::StorageError) {
+        self.fail_put
+            .store(Arc::new(Some(Failure::Persistent(err))));
+    }
+
+    /// Makes `put_with_options` return the given error on the next call only.
+    pub fn fail_put_once(&self, err: super::StorageError) {
```
rodesai, line 476: we should probably rename this and the above to `fail_put_with_options(_once)`

log/src/writer.rs
```diff
@@ -593,4 +607,121 @@ mod tests {
...
+        let (writer, mut handle, _) =
+            create_writer_with_storage(failing.clone(), LogWriterConfig::default()).await;
+        let _task = handle.spawn(writer);
+
+        // Enable put failure after writer is running
+        failing.fail_put(common::StorageError::Storage("test put error".into()));
+
+        let result = handle.try_append(make_write(&["key1"], 1000)).await;
+        assert!(
+            matches!(&result, Err(AppendError::InvalidRecord(msg)) if msg.contains("test put error")),
```
rodesai: why does this return `InvalidRecord`? Also should these tests assert that future appends also fail?

log/src/writer.rs
```diff
@@ -593,4 +607,121 @@ mod tests {
...
+        let (writer, mut handle, _) =
+            create_writer_with_storage(failing.clone(), LogWriterConfig::default()).await;
+        let _task = handle.spawn(writer);
+
+        // First append fails, then auto-clears
+        failing.fail_put_once(common::StorageError::Storage("test put error".into()));
+        let result = handle.try_append(make_write(&["key1"], 1000)).await;
+        assert!(result.is_err());
+
+        // Recover — writer task should still be alive
```
rodesai: this is not safe right? The puts for append `key1` could include writes that the second append depends on (e.g. seq allocation, segment allocation)

hachikuji: Yeah, I noticed that too. I was planning to fix it separately. I'll add a FIXME and an issue.

log/src/writer.rs
hachikuji: Yes, it is a bug. Will fix.

common/src/storage/in_memory.rs
hachikuji: I updated so that `fail_put` applies to both `put` and `put_with_options`.

log/src/writer.rs
hachikuji: Filed https://github.com/opendata-oss/opendata/issues/248.

rodesai (approved):
LGTM!

PR #245, opened 2026-02-25
[timeseries] First pass at timseries API formalization

hachikuji:
This is an attempt to clarify the timeseries APIs. I've divided the rough surface area into three RFCs (including the write RFC which already exists):

RFC-0002: Embedded write API (as today, but lifted service endpoints)
RFC-0003: Embedded read API (Promql as the first-class language for the embedded query API)
RFC-0004: Service API (specifies service endpoints including support for promql over http, otel, and remote-write)

At the end of the day, we would have `TimeSeriesDb` as the entry point for reads/writes. HTTP/OTEL services would use `TimeSeriesDb` under the covers. For OTEL data, I am proposing an `OtelSeriesBuilder` to adapt to the prometheus-like data model that the embedded write API uses. The existing service endpoints would remain the same.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -439,6 +439,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
+name = "cbindgen"
+version = "0.28.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "eadd868a2ce9ca38de7eeafdcec9c7065ef89b42b32f0839278d55f35c54d1ff"
+dependencies = [
+ "clap",
+ "heck 0.4.1",
+ "indexmap 2.12.1",
+ "log",
+ "proc-macro2",
+ "quote",
+ "serde",
+ "serde_json",
+ "syn 2.0.111",
+ "tempfile",
+ "toml 0.8.23",
+]
+
+[[package]]
 name = "cc"
 version = "1.2.49"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2467,7 +2486,7 @@ checksum = "d6790f58c7ff633d8771f42965289203411a5e5c68388703c06e14f24770b41e"
 
 [[package]]
 name = "opendata-common"
-version = "0.1.7"
+version = "0.1.8"
 dependencies = [
  "async-trait",
  "bytes",
@@ -2521,8 +2540,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "opendata-log-c"
+version = "0.1.0"
+dependencies = [
+ "bytes",
+ "cbindgen",
+ "opendata-common",
+ "opendata-log",
+ "tempfile",
+ "tokio",
+]
+
+[[package]]
 name = "opendata-timeseries"
-version = "0.1.6"
+version = "0.1.7"
 dependencies = [
  "async-trait",
  "axum",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
     "keyvalue",
     "log",
     "log/bench",
+    "log/c",
     "timeseries",
     "timeseries/bench",
     "vector",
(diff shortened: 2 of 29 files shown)


timeseries/rfcs/0003-read-api.md
```diff
@@ -0,0 +1,220 @@
...
+```rust
+impl TimeSeriesDb {
+    /// Find all series matching the given label matchers.
+    ///
+    /// Each matcher follows PromQL selector syntax (e.g., `{job="prometheus"}`).
+    /// At least one matcher is required.
+    pub async fn series(
+        &self,
+        matchers: &[&str],
+        start: Option<SystemTime>,
```
apurvam: I assume that if `start` and `end` is not specified, the range is all time? Would be good to clarify in the comments, here and in the other methods.

timeseries/rfcs/0004-service.md
```diff
@@ -0,0 +1,196 @@
...
+otel = ["dep:opentelemetry-proto", "dep:prost"]
+```
+
+| Feature | What it enables | Key dependencies |
+|---|---|---|
+| `http-server` | PromQL endpoints, server lifecycle | axum, tokio, tower |
+| `remote-write` | `POST /api/v1/write` (implies `http-server`) | prost, snap |
+| `otel` | `OtelSeriesBuilder`; `POST /v1/metrics` when `http-server` also enabled | opentelemetry-proto, prost |
+
+The `otel` feature is usable without `http-server` — the builder can be used standalone for programmatic conversion (e.g., replaying OTEL data from `LogDb`).
```
apurvam, line 71: Nice!

timeseries/rfcs/0004-service.md
```diff
@@ -0,0 +1,196 @@
...
+
+An earlier draft wrapped `TimeSeriesDb` in an `OtelTimeSeriesDb` that accepted OTEL requests and wrote decomposed series internally. This couples the conversion and storage steps. The builder approach keeps them explicit — `build()` produces series, then the caller writes them — and composes naturally with `TimeSeriesDb::write()`.
+
+### Separate server per protocol
+
+Each protocol could run its own server on a different port. This adds operational complexity with no clear benefit. A single server with feature-gated routes is simpler.
+
+## Open Questions
+
+1. **Unit suffix handling** — Should the OTEL `unit` field be appended to the metric name (e.g., `request_duration_seconds`) or stored only as metadata? Prometheus convention appends it.
```
apurvam: Since we are leaning into promQL for querying, we might as well lean into the prometheus convention here as well, IMO.

apurvam (approved):
Left a few minor comments, but this looks great!

timeseries/rfcs/0003-read-api.md
```diff
@@ -0,0 +1,220 @@
...
+- Keep the API independent of any transport layer
+
+## Non-Goals
+
+- **HTTP endpoints** — See RFC 0004
+- **PromQL language extensions** — Standard PromQL only
+- **Write API** — See RFC 0002
+- **Raw sample iteration** — Low-level access to stored samples may be added later
+
+## Design
```
agavra, line 46: Worth taking a look at `PromqlRouter` which I think is effectively this. It might be nice to make all the parameters structs so they're easy to JSON-ify and map to the Promql HTTP requests directly. I can go either way on it, the JSON-ifying can be done in the router that wraps the TimeSeriesDb

timeseries/rfcs/0004-service.md
```diff
@@ -0,0 +1,196 @@
...
+
+2. **Scope name as label** — Should `otel_scope_name` and `otel_scope_version` be included as labels by default, or only when explicitly configured?
```
agavra: I think we should match bheavior of https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/

timeseries/rfcs/0003-read-api.md
hachikuji: Yeah, it's a little more conventional to use `RangeBounds`. I'll update.

timeseries/rfcs/0003-read-api.md
hachikuji: The first sketch was basically PromqlRouter pulled into TimeSeriesDb., but I preferred more of a native rust feel to the embedded API. However, I think we can still find a nice convergence in the result types to avoid the need for translation when PromqlRouter wraps TimeSeriesDb. For example, `SeriesResponse` would use `Vec<Labels>` as opposed to `Vec<HashMap<String, String>>`. `PromqlRouter` should not be doing much more than wrapping query arguments and result types in principle.

PR #252, opened 2026-02-26
add RFC for stateless ingest

cadonna:
## Summary

The RFC proposes a design for stateless ingest for OpenData systems.

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/ingest/rfcs/0001-stateless-ingest.md
+++ b/ingest/rfcs/0001-stateless-ingest.md
@@ -357,6 +357,35 @@ This contention would also affect the queue producers.
 With a queue producer manifest and a queue consumer manifest, we can better decouple producer contention from
 consumer contention.
 
+
+### Using a counter for the batch names
+
+We could use a counter for the batch names to impose an order.
+During the flushing of batches, the ingestor:
+1. lists the prefix to which the batches are written, 
+2. finds the name with the largest number, 
+3. increases it, and 
+4. conditionally tries to write the batch named with the increased number.
+
+If the write fails because a batch with that name already exists, the ingestor increases the number again, 
+tries to write the batch again, and so forth.
+
+The collector:
+1. lists the prefix to which the batches are written,
+2. finds the batch with the lowest number,
+3. makes that batch available.
+
+Once the batch is acknowledged, the batch is removed from the prefix.
+
+This approach has the disadvantages that it cannot be easily used with multiple collectors, since a collector cannot
+know if the batch with the name with the lowest number has been already claimed by a different collector.
+
+It is also not clear if the conflict resolution would have a higher latency 
+since the conflict would be on the data batches and the conditional write is checked at the end of the write.
+That means, that maybe all the data is first send to object storage to then discover that there is a different
+object with the same name. This aspect would require some experiments.
+
+
 ## Open Questions
 
 - What format should we use for the data batches?


ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,343 @@
...
+queue.
+More specifically, the location is appended to the end of the list of pending locations in the producer queue manifest
+(`producer-q` in the diagram).
+
+Method `close()` flushes unflushed key-value pairs and terminates the ingestor.
+
+### Data Batch Format
+
+A data batch is the unit of data that an ingestor flushes to object storage.
+Each batch is a JSON array of key-value entries serialized in ingestion order:
```
apurvam: Would a binary format be more appropriate here, because performance would be critical right?

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,343 @@
...
+    "data/f47ac10b-58cc-4372-a567-0e02b2c3d479.batch": 1719400000000,
+    "data/9b2e1c4d-83f5-4a1e-b9d7-6f8e2a3c5b10.batch": 1719400005000
+  },
+  "done": [
+    "data/3d6f4a9e-c2b7-41d8-9e5f-7a1b3c8d2e6f.batch"
+  ]
+}
+```
+The `claimed` map associates each claimed location with a timestamp in milliseconds since the Unix epoch.
+This timestamp is set when the location is claimed and refreshed by `heartbeat()`.
```
apurvam: 'or refreshed' ?

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,343 @@
...
+With a very fast collector, the contention on the manifest would be quite high.
+This contention would also affect the queue producers.
+With a queue producer manifest and a queue consumer manifest, we can better decouple producer contention from
+consumer contention.
+
+## Open Questions
+
+- What format should we use for the data batches?
+- Should the clean-up of the data batches happen independently of the clean-up of the queue?
+- Should we also track the age of claimed/done batches to see if batches fall too much behind?
```
apurvam, line 393: I think by this you mean some notion of lag? I think 3 metrics are important: size of unclaimed batches, size of claimed but not done batches, size of done batches. If the first two are growing faster than the third, that's a problem.

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,343 @@
...
+
+With a very fast collector, the contention on the manifest would be quite high.
+This contention would also affect the queue producers.
+With a queue producer manifest and a queue consumer manifest, we can better decouple producer contention from
+consumer contention.
+
+## Open Questions
+
+- What format should we use for the data batches?
+- Should the clean-up of the data batches happen independently of the clean-up of the queue?
```
apurvam, line 392: IMO, we should keep things simple initially and do it all together. We can always split it out later.

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,343 @@
...
+### Using one single manifest for the queue
+
+With a very fast collector, the contention on the manifest would be quite high.
+This contention would also affect the queue producers.
+With a queue producer manifest and a queue consumer manifest, we can better decouple producer contention from
+consumer contention.
+
+## Open Questions
+
+- What format should we use for the data batches?
```
apurvam, line 391: What do you mean by format? I think a simple protobuf k/v List schema would be enough.

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,343 @@
...
+flushed objects to a queue with the internally used queue producer (`q-producer` in the diagram).
+The queue is implemented on object storage.
+Flushes are triggered after a given time interval elapsed or if a batch of the ingested data exceeds a given size.
+
+The API of the ingestor is the following:
+```rust
+impl Ingestor {
+  pub fn new(config: IngestorConfig, clock: Arc<dyn Clock>) -> Result<Self> { ... }
+
+  pub async fn ingest(&self, entries: Vec<KeyValueEntry>) -> Result<WriteWatcher> { ... }
```
agavra: curious why we specify `KeyValueEntry` instead of just `Bytes`, it's really up to the system to decide what this is (it could be a single Avro record for all we care)

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,343 @@
...
+}
+```
+The queue producer manifest takes the name specified in `manifest_path`.
+The config `flush_size_bytes` is a loose limit.
+The batch needs to exceed that size to trigger a flush to object storage.
+The config `max_unflushed_bytes` is also a loose limit.
+Each time the call to `ingest()` sees a size of unflushed key-value pairs in the ingestor that is larger
+than `max_unflushed_bytes`, the call blocks and flushes will be triggered until the size of the unflushed key-value
+pairs is less than `max_unflushed_bytes`.
+If this backpressure blocking the ingestion becomes an issue, new ingestors can be created to better distribute the
```
agavra, line 204: I do wonder how feasible this is if it creates more contention on the lock file.

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,343 @@
...
+load.
+
+A call to `ingest()` takes a vector of key-value pairs and returns a `WriteWatcher` with which the caller can await
+the completion of the flush to object storage of the vector of key-value pairs.
+
+The `WriteWatcher` has the following API:
```
agavra, line 210: nit we use `WriteHandle` for this in write coordinator, probably worth matching that?

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,343 @@
...
+```rust
+    pub fn result(&self) -> Option<Result<()>>
+
+    pub async fn await_durable(&mut self) -> Result<()>
```
agavra, line 214: I wonder if it's worth adding an API for "data durable" vs. "metadata durable (fully durable)". If you get "data durable" then we can return an ID for the batch of writes that they can retry the write without sending the same writes back. Can be an extension but an interesting failure handling situation

ingest/rfcs/0001-stateless-ingest.md
agavra: +1, I think a binary format would be nice here, it can also be streamed more easily I suspect (cc @rodesai mentioned the S3 streaming puts). 

Though we should figure out compression and encryption as well (might make streaming impossible, so we might want something like the block format in slate so we can compress + encrypt a block and then send it over the wire in a streaming put)

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,343 @@
...
+    pub key: Bytes,    // serialized as base64
+    pub value: Bytes,  // serialized as base64
+}
+```
+Both `key` and `value` are opaque byte blobs encoded as base64 in the JSON representation.
+Their semantics are defined by the database that consumes the data
+(e.g., the key could be a metric name for Timeseries or a stream name for Log).
+The ingest module does not interpret keys or values; it preserves them as-is.
+
+Each batch is stored under the configured `data_path_prefix` with a UUID v4 filename
```
agavra: Any reason not to use ULIDs? it's nice to get a time ordered id

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,343 @@
...
+A `CollectedBatch` contains the deserialized key-value entries and the location of the batch:
+```rust
+pub struct CollectedBatch {
+    pub entries: Vec<KeyValueEntry>,
+    location: String,
+}
+```
+
+By calling `next_batch()` the collector claims the next available location from the queue, reads the
+data batch from object storage, deserializes the key-value entries, and returns them as a `CollectedBatch`.
```
agavra: we should make fetching the data and deserializing the entries optional. Some systems (like log) might be able to do a metadata-only operation (assigning seqnums) without touching the data itself. wdyt? maybe overkill

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,343 @@
...
+The method removes the batch from the in-flight set so the background heartbeat stops refreshing it,
+and calls `dequeue()` on the queue consumer to mark the location as done.
+This may trigger cleanup of the manifests and the data batch in object storage if the
+`done_cleanup_threshold` is reached.
+
+When the collector is dropped, the background heartbeat task is cancelled.
+In-flight batches that are not acknowledged will eventually become stale and can be re-claimed
+by another collector.
+
+### Queue
```
agavra: nit: we can move this first? it seems like this knowledge is required to understand the previous two sections better

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,343 @@
...
+Once the data is read and processed, the producer can mark the location as done.
+To avoid contention on the producer manifest, the queue consumer keeps claimed and done locations in the consumer manifest.
+However, once in a while, the queue consumer needs to remove done locations from the producer manifest and
+from the consumer manifest to avoid unbounded growth of those manifests.
+After a configured number of done locations, the queue consumer cleans up both manifests.
+In this way the queue consumer avoid contending too much with the queue producers for the producer manifest.
+
+#### Manifest Format
+
+Both manifests are stored as JSON files in object storage.
```
agavra: nit: we should consider flatbuffer for this as well (or whatever format allows us to append without reading the existing ones)

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,343 @@
...
+Since we assume one ingestor per availability zone and a single collector which limits the contention, we decided
+against the broker.
+One reason for deciding against the broker is that the broker would be an additional component that a user needs to
+deploy and operate.
+We want to keep the ingest as simple as possible.
+In the future, it might be necessary to revise this decision and implement the stateless broker approach if
+contention gets worse.
+
+
+### Using one single manifest for the queue
```
agavra, line 353: lets add a section on a slatedb WAL-style algorithm for the queue where we just attempt to write-if-not-exists and increment the id of the file by 1 becuase then you have to deal with GC a bit more aggressively

ingest/rfcs/0001-stateless-ingest.md
cadonna: Yeah, that was more a hack/placeholder for the prototype.
My open question

>What format should we use for the data batches?

that you also commented is about this.
We should definitely use a binary format that does not need any deserialization for passing key-value pairs to the OpenData DB. 

I will propose a binary format.

cadonna: Good point about the streaming put, compression and encryption.

ingest/rfcs/0001-stateless-ingest.md
cadonna: I do not see any reason. I will change it.

ingest/rfcs/0001-stateless-ingest.md
cadonna: That needs some improvement. I am hoping that with a good format for the batches/better interface we can eliminate deserialization.

ingest/rfcs/0001-stateless-ingest.md
cadonna: Good point!

ingest/rfcs/0001-stateless-ingest.md
cadonna: The metrics you propose are useful.

With my question I meant to discover batches that for some reason are claimed but then they time out. In this case more recent batches might overtake those problematic batches and produce out-of-order batches. If claims followed by timeouts for a batch happens to often, inspection is needed.

The situation I described cannot be discovered by looking at the number of pending, claimed, and done batches.

ingest/rfcs/0001-stateless-ingest.md
cadonna: I agree that the interface can be improved here. This was more a intermediate solution for the prototype. Let me fix that.

ingest/rfcs/0001-stateless-ingest.md
cadonna: Correct!
If we hit higher contention, we need to reconsider the broker option. We just discarded the broker option because we assumed acceptable contention.

ingest/rfcs/0001-stateless-ingest.md
cadonna: Done!

ingest/rfcs/0001-stateless-ingest.md
cadonna: I recall that there were some discussions about using the ingestors as caches. For that, we would at least need the distinction between key and value. But we can change the bytes to key-value bytes once we reach that point in development.

ingest/rfcs/0001-stateless-ingest.md
cadonna: See https://github.com/opendata-oss/opendata/pull/252/changes#r2863109719

ingest/rfcs/0001-stateless-ingest.md
cadonna: That sounds interesting; however, let's leave it for the future.

Thinking about it, I think we have a GC problem here because data that is written to object storage but not written to the queue can never be removed. I have to think more about GC. But that's not related to your command.

ingest/rfcs/0001-stateless-ingest.md
cadonna: Thinking about it again, I am not sure I am following. What do you exactly mean by "append without reading the existing ones"?

cadonna: First I thought you mean without deserializing the read manifest.

ingest/rfcs/0001-stateless-ingest.md
cadonna: @agavra @apurvam I added a binary format for the data batch. I did not add any compression and encryption metadata since I need to figure out a couple of things about streaming PUTs to S3. I will update the format after that.

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,349 @@
...
+
+By calling `ack()` the caller confirms that the batch has been fully processed.
+The method removes the batch from the in-flight set so the background heartbeat stops refreshing it,
+and calls `dequeue()` on the queue consumer to mark the location as done.
+This may trigger cleanup of the manifests and the data batch in object storage if the
+`done_cleanup_threshold` is reached.
+
+When the collector is dropped, the background heartbeat task is cancelled.
+In-flight batches that are not acknowledged will eventually become stale and can be re-claimed
+by another collector.
```
samsond, line 315: Nice, stale/re-claim via heartbeat makes sense.

What if a collector claims batch A (head of pending), crashes before ack() — will the next collector re-claim A before later entries so ingest order is preserved?

samsond:
Open question: What are the exact delivery semantics of the ingest layer?

Given that unacknowledged batches can become stale and be re-claimed by another collector, there appears to be a window where a batch may be fully processed but not yet acknowledged. If the collector crashes in that window, the same batch could later be processed again.

Does this behavior intentionally define the ingest layer as at-least-once? If so, should idempotent batch processing be made an explicit requirement to ensure correctness under reprocessing?

ingest/rfcs/0001-stateless-ingest.md
cadonna: This depends on the heartbeat that is stored with the claimed location. If the heartbeat timeout expired before the new collector claimed its first location, it will re-claim the location previously claimed by the failed collector, otherwise it will claim pending locations until the claimed location times out.
The order of collection may be different from the ingest order.

With a single collector mode, we could improve the situation by using an epoch. The collector gets an epoch when it starts up. All claimed locations that have an epoch less than the epoch of the collector are re-claimed with the new epoch.

cadonna:
> Open question: What are the exact delivery semantics of the ingest layer?
> 
> Given that unacknowledged batches can become stale and be re-claimed by another collector, there appears to be a window where a batch may be fully processed but not yet acknowledged. If the collector crashes in that window, the same batch could later be processed again.
> 
> Does this behavior intentionally define the ingest layer as at-least-once? If so, should idempotent batch processing be made an explicit requirement to ensure correctness under reprocessing?

For the first cut, providing at-least-once delivery semantics is on purpose. If it is is clear that the ingest provides at-least-once, we do not need to require idempotent processing, IMO. The users of the ingest need to account for at-least-once in their processing. If their processing is not idempotent, they have to evaluate if duplicates are fine.

For exactly-once delivery semantics, we could keep track of the processed batches in the OpenData subsystem that ingested the data. For that the batch needs to be ingested in an atomic operation and the atomic operation should also include updating a list of ingested batches. Or something like this.

cadonna:
I added  description about order guarantees and delivery guarantees.

ingest/rfcs/0001-stateless-ingest.md
cadonna: Done

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,402 @@
...
+
+#### Manifest Format
+
+Both manifests are stored as JSON files in object storage.
+
+The producer manifest contains a list of pending locations:
+```json
+{
+  "pending": [
+    "data/01J5T4R3KXBMZ7QV9N2WG8YDHP.batch",
```
rodesai, line 120: It would be good to make this a little more structured than just a location string. I suspect that we may at some point want to allow the ingestors to embed some small amount of db-specific metadata here so that the collector can do db maintenance without being forced to read the data files. To start with we can just make this a struct with a single field like "path".

ingest/rfcs/0001-stateless-ingest.md
rodesai: We could also revisit per-ingestor queues with some centralized manifest for registering/terminating ingestors. So something like:

Files:
- /ingestors.manifest
- /ingestors/<id>/producer // producer queue
- /ingestors/<id>/data/... // data files
- /consumer // consumer queue

Ingestor:
- when starting up, register itself with ingestors.manifest and periodically heartbeat
- all writes go to ingestor-specific queue

Collector:
- periodically poll `ingestors.manifest` to detect new ingestors
- periodically poll the queues for each ingestor and record progress in common consumer queue file
- when an ingestor is determined to be "dead", fence it by updating its producer queue file and remove it from the shared manifest.

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,402 @@
...
+When the collector is dropped, the background heartbeat task is cancelled.
+In-flight batches that are not acknowledged will eventually become stale and can be re-claimed
+by another collector.
+
+### Delivery guarantees
+
+Once entries are confirmed to be durable they are guaranteed to be delivered to the caller of the collector 
+at least once.
+If the collector fails after it claimed a location and processed the corresponding batch, but before it was able to 
+acknowledge the batch, the new collector will re-read the batch and re-process it.
```
rodesai, line 322: did you consider just tracking the consumption status in slatedb? that would take us a step further towards more exact ingestion semantics because we can get rid of the case where data is written but not marked as such by writing the queue progress in the same atomic batch as the data write

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,402 @@
...
+
+The delivery guarantees of the entries that were not confirmed to be durable depends on the progress tracking of the
+caller of the stateless ingest.
+If they track the progress and re-ingest unacknowledged entries they can achieve at-least-once guarantee.
+
+### Observability
+
+TBD
+
+Some ideas:
```
rodesai, line 332: it would be good to include some e2e latency measurement. we'd need to record timestamps in the metadata

ingest/rfcs/0001-stateless-ingest.md
rodesai: Recording ingestion time for e2e latency tracking is another potential use case.

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,402 @@
...
+}
+```
+The queue producer manifest takes the name specified in `manifest_path`.
+The queue consumer manifest is derived from `manifest_path` by adding `consumer` in front of the file extension
+or to the end of the `manifest_path` if the path does not have an extension.
+
+The collector internally creates a queue consumer and an object store client from the configuration.
+It also spawns a background heartbeat task that periodically refreshes the timestamps of all
+in-flight batches in the consumer manifest. The heartbeat interval is derived from the configuration
+as `heartbeat_timeout_ms / 3`, ensuring claims stay fresh well before they expire.
```
rodesai, line 292: why do we need this whole heartbeat mechanism? I don't think we can practically have multiple consumers of a single queue since slatedb is single-writer. So why not just enforce that there is a single consumer and fence older consumers?

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,402 @@
...
+The ingestor flushes the batches of ingested entries to object storage and appends the locations of the
+flushed objects to the queue with the internally used queue producer (`q-producer` in the diagram).
+Flushes are triggered after a given time interval elapsed or if a batch of the ingested data exceeds a given size.
+
+The API of the ingestor is the following:
+```rust
+impl Ingestor {
+  pub fn new(config: IngestorConfig, clock: Arc<dyn Clock>) -> Result<Self> { ... }
+
+  pub async fn ingest(&self, entries: Vec<Bytes>) -> Result<WriteWatcher> { ... }
```
rodesai, line 176: Similar to the feedback above - it would be good to have a metadata vs data distinction in this api. So each entry should contain discrete metadata vs data components (which can themselves just be `Bytes`). I think this is reasonable as an extension as long as the serialized structure supports it.

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -0,0 +1,402 @@
...
+
+  pub async fn close(self) -> Result<()> { ... }
+}
+```
+
+An ingestor is constructed by calling the method `new()` passing to it the configuration and a clock.
```
rodesai, line 182: what's the clock used for on the ingestor? On the collector side I'm guessing its for heartbeating?

rodesai (commented):
This is a great start towards stateless ingest. The API is super simple so it should be pretty straightforward to play with different approaches to implementing the queue in slatedb as this evolves. I have a few bits of feedback (sorry it's rolling in late).

First, we should more explicitly model metadata vs data, especially in the persisted data model. To start with we could make each entry of the queue a struct so that it can be extended by the various dbs to include more specific metadata about each system's data rather than just specifying a location. We don't need to add that just yet, but we should leave the door open.

Second, I'm a little concerned about how loose the protocol is for consumption. In particular, it seems like it'll be somewhat common to reorder writes because the consumer handles failures using a heartbeat. This is probably good enough for systems that are insert-only like tsdb, though still a bit weird as you can introduce temporary gaps. For something like log (or even vector) I don't think that's sufficient. I think we'd need to at least guarantee that if a client successfully writes batches W1 and then W2, that W2 is applied after W1. It's possible that W1 could be re-appended after W2 due to a zombie ingestor, but that's not really different from a zombie producer in Kafka and it should be possible to deal with in the client. To guarantee this we'd need to process all unprocessed writes from the collector. We could do this by just processing everything that's not committed on a collector failover, and optionally adding some fencing mechanism to fence old collectors. One option would be to just store the collector status directly in slate - then you could even guarantee that every item that's in the queue is processed exactly once.

ingest/rfcs/0001-stateless-ingest.md
cadonna: I agree that it is useful to store more metadata in the queue and make it extensible. I am also not too happy with the JSON format. Since a new entry does not depend on any existing entry, we could use a format that does not need deserialization of existing entries when new entries are appended.

ingest/rfcs/0001-stateless-ingest.md
cadonna: I mentioned what you describe in the following comment: https://github.com/opendata-oss/opendata/pull/252#issuecomment-3984072240

ingest/rfcs/0001-stateless-ingest.md
cadonna: I wanted to keep it general enough to allow multiple consumers.
I agree with the fencing mechanism if we assume a single consumer. In fact, I described something like this in the following comment: https://github.com/opendata-oss/opendata/pull/252#discussion_r2872059854

On the other hand, if we decide to implement partitioned multiple consumers and we partition the queue, we probably do not need the heartbeat. We could use epochs and fencing because for each queue there would effectively only be one consumer.

ingest/rfcs/0001-stateless-ingest.md
cadonna: The clock is used for the flush interval. Batches are flushed at regular intervals or when they reach a given size.

ingest/rfcs/0001-stateless-ingest.md
cadonna: I had a similar design before I started with the single queue file. I started with the single queue file since it seemed simpler and we get a (weak) global ordering of the batches. If the consumer periodically polls the ingestor queues, the ordering of the batches depend on the consumer. OK, there could be some merge sort involved to sort the batches across the ingestor queues, but that is much more complex than just having a single queue.
On the other side, how much global ordering do we really need to guarantee given that the collector does not sort the entries across batches? 
The intent of my design was to have something with which we can experiment, discuss, and then improve. So I am completely open to changes that better satisfy given requirements.

PR #253, opened 2026-02-26
Add ingest crate with queue producer and consumer

cadonna:
## Summary

Adds the The foundational layer for stateless ingest: workspace setup, error types,  and the manifest-based queue (producer + consumer with CAS-loop conflict resolution).

## Test Plan

How was this tested?

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/ingest/src/queue.rs
+++ b/ingest/src/queue.rs
@@ -20,6 +20,7 @@ fn millis(time: SystemTime) -> i64 {
 }
 
 const MANIFEST_VERSION: u16 = 1;
+const UNINITIALIZED_EPOCH: u64 = u64::MAX;
 const ENTRY_LEN_SIZE: usize = 4;
 const LOCATION_LEN_SIZE: usize = 2;
 const INGESTION_TIME_MS_SIZE: usize = 8;
@@ -590,7 +591,7 @@ impl QueueConsumer {
                 object_store,
                 manifest_path,
             },
-            epoch: AtomicU64::new(0),
+            epoch: AtomicU64::new(UNINITIALIZED_EPOCH),
             clock,
             counter: ConflictCounter::new(),
             queue_len: AtomicU64::new(0),
@@ -602,7 +603,10 @@ impl QueueConsumer {
     pub async fn initialize(&self) -> Result<()> {
         loop {
             let (mut manifest, version) = self.read_manifest().await?;
-            let new_epoch = manifest.epoch + 1;
+            let mut new_epoch = manifest.epoch.wrapping_add(1);
+            if new_epoch == UNINITIALIZED_EPOCH {
+                new_epoch = new_epoch.wrapping_add(1);
+            }
             manifest.set_epoch(new_epoch);
             match self.write_manifest(&manifest, version).await {
                 Ok(()) => {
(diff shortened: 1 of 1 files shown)


hachikuji:
@cadonna One question about the RFC. Since all batches are ordered, I was wondering if it would make sense to assign them sequence numbers. For example:
```json
 { "next_seq": 8, "pending": [
      {"seq": 5, "loc": "data/01J5T4...batch"},
      {"seq": 6, "loc": "data/01J5T4...batch"},
      {"seq": 7, "loc": "data/01J5T4...batch"}
  ]}
```
I was thinking it might simplify the consumer implementation if it could track progress using sequence watermarks. In particular, would we need the done list at all?

cadonna:
> @cadonna One question about the RFC. Since all batches are ordered, I was wondering if it would make sense to assign them sequence numbers. For example:
> 
> ```json
>  { "next_seq": 8, "pending": [
>       {"seq": 5, "loc": "data/01J5T4...batch"},
>       {"seq": 6, "loc": "data/01J5T4...batch"},
>       {"seq": 7, "loc": "data/01J5T4...batch"}
>   ]}
> ```
> 
> I was thinking it might simplify the consumer implementation if it could track progress using sequence watermarks. In particular, would we need the done list at all?

Yeah, that is a good point. We just need a place where to persist the watermark for fail-over. That could be done in slateDB itself. That has the nice side-effect that we would get exactly-once ingest if we write the watermark to slateDB in the same batch write as the ingested data.

However, this assumes that we will have a single writer or multiple writer with partitioned queues. Currently, we assume a single writer but I would like to keep the design general.

ingest/src/queue.rs
```diff
@@ -0,0 +1,1563 @@
...
+    *offset += 2;
+
+    let location = String::from_utf8(data[*offset..*offset + location_len].to_vec())
+        .map_err(|e| Error::Serialization(e.to_string()))?;
+    *offset += location_len;
+
+    let ingestion_time_ms = i64::from_le_bytes(data[*offset..*offset + 8].try_into().unwrap());
+    *offset += 8;
+
+    let metadata_len = entry_len - 2 - location_len - 8;
```
samsond: I noticed that if a corrupt manifest has entry_len < 10 or location_len > entry_len - 10, the calculation will underflow.

Would it make sense to add validation after line 197 to catch these cases early?`

ingest/src/queue.rs
```diff
@@ -0,0 +1,1563 @@
...
+            Some(decode_entry(self.data, &mut self.offset, self.entries_end))
+        } else if self.appended_remaining > 0 {
+            self.appended_remaining -= 1;
+            Some(decode_entry(
+                self.appended,
+                &mut self.appended_offset,
+                self.appended.len(),
+            ))
+        } else {
+            None
```
samsond, line 412: The iterator returns None when remaining reaches zero, but doesn't verify that all bytes were consumed. If a corrupted manifest has footer.entry_count = 1 but the payload contains 2  entries, the iterator stops after the first entry and silently ignores the rest.

Could we add validation here to check if self.offset != self.entries_end and return  Error::Serialization instead? This would catch footer count mismatches.

ingest/src/queue.rs
```diff
@@ -0,0 +1,1563 @@
...
+                .filter(|entry| !done_set.contains(entry.location.as_str()))
+                .collect();
+            let producer = ProducerManifest::from_entries(&kept);
+            self.counter.record_write();
+            match self
+                .producer_store
+                .write_producer_manifest(&producer, version)
+                .await
+            {
+                Ok(()) => break,
```
samsond: After successfully writing the filtered manifest, the code breaks without updating producer_len. 

This causes len() to return stale values after cleanup. Could we update the cache here before  breaking, similar to how read_producer_manifest updates it at line 622?

cadonna:
> However, this assumes that we will have a single writer or multiple writer with partitioned queues. Currently, we assume a single writer but I would like to keep the design general.

We decided to assume a single collector. I changed the queue to fence old consumers with an epoch.

I also added the sequence number proposed in https://github.com/opendata-oss/opendata/pull/253#issuecomment-3987797723

ingest/src/queue.rs
cadonna: Yeah, that makes sense!
I added verifications.

ingest/src/queue.rs
cadonna: I added a verification.

ingest/src/queue.rs
cadonna: Good catch!
I update the queue length also on write now.

hachikuji:
@cadonna Thanks for the updates. I reviewed the latest RFC and it LGTM. I was wondering if the sequence number was doing much for us since we seem to not need it for tracking positions. However, it is a small addition and it probably helps debugging if nothing else, so I'm in favor of keeping it.

I'll get through the code tomorrow.

cadonna:
> @cadonna Thanks for the updates. I reviewed the latest RFC and it LGTM. I was wondering if the sequence number was doing much for us since we seem to not need it for tracking positions. However, it is a small addition and it probably helps debugging if nothing else, so I'm in favor of keeping it.
> 
> I'll get through the code tomorrow.

I am afraid I am not completely following, the sequence number is used at least for selectively reading entries and removing entries from the queue. See read `QueueConsumer::read()` and `QueueConsumer::dequeue()`. Maybe that is not described too much in the RFC.

How the sequence number is passed to and from the queue is part of the next PRs in this series.

hachikuji:
> I am afraid I am not completely following, the sequence number is used at least for selectively reading entries and removing entries from the queue. See read QueueConsumer::read() and QueueConsumer::dequeue(). Maybe that is not described too much in the RFC.

Oh yeah, but I didn't see that as fundamentally different from dequeuing by count. Perhaps I'm missing something. In any case, I prefer to have the sequence numbers. They will be a nice anchor for reasoning about the system.

PR #254, opened 2026-02-26
[timeseries] Fix bucket preloading to account for PromQL offset/@ modifiers

### LINKED ISSUE #249: [timeseries] PromQL instant queries can miss offset data when adjusted evaluation time falls outside preloaded buckets
##  Description
Instant query bucket discovery is done before PromQL time modifiers are applied. As a result, offset (and other time modifiers) can shift evaluation into buckets that were never loaded, producing empty or partial results even when data exists.

This is most visible when the offset crosses an hour bucket boundary (current bucketing model), e.g. offset 30m works but offset 1h returns no data.

## Steps to Reproduce

1. Baseline: explicit time query returns data

```bash
  curl -s --data-urlencode 'query=my_metric' \
    --data-urlencode 'time=1772083854' \
    'http://localhost:9090/api/v1/query'
  # expected: non-empty result
```

2. Equivalent offset form returns empty

```bash
  curl -s --data-urlencode 'query=my_metric offset 1h' \
    'http://localhost:9090/api/v1/query'
  # actual: empty result
```

3. Wide range query can still see the same data
```bash
  curl -s --data-urlencode 'query=my_metric' \
    --data-urlencode 'start=1772065000' \
    --data-urlencode 'end=1772087000' \
    --data-urlencode 'step=300' \
    'http://localhost:9090/api/v1/query_range'
  # expected/actual: data present in range response
```
##  Expected Behavior
my_metric offset 1h at time T should match my_metric at time T-1h (subject to lookback/staleness semantics), regardless of bucket boundary.

## Actual Behavior
Offset queries may return empty when adjusted time falls in an older bucket not included in the preloaded reader window.

## Root Cause
  - Instant query routing preloads buckets for [query_time - lookback_delta, query_time] using unadjusted
    stmt.start.
  - Evaluator applies time modifiers (offset, @) later during selector evaluation.
  - If adjusted eval time is outside preloaded buckets, selectors find no samples.

Note: Range queries are safer when [start, end] is wide, but can also be affected if modifiers push effective eval time outside [start - lookback_delta, end].

## Proposed Fix
Before constructing query readers, compute required evaluation-time bounds from the AST and expand preload range accordingly.

  - Extend preload start backward for positive offsets.
  - Extend preload end forward for negative offsets.
  - Account for @ <ts>, @ start(), and @ end() semantics.
  - Apply this in both instant and range query paths.

##  Suggested Validation Tests

  1. Instant query: metric offset 1h equals explicit time=T-1h.
  2. Instant query across hour boundary: 39m vs 41m both behave correctly when data exists.
  3. Range query with large positive/negative offsets still returns expected series.

apurvam:
## Summary

Instant and range query bucket discovery used the unadjusted query time, missing data when offset or @ modifiers shifted evaluation into unloaded buckets. Add compute_preload_bounds() to walk the AST and determine the full time window needed, then use it in both query() and query_range().

Also clamp computed bounds so that start > end or negative end values cannot cause internal errors from get_buckets_in_range(), and fix the ms-to-seconds ceil conversion to not over-fetch on exact boundaries.

## Related Issues

Fixes #249 

## Test Plan

Added unit and integration tests to check that the right data is loaded for the offset / @ modifiers in the query. these tests fail without the fix.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -8,7 +8,7 @@ use crate::index::{ForwardIndexLookup, InvertedIndexLookup, SeriesSpec};
 use crate::model::Sample;
 use crate::model::SeriesFingerprint;
 use crate::model::{Label, SeriesId, TimeBucket};
-use crate::promql::functions::FunctionRegistry;
+use crate::promql::functions::{FunctionRegistry, PromQLArg};
 use crate::promql::selector::evaluate_selector_with_reader;
 use crate::promql::timestamp::Timestamp;
 use crate::query::QueryReader;
@@ -1218,7 +1218,8 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
             ExprResult::InstantVector(samples) => {
                 // Try instant vector function first
                 if let Some(func) = registry.get(call.func.name) {
-                    let result = func.apply(samples, eval_timestamp_ms)?;
+                    let result =
+                        func.apply(PromQLArg::InstantVector(samples), eval_timestamp_ms)?;
                     Ok(ExprResult::InstantVector(result))
                 } else {
                     Err(EvaluationError::InternalError(format!(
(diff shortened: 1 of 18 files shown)


timeseries/src/promql/evaluator.rs
```diff
@@ -365,6 +365,263 @@ impl ExprResult {
...
+/// Walk the AST and compute the time window needed for bucket preloading.
+///
+/// For each selector, compute the effective evaluation time by applying
+/// @ and offset modifiers, then expand by lookback_delta (vector) or
+/// range (matrix). Returns (earliest_secs, latest_secs) covering all
+/// selectors in the expression.
+///
+/// Returns `None` when the expression contains no selectors (e.g. `1 + 2`),
+/// allowing the caller to fall back to the default window.
+pub(crate) fn compute_preload_bounds(
```
agavra: I think this is problematic, we need disjoint ranges. 

Claude's summary confirming my suspicion:

The `merge_bounds` approach collapses all selector windows into a single `(min_start, max_end)` range. For expressions with widely separated time windows (e.g. `metric - metric offset 5d`), this causes `get_buckets_in_range` to load every intermediate bucket — potentially ~120 hour-buckets when only 2-4 are needed.

This is fine for correctness, but could be a significant performance issue. A fix would be returning disjoint ranges from `compute_preload_bounds` instead of a single merged interval.

rodesai:
I'm wondering why do we need to precompute the range at all? We could just change the query reader to (1) always read from a fixed snapshot that it stores internally and (2) just load buckets on demand as needed.

timeseries/src/promql/evaluator.rs
apurvam: I updated the PR and removed the `merge_bounds` method, instead creating a query reader per disjoint range. This should address the performance issue you noted.

apurvam:
> I'm wondering why do we need to precompute the range at all? We could just change the query reader to (1) always read from a fixed snapshot that it stores internally and (2) just load buckets on demand as needed.

This is indeed a better approach. I _think_ it's a bigger refactor? I filed https://github.com/opendata-oss/opendata/issues/256 to track it as an improvement.

timeseries/src/promql/evaluator.rs
```diff
@@ -365,6 +365,290 @@ impl ExprResult {
...
+        .map(|(lo, hi)| {
+            let start_secs = lo.div_euclid(1000);
+            let end_secs = hi.div_euclid(1000) + i64::from(hi.rem_euclid(1000) != 0);
+            (start_secs, end_secs)
+        })
+        .collect();
+    normalize_ranges(ranges_secs)
+}
+
+/// Sort ranges by start, merge overlapping ones, and clamp start to >= 0.
```
agavra: why are we clamping start to >= 0? negative timestamps represent earlier than UNIX epoch and are valid timestamps

timeseries/src/promql/tsdb_router.rs
```diff
@@ -2134,4 +2150,156 @@ mod tests {
         assert!(data.contains_key("http_requests_total"));
         assert!(data.get("http_requests_total").unwrap().len() == 4)
     }
+
+    // ── offset/@ preload integration tests ───────────────────────────
```
agavra, line 2154: consider adding some `promqltest` tests as well that cover this

timeseries/src/promql/evaluator.rs
apurvam: Good call. I dropped this.

PR #257, opened 2026-02-27
[timeseries] Remove write_with_options and ingest_samples on TimeSeriesDb

hachikuji:
This patch removes `write_with_options` which was only used to pass `await_durable`. It makes the API a little simpler to require users to explicitly `flush`. I've also removed `ingest_samples` which was inadvertently made public.

### DIFF
--- a/timeseries/src/timeseries.rs
+++ b/timeseries/src/timeseries.rs
@@ -8,7 +8,7 @@ use std::sync::Arc;
 
 use common::{StorageRuntime, StorageSemantics, create_storage};
 
-use crate::config::{Config, WriteOptions};
+use crate::config::Config;
 use crate::error::Result;
 use crate::model::Series;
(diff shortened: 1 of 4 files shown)


timeseries/src/timeseries.rs
```diff
@@ -140,10 +136,14 @@ impl TimeSeriesDb {
...
     pub async fn write_with_options(
         &self,
-        _series: Vec<Series>,
-        _options: WriteOptions,
+        series: Vec<Series>,
+        options: WriteOptions,
     ) -> Result<()> {
-        todo!()
+        self.tsdb.ingest_samples(series).await?;
+        if options.await_durable {
```
hachikuji: To be honest, I'm not really sure we want to keep this. I kind of preferred having users explicitly call `flush` for durability.

Issue #128, opened 2026-01-28
Use Slatedb-backed memory storage for tests


hachikuji:
Currently our tests rely on an in-memory implementation of the Storage trait. It would be nice to rely on Slatedb instead with an in-memory object storage. This is straightforward, but then we need to be more diligent about calling `close()` since SlateDb will leak resources. I am not sure the best pattern for this, but perhaps a macro could work. The invocation might look like this:

```rust
#[opendata::test]
fn my_test(storage: &Storage) { ... }
```

which could be expanded into something like this:
```rust
#[tokio::test]
async fn my_test() {
    let storage = Storage::in_memory().await;
    my_test__inner(&storage).await;
    storage.close().await;
}

async fn my_test__inner(storage: &Storage) {
    ...
}
```
I'm not sure how to do this or if there are simpler alternatives.

KyleSmith19091:
Is this being worked on? I've got an idea for an implementation.

cadonna:
Welcome @KyleSmith19091! 
I assigned the issue to you. Looking forward to your contribution!

KyleSmith19091:
Awesome! thank you @cadonna will get started on it

KyleSmith19091:
This issue can be closed

PR #258, opened 2026-02-27
Storage test macro

KyleSmith19091:
## Summary
This PR adds a procedural macro `#[storage_test]` that abstracts the setup and teardown of storage-backed tests. It also adds a new macros crate. The crate is also where more general opendata macros can be in the future. The code the macro generates can be summarised as follows:

```rust
#[storage_test(merge_operator = OpenTsdbMergeOperator)]
async tsdb_test(storage: Arc<dyn Storage>) {
   let tsdb = Tsdb::new(storage);
  ... rest of test
}
```
Is expanded to

```rust
#[::tokio_test]
async fn tsdb_test() {
  let storage: std::sync::Arc<dyn Storage> = std::sync::Arc::new(
         common::storage::in_memory::InMemoryStorage::with_merge_operator(
              std::sync::Arc::new(#merge_op)
        )
  );
 tsdb_test__inner(storage.clone()).await;
 let _ = storage.close().await;
}

async fn tsdb_test__inner(storage: std::sync::Arc<dyn Storage>) {
    ...test body
}
```

## Related Issues

https://github.com/opendata-oss/opendata/issues/128

### Macro Crate testing
- The macro is tested in macros/src/test/storage.rs
- We test the macro expansion by using a combination of string assertions and AST assertions.

### Usage testing
- The tests that created an instance of in memory storage were replaced with the macro and the tests should still pass.

## Checklist

- [x] Tests added/updated


KyleSmith19091:
Some things that need to be discussed, that I'm unsure about:
- In macros/src/test/storage.rs:95 the result of closing is ignored, not sure if this is acceptable?
- How pedantic do we need to be about closing, if the (inner) test function panics then we might skip closing storage explicitly, but I guess the process will probably crash so we won't leak any resource (unless a test no longer uses an in memory store)

macros/src/test/storage.rs
```diff
@@ -0,0 +1,299 @@
...
+        quote! {
+            let storage: std::sync::Arc<dyn Storage> = std::sync::Arc::new(
+                common::storage::in_memory::InMemoryStorage::default()
+            );
+        }
+    };
+
+    quote! {
+         #[::tokio::test]
+        #[allow(unused_must_use)]
```
hachikuji, line 171: Is this for the `storage.close()`?

macros/src/test/storage.rs
```diff
@@ -0,0 +1,299 @@
...
+    } else {
+        quote! {
+            let storage: std::sync::Arc<dyn Storage> = std::sync::Arc::new(
+                common::storage::in_memory::InMemoryStorage::default()
+            );
+        }
+    };
+
+    quote! {
+         #[::tokio::test]
```
hachikuji: I wonder if it would be worthwhile to pass through some of the common attributes, such as "multi_thread". No need to cover everything (such as start_paused), but we do use multi_thread quite a bit.

macros/src/test/storage.rs
```diff
@@ -0,0 +1,299 @@
...
+            ));
+        }
+
+        Ok(TestMacroArgs { merge_operator })
+    }
+}
+
+pub fn test_impl(args: TokenStream, input: TokenStream) -> TokenStream {
+    // parse arguments to the macro (see Parse impl for TestMacroArgs for implementation)
+    let args_parsed = parse2::<TestMacroArgs>(args).expect("failed to parse macro args");
```
hachikuji: I am not that familiar with macro idioms, but I wonder if we should use a compile error or `syn::parse_macro_input` instead of the panic.

hachikuji:
@KyleSmith19091 

> In macros/src/test/storage.rs:95 the result of closing is ignored, not sure if this is acceptable?

Does it make sense to fail instead?

> How pedantic do we need to be about closing, if the (inner) test function panics then we might skip closing storage explicitly, but I guess the process will probably crash so we won't leak any resource (unless a test no longer uses an in memory store)

Yeah, that seems acceptable. Not too worried about leaks from failing tests.

macros/src/test/storage.rs
rodesai: I think a clean way to do this would be to look out for the tokio::test macro when parsing the input, and use the one specified if provided. If it's not provided then the test can just default to vanilla `tokio::test`

macros/src/test/storage.rs
KyleSmith19091: Yes that is to ignore the unused result from calling .close method

macros/src/test/storage.rs
KyleSmith19091: Yes will do

KyleSmith19091: Ok I refactored it to copy the tokio test macro in the expanded output if specified, if not it will use the default tokio test macro:

```rust
#[storage_test(merge_operator = OpenTsdbMergeOperator)]
#[tokio::test(flavor = "multi_threaded", worker_threads = 2)]
async tsdb_test(storage: Arc<dyn Storage>) {
   let tsdb = Tsdb::new(storage);
  ... rest of test
}
```

Will be expanded to
```rust
#[tokio::test(flavor = "multi_threaded", worker_threads = 2)]
async fn tsdb_test() {
  let storage: std::sync::Arc<dyn Storage> = std::sync::Arc::new(
         common::storage::in_memory::InMemoryStorage::with_merge_operator(
              std::sync::Arc::new(#merge_op)
        )
  );
 tsdb_test__inner(storage.clone()).await;
 let _ = storage.close().await;
}

async fn tsdb_test__inner(storage: std::sync::Arc<dyn Storage>) {
    ...test body
}
```

macros/src/test/storage.rs
KyleSmith19091: The `syn::parse_macro_input` is definitely more ergonomic, but it returns the `proc_macro::TokenStream` type, which can only be used in the root of the crate. Using a compile error is still better so I refactored it to be the following:
```rust
// parse arguments to the macro (see Parse impl for TestMacroArgs for implementation)
let args_parsed = match parse2::<TestMacroArgs>(args)
  Ok(v) => v,
  Err(e) => return e.to_compile_error(),
};

// parse string below macro as free standing function
let item_fn = match parse2::<ItemFn>(input) {
  Ok(v) => v,
  Err(e) => return e.to_compile_error(),
};
```

KyleSmith19091:
The reason for adding the macro is a bit unfortunate, I completely agree that it would be better in the common crate, but according to the [docs](https://doc.rust-lang.org/reference/procedural-macros.html) macros need to be in their own crate.

Screenshot 2026-02-28 at 11 06 28

macros/src/test/storage.rs
```diff
@@ -0,0 +1,376 @@
...
+            .get("my_test")
+            .expect("Should have wrapper function named 'my_test'");
+        assert!(
+            has_attribute(wrapper, "tokio::test"),
+            "Should have #[tokio::test] attribute"
+        );
+
+        // verify wrapper body contains storage creation with merge operator
+        assert!(wrapper.attrs.iter().any(|attr| {
+            println!("{}", attr.to_token_stream());
```
hachikuji: nit: remove this?

macros/src/test/storage.rs
```diff
@@ -0,0 +1,376 @@
...
+    parse_quote, parse2,
+};
+
+/// Parsed arguments for the storage test macro
+struct TestMacroArgs {
+    merge_operator: Option<Expr>,
+}
+
+impl Parse for TestMacroArgs {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
```
hachikuji, line 17: What happens if the user's function provides no storage parameter or if the storage parameter is not named "storage"?

macros/src/test/storage.rs
KyleSmith19091: Yes sorry this was just when I was debugging, I removed it

macros/src/test/storage.rs
KyleSmith19091: Good question, I did not think about it. What will happen is since the macro expands in-place even if you don't have a storage parameter the body of the function is still valid since the wrapped function will always have a parameter called storage.

I upgraded the macro to validate the type of the parameter of the function it is placed above. It will then also respect the name of the parameter so you don't have to call the parameter "storage". Also it will throw a compile time error if the parameter is not explicitly set on the function.

timeseries/src/tsdb.rs
```diff
@@ -1284,7 +1271,6 @@ mod tests {
         ));
     }
 
-    #[tokio::test]
```
hachikuji, line 1287: Looks like we lost this?

KyleSmith19091: Yes not sure why I removed it, it's back now

PR #260, opened 2026-02-28
[timeseries] Implement RFC 3

hachikuji:
This patch does most of the patch to expose the read APIs documented in RFC 3. I had to make minor changes to the RFC in order to adapt to the existing prometheus surface area (in particular adding QueryValue to distinguish scalar values). Since we now have the read API on TimeSeries, I thought we didn't need `PromqlRouter` anymore, so I've removed it and migrated test coverage. The most interesting code is in response.rs. This contains the logic to convert from the data types used in the embedded API to the JSON output. 

Note I'm working on a bench to make sure there is no regression on the serialization to JSON compared to main. I will post the results shortly.

### DIFF
--- a/timeseries/rfcs/0003-read-api.md
+++ b/timeseries/rfcs/0003-read-api.md
@@ -55,8 +55,10 @@ impl TimeSeriesDb {
     ///
     /// ```rust
     /// let result = tsdb.query("rate(http_requests_total[5m])", None).await?;
-    /// for sample in result.into_samples() {
-    ///     println!("{}: {}", sample.labels.metric_name(), sample.value);
+    /// for rs in result.into_matrix() {
+    ///     for (ts, val) in &rs.samples {
+    ///         println!("{}: {} = {}", rs.labels.metric_name(), ts, val);
+    ///     }
     /// }
     /// ```
     pub async fn query(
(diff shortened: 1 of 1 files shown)


hachikuji:
## Response serialization benchmark

Branch `response-bench` adds a Criterion microbenchmark (`timeseries/benches/response_serialization.rs`) covering the full path from model types → response conversion → JSON bytes.

```
cargo bench -p opendata-timeseries --features testing --bench response_serialization
```

### Results: `timeseries-read-api` vs `main`

| Benchmark | main | This branch | Speedup |
|---|---|---|---|
| **range** 10s×10p | 16.0 µs | 11.7 µs | **1.37x** |
| **range** 10s×100p | 110 µs | 77.5 µs | **1.42x** |
| **range** 100s×100p | 986 µs | 736 µs | **1.34x** |
| **range** 1000s×100p | 9.94 ms | 7.23 ms | **1.37x** |
| **vector** 10s | 12.7 µs | 4.75 µs | **2.67x** |
| **vector** 100s | 124 µs | 46.4 µs | **2.67x** |
| **vector** 1000s | 1.34 ms | 474 µs | **2.83x** |
| **scalar** | 185 ns | 169 ns | **1.09x** |

Range (matrix) responses are **~1.3–1.4x faster** and vector responses are **~2.7x faster**. The vector improvement is larger because main serializes vectors through an intermediate `serde_json::Value`, while this branch uses direct newtype serialization with `ryu::Buffer` for zero-alloc float formatting.

timeseries/src/model.rs
```diff
@@ -396,6 +401,195 @@ impl SeriesBuilder {
...
+            .collect();
+        labels.sort();
+        Self(labels)
+    }
+}
+
+/// The result of an instant PromQL query.
+///
+/// PromQL expressions evaluate to either a scalar (e.g. `1+1`) or a
+/// vector of time series samples (e.g. `http_requests_total`).
```
cadonna: I am not sure about this. According to https://prometheus.io/docs/prometheus/latest/querying/basics/#expression-language-data-types can also have a range vector (aka matrix) as a result. 
For example, the following instant query results in a range vector:
```
prometheus_http_requests_total[5m]
``` 
You can test it here: https://demo.promlens.com/

timeseries/src/promql/tsdb_router.rs
```diff
@@ -1,37 +1,22 @@
...
-use common::clock::SystemClock;
 use promql_parser::parser::{EvalStmt, Expr, VectorSelector};
 
 /// Compute the disjoint preload ranges for bucket discovery.
 ///
 /// Uses `compute_preload_ranges` to account for `offset`/`@` modifiers.
 /// Falls back to a single `[(default_start, default_end)]` for
 /// selector-free expressions.
-fn preload_ranges(stmt: &EvalStmt, default_start: i64, default_end: i64) -> Vec<(i64, i64)> {
+pub(crate) fn preload_ranges(
```
cadonna: Why did you make these functions `pub(crate)` instead of moving them to `tsdb.rs` since they are only used there?

hachikuji:
@cadonna Thanks for the comments. I pushed an update.

Issue #191, opened 2026-02-12
binaries have nothing on stdout


agavra:
When running one of our binaries (e.g. `./opendata-timeseries --config prometheus.yaml --port 9090`) there is nothing printed to standard out. While it is working, this is a bit disconcerting from a quickstart experience.

KyleSmith19091:
I think this issue can be closed

agavra:
Yup, Thanks @KyleSmith19091 👍

PR #261, opened 2026-02-28
Change default logging level from error to info for Timeseries

KyleSmith19091:
## Summary

Changes tracing_subscriber construction to correctly set default logging level when `RUST_LOG` is not set

## Related Issues
https://github.com/opendata-oss/opendata/issues/191

# Fixes 
- Instead of just passing `EnvFilter::from_default_env()` to `env_filter` for the tracing subscriber construction which sets the the logging severity level to ERROR by default, use `try_from_default_env()` and default to info when `RUST_LOG` is not set

## Test Plan
Run the binary to see if INFO logs show up.

## Checklist

- [x] Tests added/updated


agavra (approved):
Nice catch, thanks @KyleSmith19091

PR #263, opened 2026-03-02
support batch merge and implement for vector postings

rodesai:
## Summary

- adds support for batch merge to the opendata merge operator and slatedb adapter.
- implements batch merge for vector posting lists
- convert the posting list get to a scan to force posting list mergees to be loaded concurrently

## Test Plan

- unit tests
- benchmarks show improved throughput/latency

## Checklist

- [x] Tests added/updated


vector/src/storage/merge_operator.rs
```diff
@@ -69,6 +69,20 @@ impl common::storage::MergeOperator for VectorDbMergeOperator {
...
+        let prefix =
+            KeyPrefix::from_bytes_versioned(key, KEY_VERSION).expect("Failed to decode key prefix");
+        let record_type = RecordType::from_id(prefix.tag().record_type())
+            .expect("Failed to get record type from record tag");
+
+        match record_type {
+            RecordType::PostingList => {
+                merge_batch_posting_list(existing_value, operands, self.dimensions)
+            }
+            _ => default_merge_batch(key, existing_value, operands, |k, e, v| self.merge(k, e, v)),
```
agavra, line 83: it might be a bigger change, but i'm wondering if it's worth keeping only the batch merge API for opendata systems since I don't think merge directly will be called anymore (only merge_batch will be called because we're overriding the default `merge_batch`), right?

agavra (approved):
LGTM, wondering whether we should just bite the bullet and only support `merge_batch`

vector/src/storage/merge_operator.rs
rodesai: Yeah we _should_ only see slate calling `merge_batch`. Though it's always possible it gets called from slate since its part of the trait. We could just implement `merge` by calling `merge_batch` though. Switching all the merge impls in the codebase to `merge_batch` is a larger change. I'll file an issue to track it.

rodesai: https://github.com/opendata-oss/opendata/issues/272

PR #270, opened 2026-03-02
[timeseries] Add TimeSeriesDbReader for read-only access without fencing

### LINKED ISSUE #268: Create TimeseriesDBReader trait for read-only timeseries access
## Summary

Create TimeseriesDBReader trait for read-only timeseries access

## Motivation

This will allow you to embed read-only versions of the timeseries db, useful for building benchmark or other tools, or for debugging production data. 

## Proposed Solution

The implementation will be modeled on the lines of `LogDBReader`.

apurvam:
## Summary

Add `TimeSeriesDbReader`, a read-only embedded database type that uses SlateDB's `DbReader` to coexist with a production writer without fencing. This prevents the scenario where a benchmark or analytics workload opens a full `Db` against production S3 and fences the production writer.

Design:
- Extract shared eval/discovery logic from Tsdb into pub(crate) free functions generic over `&impl QueryReader`, reused by both writer and reader paths
- `TsdbReader` (internal) wraps `StorageRead` with a moka query cache
- `TimeSeriesDbReader` (public) exposes the same read API as `TimeSeriesDb` (query, query_range, series, labels, label_values, metadata) minus write/flush
- `ReaderConfig` with configurable `refresh_interval` for manifest polling

Also:
- `range_bounds_to_secs` moved to util.rs, returns Result rejecting pre-epoch timestamps instead of silently expanding
- Zero-step guard in `evaluate_range` prevents infinite loops
- Remove unused `_storage` param from `TsdbQueryReader::new()`
- Add `MiniQueryReader::new()` constructor
- Add `query_bench` example for benchmarking reader queries
- Update RFC 0003 with reader type rationale
- 
## Related Issues

Fixes #268 
## Test Plan

Added extensive unit and integration tests. 

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/reader.rs
+++ b/timeseries/src/reader.rs
@@ -176,7 +176,7 @@ impl TimeSeriesDbReader {
 
     /// Creates a TimeSeriesDbReader from an existing storage implementation.
     pub(crate) fn from_storage(storage: Arc<dyn StorageRead>) -> Self {
-        Self::from_storage_with_capacity(storage, 50)
+        Self::from_storage_with_capacity(storage, crate::config::DEFAULT_CACHE_CAPACITY)
     }
 
     fn from_storage_with_capacity(storage: Arc<dyn StorageRead>, cache_capacity: u64) -> Self {
@@ -708,5 +708,43 @@ mod tests {
             "expected InvalidQuery, got {:?}",
             err
         );
+    }
+
+    #[test]
+    fn reader_config_deserializes_default_cache_capacity() {
+        let yaml = r#"
+storage:
+  type: InMemory
+"#;
+        let config: crate::config::ReaderConfig = serde_yaml::from_str(yaml).unwrap();
+        assert_eq!(config.cache_capacity, crate::config::DEFAULT_CACHE_CAPACITY);
+    }
+
+    #[test]
+    fn reader_config_honors_custom_cache_capacity() {
+        let yaml = r#"
+storage:
+  type: InMemory
+cache_capacity: 200
+"#;
+        let config: crate::config::ReaderConfig = serde_yaml::from_str(yaml).unwrap();
+        assert_eq!(config.cache_capacity, 200);
+    }
+
+    #[test]
+    fn from_storage_uses_default_cache_capacity() {
+        let storage = create_shared_storage();
+        let reader = TimeSeriesDbReader::from_storage(storage);
+        assert_eq!(
+            reader.query_cache.policy().max_capacity(),
+            Some(crate::config::DEFAULT_CACHE_CAPACITY)
+        );
+    }
+
+    #[test]
+    fn from_storage_with_capacity_honors_custom_value() {
+        let storage = create_shared_storage();
+        let reader = TimeSeriesDbReader::from_storage_with_capacity(storage, 123);
+        assert_eq!(reader.query_cache.policy().max_capacity(), Some(123));
     }
 }
--- a/timeseries/src/config.rs
+++ b/timeseries/src/config.rs
@@ -113,8 +113,11 @@ fn default_refresh_interval() -> Duration {
     Duration::from_secs(1)
 }
 
+/// Default number of bucket readers cached in memory.
+pub(crate) const DEFAULT_CACHE_CAPACITY: u64 = 50;
+
 fn default_cache_capacity() -> u64 {
-    50
+    DEFAULT_CACHE_CAPACITY
 }
 
 impl Default for ReaderConfig {


timeseries/src/tsdb_reader.rs
```diff
@@ -0,0 +1,286 @@
...
+        let storage = self.storage.clone();
+        self.query_cache
+            .get_with(bucket, async move {
+                Arc::new(MiniQueryReader::new(bucket, storage))
+            })
+            .await
+    }
+
+    /// Evaluate an instant PromQL query.
+    pub(crate) async fn eval_query(
```
hachikuji: I think there's a lot of duplication here with `Tsdb`. In `LogDb`, we factored out a common LogRead trait. I think it's useful to similarly add `TimeSeriesRead` as the shared trait between `TimeSeriesDb` and `TimeSeriesDbReader`. That might also suggest patterns to avoid the duplication.

apurvam: Agreed. Most of the substance is shared functions, but there is a lot of boiler plate. Maybe having a `readview` that can be leveraged across the main tsdb + the reader-only would be a better approach. Or else introduce a shared trait.

apurvam: @hachikuji I refactored the code and introduced a `TsdbEngine` trait that implements the query methods once. the `TimeseriesDbReader` and `TimeseriesDb` delegate to these methods. Let me know if this looks good, or if you have a sense of how this can be improved further.

apurvam: I think `TsdbEngine` probably is better named `TsdbReaderEngine` or something, since it handles the read heavy lifting.

timeseries/src/reader.rs
```diff
@@ -0,0 +1,747 @@
...
+
+    async fn make_query_reader(&self, start: i64, end: i64) -> Result<ReaderQueryReader> {
+        let buckets = self
+            .storage
+            .get_buckets_in_range(Some(start), Some(end))
+            .await?;
+
+        let mut readers = Vec::new();
+        for bucket in buckets {
+            let mini = self.get_or_load_bucket(bucket).await;
```
hachikuji: Might be better to load these in parallel? To do that, you can store the future in a `FuturesUnordered` for example. Similar for `make_query_reader_for_ranges`.

timeseries/src/tsdb.rs
```diff
@@ -83,6 +88,437 @@ async fn get_matching_series_multi_bucket<R: QueryReader>(
...
 
+// ── TsdbEngine trait ─────────────────────────────────────────────────
+//
+// Factors out the 6 duplicated eval/find methods shared between `Tsdb`
+// (writer) and `TimeSeriesDbReader` (reader). Each implementor provides
+// its own `QueryReader` construction; the query logic lives once in the
+// default methods.
+
+#[async_trait]
+pub(crate) trait TsdbEngine: Send + Sync {
```
hachikuji: Maybe `TsdbReadEngine` to emphasize that it is only used on the read side?

timeseries/src/reader.rs
```diff
@@ -0,0 +1,747 @@
...
+            StorageSemantics::new().with_merge_operator(Arc::new(OpenTsdbMergeOperator)),
+            reader_options,
+        )
+        .await?;
+        Ok(Self::from_storage(storage))
+    }
+
+    /// Creates a TimeSeriesDbReader from an existing storage implementation.
+    #[cfg(test)]
+    pub(crate) fn from_storage(storage: Arc<dyn StorageRead>) -> Self {
```
hachikuji, line 178: Not sure why we have two of `from_storage`.

timeseries/src/reader.rs
```diff
@@ -0,0 +1,747 @@
...
+/// let reader = TimeSeriesDbReader::open(config).await?;
+///
+/// let result = reader.query("rate(http_requests_total[5m])", None).await?;
+/// ```
+pub struct TimeSeriesDbReader {
+    storage: Arc<dyn StorageRead>,
+    /// LRU cache for read-only query buckets.
+    query_cache: Cache<TimeBucket, Arc<MiniQueryReader>>,
+    /// Metadata catalog (always empty for reader — metadata is populated during writes).
+    metadata_catalog_: RwLock<HashMap<String, Vec<MetricMetadata>>>,
```
hachikuji: Drop the trailing underscore?

timeseries/src/reader.rs
```diff
@@ -0,0 +1,747 @@
...
+        matchers: Option<&[&str]>,
+        range: impl RangeBounds<SystemTime>,
+    ) -> std::result::Result<Vec<String>, QueryError> {
+        find_label_values_in_range(self, label_name, matchers, range).await
+    }
+
+    /// Returns metric metadata, optionally filtered to a single metric.
+    ///
+    /// Note: metadata is populated from write operations and not persisted to
+    /// storage. A reader opened against existing data will return empty metadata.
```
hachikuji: Hmm, if it's always empty on the reader, perhaps it should not be exposed from TsdbEngine?

timeseries/examples/query_bench.rs
```diff
@@ -0,0 +1,205 @@
...
+                name,
+                elapsed.as_secs_f64() * 1000.0,
+                count
+            );
+
+            if let Err(e) = result {
+                eprintln!("  error: {}", e);
+            }
+        } else if *name == "label_values" {
+            let start = Instant::now();
```
hachikuji: nit: as far as I can tell, these branches look the same. Can we factor out a helper?

timeseries/src/util.rs
```diff
@@ -159,6 +159,29 @@ pub(crate) fn range_bounds_to_system_time(
     (start, end)
 }
 
+/// Convert a `RangeBounds<SystemTime>` into `(start_secs, end_secs)` as `i64`.
+///
+/// Returns an error if either bound resolves to a time before the Unix epoch.
+/// Unbounded starts resolve to 0, unbounded ends resolve to `i64::MAX`.
+pub(crate) fn range_bounds_to_secs(
```
hachikuji, line 166: The definition changed when we moved this for some reason. I am wondering if the error cases are actually possible?

common/src/coordinator/mod.rs
```diff
@@ -2743,7 +2743,7 @@ mod tests {
         let flusher = TestFlusher::default();
         let mut config = test_config();
         config.flush_size_threshold = usize::MAX;
-        config.flush_interval = Duration::from_hours(24);
+        config.flush_interval = Duration::from_secs(24 * 60 * 60);
```
hachikuji: nit: Not sure why this changed.

apurvam: I had rustc at 1.89 on my local machine. on 1.89 'from_hours` was marked as unstable. I upgraded to 1.93 (which matches our CI environment), and will revert this change.

timeseries/src/tsdb.rs
apurvam: updated.

timeseries/src/reader.rs
apurvam: Agreed. I think one was to make it available for testing. But it was unnecessary.

timeseries/src/util.rs
apurvam: Codex and Claude suggest that pre-epoch system time is possible in rust. added a couple of tests to test this edge case.

timeseries/src/reader.rs
apurvam: Updated the code to load buckets in parallel. PTAL.

timeseries/examples/query_bench.rs
apurvam: done.

timeseries/src/reader.rs
apurvam: Good call. I dropped it from `TsdbReadEngine` and moved it to Tsdb.

timeseries/src/reader.rs
```diff
@@ -0,0 +1,737 @@
...
+            StorageSemantics::new().with_merge_operator(Arc::new(OpenTsdbMergeOperator)),
+            reader_options,
+        )
+        .await?;
+        Ok(Self::from_storage(storage))
+    }
+
+    /// Creates a TimeSeriesDbReader from an existing storage implementation.
+    pub(crate) fn from_storage(storage: Arc<dyn StorageRead>) -> Self {
+        let query_cache = Cache::builder().max_capacity(50).build();
```
hachikuji: We can consider separately moving this to `ReaderConfig` perhaps?

timeseries/Cargo.toml
```diff
@@ -78,8 +80,10 @@ fail.workspace = true
 proptest.workspace = true
 rstest.workspace = true
 dashu-float.workspace = true
+tempfile = "3"
```
hachikuji, line 83: nit: we can use tempfile.workspace

timeseries/src/reader.rs
```diff
@@ -0,0 +1,737 @@
...
+        range: impl RangeBounds<SystemTime>,
+    ) -> std::result::Result<Vec<String>, QueryError> {
+        find_label_values_in_range(self, label_name, matchers, range).await
+    }
+
+    /// Returns metric metadata, optionally filtered to a single metric.
+    ///
+    /// Note: metadata is populated from write operations and not persisted to
+    /// storage. A reader opened against existing data will always return empty.
+    pub async fn metadata(
```
hachikuji: Did we mean to get rid of this?

apurvam: good catch. deleted.

timeseries/src/reader.rs
apurvam: done.

timeseries/Cargo.toml
apurvam: Yes. will touch the other modules too. Will do in a followup.

PR #274, opened 2026-03-03
[vector] Add custom Error/Result types

hachikuji:
Add vector `Error` and `Result` types (same pattern as `LogDb` and `TimeSeriesDb`).


hachikuji:
This fixes #285.

PR #275, opened 2026-03-03
Specify common source/sink abstraction for write pipelines

hachikuji:
We should support stateless ingest to timeseries. This requires introducing a sink abstraction so that it is possible to write directly to TimeSeriesDb or through the stateless ingest layer. This patch proposes to separate ingest into protocol and sink layers to provide maximum flexibility.

### DIFF
--- a/rfcs/0006-write-pipeline.md
+++ b/rfcs/0006-write-pipeline.md
@@ -1,0 +1,163 @@
+# RFC 0006: Write Pipeline (Source/Sink)
+
+**Status**: Draft
+
+**Authors**:
+- [Jason Gustafson](https://github.com/hachikuji)
+
+## Summary
+
+This RFC proposes a shared write pipeline model for OpenData systems based on `Source` and `Sink` abstractions over a generic payload type `T`.
+The model supports direct writes and staged writes with intermediate `Sink`/`Source` pairs (for example, stateless ingest), while keeping protocol handling and storage concerns decoupled.
+
+## Motivation
+
+Multiple OpenData systems need the same end-to-end write pipeline shape:
+
+- A frontend component accepts writes from external protocols.
+- A terminal writer persists data into a database.
+- Optional intermediate stages provide durability, buffering, replay, or transport.
+
+Without a shared model, each database crate redefines naming, lifecycle, and acknowledgement semantics.
+That duplication makes cross-system ingest features harder to implement and reason about.
+
+## Goals
+
+- Define a reusable source/sink pipeline model for all OpenData systems.
+- Support push and pull source semantics with explicit acknowledgement.
+- Support intermediate composable stages that are both sink and source.
+- Keep payload type generic so each database provides its own serialization and domain model.
+
+## Non-Goals
+
+- Standardize one global record schema for all databases.
+- Define protocol-specific HTTP/gRPC request parsing.
+- Define datastore-specific idempotency logic in detail.
+
+## Design
+
+### Core Abstractions
+
+The pipeline is generic over payload `T`:
+
+- `Source<T>` emits payloads.
+- `Sink<T>` accepts payloads.
+- Intermediate stages may implement both (`Sink<T> + Source<T>`).
+
+Conceptual direct pipeline:
+
+```text
+Source<T> -> Sink<T>
+```
+
+Conceptual staged pipeline:
+
+```text
+Source<T> -> StageA(Sink<T> + Source<T>) -> StageB(Sink<T> + Source<T>) -> Sink<T>
+```
+
+### System Roles in OpenData
+
+This model characterizes OpenData components into three functional roles:
+
+- **End systems**: terminal systems that persist and serve domain data.
+  - Example: TimeSeries, Vector.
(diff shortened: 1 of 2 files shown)


rfcs/0006-write-pipeline.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Shared payload type across all databases
+
+A single canonical payload for all systems could be defined.
+Rejected because data models differ significantly and would require lossy or over-generalized abstractions.
+
+## Open Questions
+
+- Should `nack` be part of the pull-source contract in addition to timeout-based redelivery?
+- Should a shared `Batch` metadata schema (ids, trace fields) be standardized in `common`?
```
cadonna: For both of these, I would wait for the requirement coming up before designing and implementing.

rfcs/0006-write-pipeline.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Design
+
+### Core Abstractions
+
+The pipeline is generic over payload `T`:
+
+- `Source<T>` emits payloads.
+- `Sink<T>` accepts payloads.
+- Intermediate stages may implement both (`Sink<T> + Source<T>`).
```
cadonna: I assume this might also be `Sink<T> + Source<U>`

rfcs/0006-write-pipeline.md
```diff
@@ -0,0 +1,163 @@
...
+
+Conceptual direct pipeline:
+
+```text
+Source<T> -> Sink<T>
```
agavra, line 50: Based on below isn't this flipped? The sink accepts writes and the source consumes them. Might be best to choose different words if that's the case. I think it's technically right but a little confusing.

hachikuji: I see the source as generating inputs for the sink to consume. Does the doc suggest otherwise somewhere? 

I'd love to hear alternative naming ideas. Source/sink confused me about half the time in Kafka connect. Maybe source is more of a generator? Sink is more of a Writer? Dunno.

agavra: 😵‍💫 I think this confused me:
```
- `StatelessIngestSink<T>` accepts `T`, serializes it, and appends durable batches.
- `StatelessIngestSource<T>` reads queued batches, deserializes to `T`, and emits downstream.
```
I guess for intermediate stages it's "flipped". I took another look and the rest aligns with the initial terminology.

agavra: I think `PullSource` is also confusing me because that feels like a Sink, I guess it would just run collocated with a Sink?

agavra: Anyway, I think it'll be clearer when we have specific implementations of these things. In the tsdb context it all makes sense.

agavra (approved):
LGTM, I'm a little caught up on the naming but in practice I think it makes sense.

rfcs/0006-write-pipeline.md
hachikuji: I think we could probably unify push/pull sources. I think pull is the more general of the two. We usually end up throwing a queue in the middle of a push api and the write logic.

cadonna (commented):
Thanks for the RFC, @hachikuji !

I had a couple of comments.

rfcs/0006-write-pipeline.md
```diff
@@ -0,0 +1,158 @@
...
+
+- Push source: receives upstream input and immediately invokes a downstream sink.
+- Pull source: exposes batches for a downstream consumer to retrieve and acknowledge.
+
+Proposed traits:
+
+```rust
+#[async_trait]
+pub trait Sink<T>: Send + Sync {
+    async fn accept(&self, payload: T) -> Result<SinkAck>;
```
cadonna, line 85: Should this interface also take optional metadata?
Is `payload` a single entry or already a batch of entries? If the former, should the interface take a batch of entries? Taking a batch would make the `Sink` API symmetric to the `Source` API. The `Source` can also adapt to the `Sink`.

rfcs/0006-write-pipeline.md
```diff
@@ -0,0 +1,158 @@
...
+}
+
+#[async_trait]
+pub trait PullSource<T>: Send + Sync {
+    async fn next_batch(&self) -> Result<Option<Batch<T>>>;
+    async fn ack(&self, batch: &Batch<T>) -> Result<()>;
+}
+```
+
+Push sources are typically adapters at protocol boundaries and can be represented as handlers that parse wire input and call `Sink<T>::accept`.
```
cadonna, line 95: Could you give an actual example for this?
I think, I understand what you mean, but we could also model that in terms of a trait. For example, a trait that takes a callback that is called when data is pushed.

cadonna (approved):
@hachikuji Thanks for the updates!

This RFC can be merged as a draft, IMO.

Some details need some discussion.

hachikuji:
@agavra @cadonna I'm sort of inclined to close this for now. I haven't merged because I couldn't quite convince myself that a framework was adding much value at this early stage. What do you think?

hachikuji:
Closed for now. We can always bring it back if we find a need.

PR #277, opened 2026-03-04
Fixup a few doc issues

hachikuji:
This patch fixes up a few stale references/inconsistencies/broken links in the docs.


hachikuji:
Shouldn't be anything controversial in here, so I'll just merge.

PR #278, opened 2026-03-04
Upgrade SlateDb to 0.11

hachikuji:
Upgrade to SlateDb 0.11.0. Fixes a few naming changes and changes to the Clock api.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,8 +41,8 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 serde_with = { version = "3", features = ["base64"] }
 serde_yaml = "0.9"
-slatedb = "0.11.0"
-slatedb-common = "0.11.0"
+slatedb = "0.11.1"
+slatedb-common = "0.11.1"
 thiserror = "2.0"
 tokio = { version = "1.0", features = ["full"] }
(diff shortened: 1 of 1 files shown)


hachikuji:
I am not sure why merge is hanging when using `MockSystemClock`. It looks like an issue in SlateDb as well. Will dig into it tomorrow...

hachikuji:
The issue in the hanging test is that we are awaiting durable without any way to advance the clock. This led to https://github.com/opendata-oss/opendata/pull/283. Once merged, I'll come back to this. We probably do not need a durable write for the test case, but I'll check whether both cases are interesting.

Cargo.toml
```diff
@@ -41,7 +41,8 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 serde_with = { version = "3", features = ["base64"] }
 serde_yaml = "0.9"
-slatedb = "0.10.1"
+slatedb = "0.11.0"
+slatedb-common = "0.11.0"
```
agavra:
```suggestion
slatedb = "0.11.1"
slatedb-common = "0.11.1"
```
there was a bad bug in 0.11.0 that I think Chris _just_ fixed and released.

Issue #250, opened 2026-02-26
[timeseries] Support functions in PromQL with more than one argument


cadonna:
## Summary

Currently, the timeseries DB only supports PromQL functions with a single argument. However, PromQL has a myriad of function with multiple arguments and with arguments of distinct types (https://prometheus.io/docs/prometheus/latest/querying/functions/). Timeseries DB should support all (or at least a good portion of) PromQL functions. To that end, the function specification within the PromQL layer needs to be generalized.

## Motivation

To support a large variety of production use cases timeseries DB needs to support a large variety of functions in PromQL. 

## Proposed Solution

The exact solution is to be determined, but the following could be a good starting point: https://github.com/prometheus/prometheus/blob/5bd0d00f8ca1ec808015b30f30b3cf1dba656518/promql/functions.go 

## Alternatives Considered

Limit PromQL in timeseries DB to single argument functions is not really an alternative if we want to support a large variety of use cases.

## Additional Context

The original Prometheus implementation in go provides context: https://github.com/prometheus/prometheus/tree/5bd0d00f8ca1ec808015b30f30b3cf1dba656518/promql

samsond:
@cadonna I think #286 addresses the design piece of this issue via RFC 0005. Any feedback is appreciated.

PR #280, opened 2026-03-04
promql: add heap-based k-aggregation and ordered promqltest assertions

### LINKED ISSUE #273: PromQL: Implement topk aggregation
Scope
* Hook AggregateOp::TopK into evaluator’s aggregation path.
* Correct grouping semantics (by() / without()).
* Output labelset matches PromQL aggregation rules for grouping.

Acceptance
* Pass promqltest cases for topk basic + grouped variants.
* Edge cases: k<=0 empty; k>group_size returns all.

samsond:
## Summary

This PR adds first-class evaluator support for topk and bottomk aggregations. It hooks AggregateOp::{TopK, BottomK} into the aggregation evaluation path with Prometheus-compatible grouping and label output semantics

TopK/BottomK result in dashboard

TopK/BottomK result in dashboard

<hr>

- implement dedicated topk/bottomk aggregation path in evaluator with deterministic heap selection and Prometheus-compatible NaN ordering
- preserve k-aggregation labels/series semantics for by/without grouping and scalar parameter coercion in aggregate evaluation
- add topk-focused evaluator tests to cover grouping and drop_name behavior
- add expect ordered assertions for order-sensitive topk/bottomk result validation
- include upstream aggregators.test fixture

## Related Issues

Fixes #273 

## Test Plan

How was this tested?

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -1837,10 +1837,11 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
             // Accept a single-element vector here so k-aggregations can consume scalar params.
             ExprResult::InstantVector(mut samples) => {
                 if samples.len() == 1 {
-                    samples
-                        .pop()
-                        .expect("single-sample vector must have one value")
-                        .value
+                    samples.pop().map(|sample| sample.value).ok_or_else(|| {
+                        EvaluationError::InternalError(
+                            "aggregation parameter expected one sample".to_string(),
+                        )
+                    })?
                 } else {
                     f64::NAN
                 }


timeseries/src/promql/evaluator.rs
```diff
@@ -1709,6 +1806,177 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
+            .await?
+        {
+            ExprResult::Scalar(value) => value,
+            // `scalar()` currently returns an instant vector in this engine.
+            // Accept a single-element vector here so k-aggregations can consume scalar params.
+            ExprResult::InstantVector(mut samples) => {
+                if samples.len() == 1 {
+                    samples
+                        .pop()
+                        .expect("single-sample vector must have one value")
```
agavra: can we return an error instead of a panic?

samsond:
> Nice, thanks @samsond this looks good to me. Also seems to me like this "fixes" #250 - I think we can close that ticket?

I don't think this fully closes #250 though. This PR fixes topk/bottomk via the aggregation/operator path (Expr::Aggregate), but #250 is about multi-arg support in the function-call path (Expr::Call). We still enforce single-arg calls in evaluate_call, so normal multi-argument PromQL functions are still blocked: https://github.com/opendata-oss/opendata/blob/main/timeseries/src/promql/evaluator.rs#L1370

timeseries/src/promql/evaluator.rs
samsond: Makes sense. Fixed.

PR #281, opened 2026-03-04
[timeseries] Support one-to-many & many-to-one vector matching

### LINKED ISSUE #85: [tsdb] support many-to-one and one-to-many match cardinality for vector matching
## Summary

Support many-to-one and one-to-many match cardinality for vector match in PromQL expressions using `group_left/group_right`. See the PromQL query language documentation for more details on expected behavior: https://prometheus.io/docs/prometheus/latest/querying/operators/#vector-matching

## Motivation

Full PromQL support

## Proposed Solution

We should implement the matching logic in `Evaluator#evaluate_binary_expr`

micllam:
## Summary

Adds support for one-to-many and many-to-one vector matching.

I modified the existing one-to-one matching logic to cover these other two cases.

## Related Issues

Fixes #85.

## Test Plan

Added unit tests.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -1,4 +1,5 @@
-use std::collections::{HashMap, HashSet};
+use std::cmp::Ordering;
+use std::collections::{BinaryHeap, HashMap, HashSet};
 use std::fmt::{Display, Formatter};
 use std::future::Future;
 use std::pin::Pin;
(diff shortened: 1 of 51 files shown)


timeseries/src/promql/evaluator.rs
```diff
@@ -1562,49 +1563,84 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
                     }
                 }
 
-                // Build right-side index keyed by match signature
-                let mut right_map: HashMap<Vec<(String, String)>, EvalSample> = HashMap::new();
-                for sample in right_vector {
+                let is_group_right = matches!(card, VectorMatchCardinality::OneToMany(_));
+                let is_one_to_one = matches!(card, VectorMatchCardinality::OneToOne);
+                let group_labels: Option<&Vec<String>> = match card {
```
hachikuji: nit: could we use `card.labels()` directly?

timeseries/src/promql/evaluator.rs
```diff
@@ -1562,49 +1563,84 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
+                    let key = Self::compute_binary_match_key(&many_sample.labels, matching);
 
-                    // Look up matching right sample
-                    let right_sample = match right_map.get(&key) {
-                        Some(rs) => rs,
-                        None => continue, // Unmatched left samples silently dropped
+                    // Look up matching "one" sample
+                    let one_sample = match one_map.get(&key) {
+                        Some(s) => s,
+                        None => continue, // Unmatched "one" samples silently dropped
```
hachikuji: nit: the comment seems misleading. Aren't we dropping the many-side sample?

timeseries/src/promql/evaluator.rs
```diff
@@ -1562,49 +1563,84 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
+                let is_one_to_one = matches!(card, VectorMatchCardinality::OneToOne);
+                let group_labels: Option<&Vec<String>> = match card {
+                    VectorMatchCardinality::ManyToOne(ll)
+                    | VectorMatchCardinality::OneToMany(ll) => Some(&ll.labels),
+                    _ => None,
+                };
+
+                // Determine which side is "one" vs "many" for matching purposes.
+                // For one-to-one mappings, we treat the right-hand side as the "one" side.
+                let (one_vec, one_side_name, many_vec, many_side_name) = if is_group_right {
```
hachikuji: nit: lots of variables to track. Wonder if we could just inline the name assignment in the error case below?

timeseries/src/promql/evaluator.rs
micllam: Done. Didn't spot that method before.

timeseries/src/promql/evaluator.rs
micllam: Modified the comment slightly. Happy to make further changes.

timeseries/src/promql/evaluator.rs
micllam: Done.

micllam:
Thanks for your review.

I've added some tests to promqltest. I noticed that `assert_results` doesn't check if we omitted any labels from the expected results, which could be useful to check for these new tests.

timeseries/src/promql/evaluator.rs
```diff
@@ -1659,49 +1563,82 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
-                    // One-to-one check: only error on duplicate left keys that have a right match
-                    if !left_seen.insert(key) {
-                        return Err(EvaluationError::InternalError(
-                            "many-to-many matching not allowed: found duplicate series on the left side of the operation".to_string(),
-                        ));
+                    // One-to-one check: only error on duplicate "many" keys that have a "one" match
+                    if is_one_to_one && !one_to_one_seen.insert(key) {
+                        return Err(EvaluationError::InternalError(format!(
+                            "many-to-many matching not allowed: found duplicate series on the {} side of the operation",
+                            if is_group_right { "right" } else { "left" }
```
micllam: As this check is only performed in the one-to-one case, perhaps I should just use "left" directly without checking `is_group_right`?

micllam: Sorry for the extra noise, but I decided to remove the check on `is_group_right` here.

PR #283, opened 2026-03-04
Fix inconsistent await_durable defaults across Storage write methods

### LINKED ISSUE #282: Inconsistent await_durable defaults between put, merge, and apply
## Summary

The `Storage` trait has inconsistent durability behavior across write methods due to a mismatch between opendata's `WriteOptions::default()` and SlateDB's `WriteOptions::default()`.

- **opendata** `common::WriteOptions` derives `Default`, giving `await_durable: false`
- **SlateDB** `WriteOptions::default()` sets `await_durable: true`

## Current Behavior

| Method | SlateDB call | Effective `await_durable` |
|--------|-------------|--------------------------|
| `put()` | `write_with_options(batch, {await_durable: false})` | **false** |
| `put_with_options(..., default)` | same | **false** |
| `merge()` | `db.write(batch)` → uses SlateDB default | **true** |
| `apply()` | `db.write(batch)` → uses SlateDB default | **true** |

`merge()` and `apply()` call `db.write(batch)`, which delegates to `write_with_options` with SlateDB's default (`await_durable: true`). Meanwhile, `put` explicitly passes opendata's default (`await_durable: false`).

## Impact

In production, merges flow through `apply()` via `RecordOp::Merge`, used by:
- **vector/** — posting list updates, deletions, centroid stats, centroid chunks
- **timeseries/** — inverted index, time series values

These are all silently awaiting durability on every write, while `put`-based paths (log writer, keyvalue) are not. This is likely unintentional and may be causing unnecessary latency in vector and timeseries writes.

## Relevant Code

- `common/src/storage/mod.rs` — `WriteOptions` struct with `#[derive(Default)]` (`await_durable: false`)
- `common/src/storage/slate.rs:260-274` — `merge()` uses `db.write(batch)` (no options)
- `common/src/storage/slate.rs:211-228` — `apply()` uses `db.write(batch)` (no options)
- `common/src/storage/slate.rs:236-252` — `put_with_options()` uses `write_with_options` with explicit options

## Suggested Fix

Ensure `merge()` and `apply()` also pass `WriteOptions` through to SlateDB (similar to `put_with_options`), so callers can control durability consistently. This likely means:
1. Adding `_with_options` variants for `merge` and `apply`, or adding `WriteOptions` to the existing signatures
2. Making the default behavior consistent across all write paths

hachikuji:
## Summary

Fixes #282.

- **Storage trait**: `apply()` and `merge()` now default to `await_durable: false` (matching `put()`), delegating to new `apply_with_options` / `merge_with_options` methods. Previously, the SlateDB impl called `db.write(batch)` which silently defaulted to `await_durable: true`.
- **Timeseries flush**: `MiniTsdb::flush()` was not flushing to durable storage (`flush_storage: false`). `Tsdb::flush()` now flushes all bucket deltas in parallel via `FuturesUnordered`, then issues a single `storage.flush()`.
- **Vector flush**: `VectorDb::flush()` now passes `flush_storage: true` and awaits `Durability::Durable`.
- **Tests**: Added 6 SlateDB integration tests verifying `await_durable` behavior via `DbReader` visibility, plus a vector durability test that reopens the database from disk.

## Test plan

- [x] `cargo test --all --all-features` — all passing
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — clean
- [x] `cargo doc --all --no-deps --all-features` — clean

### DIFF
--- a/common/src/storage/mod.rs
+++ b/common/src/storage/mod.rs
@@ -208,6 +208,9 @@ pub fn default_merge_batch(
 /// Iterator over storage records.
 #[async_trait]
 pub trait StorageIterator {
+    /// Returns the next record from this iterator.
+    ///
+    /// Returns `Ok(None)` when the iterator is exhausted.
     async fn next(&mut self) -> StorageResult<Option<Record>>;
 }
 
@@ -218,6 +221,9 @@ pub trait StorageIterator {
 /// we can write code that works with both storage types.
 #[async_trait]
 pub trait StorageRead: Send + Sync {
+    /// Retrieves a single record by exact key.
+    ///
+    /// Returns `Ok(None)` if the key is not present.
     async fn get(&self, key: Bytes) -> StorageResult<Option<Record>>;
 
     /// Returns an iterator over records in the given range.
@@ -264,12 +270,22 @@ pub trait Storage: StorageRead {
     }
 
     /// Applies a batch of mixed operations with custom write options.
+    ///
+    /// `options.await_durable` controls whether this call returns only after
+    /// the batch reaches durable storage.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the batch cannot be applied.
     async fn apply_with_options(
         &self,
         ops: Vec<RecordOp>,
         options: WriteOptions,
     ) -> StorageResult<()>;
 
+    /// Writes records to storage.
+    ///
+    /// Uses `WriteOptions::default()` (`await_durable: false`).
     async fn put(&self, records: Vec<PutRecordOp>) -> StorageResult<()> {
         self.put_with_options(records, WriteOptions::default())
             .await
@@ -284,6 +300,10 @@ pub trait Storage: StorageRead {
     ///
     /// * `records` - The records to write
     /// * `options` - Write options controlling durability behavior
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if any write in the batch fails.
     async fn put_with_options(
         &self,
         records: Vec<PutRecordOp>,
@@ -306,6 +326,14 @@ pub trait Storage: StorageRead {
     }
 
     /// Merges values with custom write options.
+    ///
+    /// `options.await_durable` controls whether this call returns only after
+    /// the batch reaches durable storage.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if no merge operator is configured or if the merge
+    /// operation fails.
     async fn merge_with_options(
         &self,
         records: Vec<MergeRecordOp>,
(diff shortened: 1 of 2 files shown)


common/src/storage/mod.rs
```diff
@@ -284,7 +298,19 @@ pub trait Storage: StorageRead {
...
+    async fn merge(&self, records: Vec<MergeRecordOp>) -> StorageResult<()> {
+        self.merge_with_options(records, WriteOptions::default())
+            .await
+    }
+
+    /// Merges values with custom write options.
+    async fn merge_with_options(
+        &self,
+        records: Vec<MergeRecordOp>,
+        options: WriteOptions,
```
apurvam, line 340: nit: options sometimes has a leading _, and sometimes not. would be good to standardize on one or the other.

common/src/storage/slate.rs
```diff
@@ -252,25 +254,31 @@ impl Storage for SlateDbStorage {
         Ok(())
     }
 
-    /// Merges values for the given keys using SlateDB's merge operator.
```
apurvam, line 255: did you mean to delete the comment. Seems it's valid for the `with_options` variant too.

timeseries/src/tsdb.rs
```diff
@@ -621,12 +622,20 @@ impl Tsdb {
         Ok(TsdbQueryReader::new(readers))
     }
 
-    /// Flush all dirty buckets in the ingest cache to storage.
+    /// Flush all dirty buckets to durable storage.
```
apurvam, line 625: should we add a test to check that flush now works as intended?

log/src/server/handlers.rs
```diff
@@ -350,11 +350,11 @@ mod tests {
 
         #[async_trait]
         impl Storage for ConfigurableStorage {
-            async fn apply(&self, _ops: Vec<RecordOp>) -> common::StorageResult<()> {
-                self.check_failure()
-            }
-
-            async fn put(&self, _records: Vec<PutRecordOp>) -> common::StorageResult<()> {
+            async fn apply_with_options(
```
apurvam, line 353: It looks like the non `_options` methods weren't being called at all (no call sites were updated in the patch). Do we even need these variants?

apurvam (commented):
Left a few comments, but it LGTM for most part.

common/src/storage/mod.rs
hachikuji: This is a rust convention. The underscore is added to explicitly mark an argument as unused.

timeseries/src/tsdb.rs
hachikuji: We had one test that already surfaced the bug, but it was a little indirect. Let me add one that is more explicit.

log/src/server/handlers.rs
hachikuji: The patch doesn't remove any of the methods, so there should not be any updates for callers. The `put` here is not removed, but just moved into the `Storage` trait.

common/src/storage/slate.rs
hachikuji: I think convention is to document the trait, not the implementation. Let me do a pass on the trait docs though since there are a couple minor gaps.

hachikuji:
@apurvam Thanks for the comments. I pushed an update to add an explicit flush test for timeseries and to address gaps in the docs for `Storage`.

log/src/server/handlers.rs
apurvam: ah that makes sense.

apurvam (approved):
LGTM. thanks for the patch!

Issue #94, opened 2026-01-17
[tsdb] support the Prometheus federate API


rodesai:
We should support the Prometheus `federate` API from our timeseries server. See https://prometheus.io/docs/prometheus/latest/federation/ for more details.

ryandielhenn:
Interested, looking into this.

PR #284, opened 2026-03-04
Support the `/federate` Prometheus API for timeseries

ryandielhenn:
## Summary

This PR adds support to the timeseries server for the Prometheus federate API.

## Related Issues

#94 

## Test Plan

Tests that responses contain the correct series for a given selector, that multiple overlapping selectors return the union with no duplicates, that label values are properly escaped in the exposition format output, that series with unsorted labels are correctly deduplicated, that nonexistent metrics return an empty response, and that missing match[] parameters return a proper error.

This was also tested by spinning up the `opendata-timeseries` quickstart, ingesting data into the tsdb using a python script, and checking the endpoint at `localhost:9090/federate` with a few match parameters e.g. `http://localhost:9090/federate?match[]=up`

## References
https://prometheus.io/docs/prometheus/latest/federation/
https://prometheus.io/docs/instrumenting/exposition_formats/

## Checklist

- [X] Tests added/updated

### DIFF
--- a/timeseries/src/server/http.rs
+++ b/timeseries/src/server/http.rs
@@ -383,7 +383,12 @@ async fn handle_federate(
 
         let samples = match result {
             QueryValue::Vector(s) => s,
-            _ => continue,
+            _ => {
+                return Err(Error::InvalidInput(format!(
+                    "match[] must be a vector selector, got: {selector}"
+                ))
+                .into());
+            }
         };
 
         for sample in samples {


timeseries/tests/http_server.rs
```diff
@@ -447,6 +460,170 @@ async fn test_metadata() {
...
+
+    let req = Request::get("/federate?match[]=test_metric")
+        .body(Body::empty())
+        .unwrap();
+
+    let resp = app.oneshot(req).await.unwrap();
+
+    assert_eq!(resp.status(), StatusCode::OK);
+    let body = body_string(resp).await;
+    assert!(
```
apurvam, line 554: I have a question: is the federate response supposed to be in a schema / format that's different from other end points? I looked up the [federate docs](https://prometheus.io/docs/prometheus/latest/federation/) but it doesn't specify the response format. Your code seems to implement a specific structure, so I assume there is a spec. If so, we should definitely test that the responses match the spec.

ryandielhenn: I assumed the text format specified here for the response because it seems to be the default - https://prometheus.io/docs/instrumenting/exposition_formats/.

I think you're right to double check this. I will look into it more.

apurvam: ok. But then that would be the same output as the `/metrics` endpoint right? Could we use the same code to construct the response? 

The [docs](https://prometheus.io/docs/prometheus/latest/federation/#configuring-federation) indicate that there needs to be an additional config layer for `/federate`. Basically, a 'destination' federates from 'sources' by scraping particular selectors on the sources, collecting the responses, and forwarding them. 

I'm not sure what the expectation for checks on the source side (which you are implementing) needs to be. It would be worth making sure that the source is respecting its config as part of this patch. 

If I'm off base here, let me know. I only took a cursory look at the docs.

ryandielhenn: The difference is in that `/federate` is giving actual data that was ingested to the tsdb whereas `/metrics` is giving metrics about the server process. You're right that the output format is the same, but internally the data sources and types are different. 

Digging deeper into your point about being able to reuse the code from `/metrics`, 
- `prometheus_client::encoding::text::encode(&mut buffer, &self.registry)` accepts `prometheus_client::registry::Registry`. 
- `/federate` is supposed to return the latest samples based on the union of series found by the match[] selectors from the query. The data sources are fundamentally different.

ryandielhenn: > The [docs](https://prometheus.io/docs/prometheus/latest/federation/#configuring-federation) indicate that there needs to be an additional config layer for /federate. Basically, a 'destination' federates from 'sources' by scraping particular selectors on the sources, collecting the responses, and forwarding them.
> I'm not sure what the expectation for checks on the source side (which you are implementing) needs to be. It would be worth making sure that the source is respecting its config as part of this patch.

IIUC that config is for the destination side, the source just answers whatever match[] selectors come in.

ryandielhenn: I did a bit more digging into the output format. While PrometheusText 0.0.4 is the standard baseline, Prometheus actually uses HTTP content negotiation to support OpenMetrics and Protobuf as well (docs here https://prometheus.io/docs/instrumenting/content_negotiation/).

The logic follows a hierarchy: it prioritizes the Accept header, then a user-configured fallback, and finally defaults to 0.0.4 (what's used in this PR) if nothing else is specified.

Would you prefer we stick to the baseline 0.0.4 default for this PR and create an issue for adding support for other output formats based on the Accept header? Alternatively do you want the implementation in this PR to explicitly support negotiation for OpenMetrics/Protobuf?

apurvam: I think PrometheusText 0.0.4 is enough for now. If there is demand for more, we can add it. 

@samsond if this change looks good to you, I'm happy to merge.. you are more enmeshed in PromQL than me.

samsond: Sure @apurvam - Let me quickly review it.

timeseries/src/server/http.rs
```diff
@@ -339,6 +352,106 @@ async fn handle_metadata(
...
+            .await
+            .map_err(|e| match e {
+                crate::error::QueryError::InvalidQuery(msg) => Error::InvalidInput(msg),
+                crate::error::QueryError::Execution(msg) => Error::Internal(msg),
+                crate::error::QueryError::Timeout => Error::Internal("query timed out".to_string()),
+            })?;
+
+        let samples = match result {
+            QueryValue::Vector(s) => s,
+            _ => continue,
```
samsond: If IIUC this `_ => continue` silently accepts invalid `match[]` values for `/federate`. Can we return maybe `Error::InvalidInput` (HTTP 400) when a `match[]` is not a vector selector / does not evaluate to `QueryValue::Vector`? Ref: https://github.com/prometheus/prometheus/blob/main/web/federate.go#L66-L70

samsond (commented):
Overall LGTM, Thanks.

Left one review inline.

samsond (approved):

timeseries/src/server/http.rs
ryandielhenn: Nice catch.

PR #286, opened 2026-03-04
rfc(timeseries): add RFC 0005 for generalized PromQL function signatures

samsond:
## Summary

This PR adds RFC 0005: Generalized PromQL Function Signatures to unblock multi-argument, mixed-type, optional/variadic, and scalar-returning PromQL functions in opendata-timeseries.

Today, function execution is effectively constrained to single-argument calls and special-cased function categories, which prevents supporting a large portion of valid PromQL functions (e.g., round(v[, to_nearest]), clamp(v,min,max), label_replace(...), label_join(...), time(), pi()). This RFC proposes a unified function argument model + dispatch interface so we can implement these functions incrementally while staying Prometheus-compatible.

## Related Issues

Also see #250

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/timeseries/src/promql/functions.rs
+++ b/timeseries/src/promql/functions.rs
@@ -259,13 +259,21 @@ fn max_with_nan(left: f64, right: f64) -> f64 {
     }
 }
 
+fn exact_arity_error(
+    function_name: &str,
+    expected_args: usize,
+    actual_args: usize,
+) -> EvaluationError {
+    EvaluationError::InternalError(format!(
+        "{function_name} requires exactly {expected_args} argument(s), got {actual_args}"
+    ))
+}
+
 struct ClampMaxFunction;
 
 impl PromQLFunction for ClampMaxFunction {
     fn apply(&self, _arg: PromQLArg, _eval_timestamp_ms: i64) -> EvalResult<Vec<EvalSample>> {
-        Err(EvaluationError::InternalError(
-            "clamp_max requires two arguments".to_string(),
-        ))
+        Err(exact_arity_error("clamp_max", 2, 1))
     }
 
     fn apply_args(
@@ -273,22 +281,13 @@ impl PromQLFunction for ClampMaxFunction {
         args: Vec<PromQLArg>,
         _eval_timestamp_ms: i64,
     ) -> EvalResult<Vec<EvalSample>> {
-        let mut args_iter = args.into_iter();
-        let Some(first_arg) = args_iter.next() else {
-            return Err(EvaluationError::InternalError(
-                "clamp_max requires two arguments".to_string(),
-            ));
-        };
-        let Some(second_arg) = args_iter.next() else {
-            return Err(EvaluationError::InternalError(
-                "clamp_max requires two arguments".to_string(),
-            ));
-        };
-        if args_iter.next().is_some() {
-            return Err(EvaluationError::InternalError(
-                "clamp_max accepts at most two arguments".to_string(),
-            ));
+        if args.len() != 2 {
+            return Err(exact_arity_error("clamp_max", 2, args.len()));
         }
+
+        let mut args_iter = args.into_iter();
+        let first_arg = args_iter.next().expect("validated args.len() == 2");
+        let second_arg = args_iter.next().expect("validated args.len() == 2");
 
         let mut samples = first_arg.into_instant_vector()?;
         let max = second_arg.into_scalar()?;
@@ -303,9 +302,7 @@ struct ClampMinFunction;
 
 impl PromQLFunction for ClampMinFunction {
     fn apply(&self, _arg: PromQLArg, _eval_timestamp_ms: i64) -> EvalResult<Vec<EvalSample>> {
-        Err(EvaluationError::InternalError(
-            "clamp_min requires two arguments".to_string(),
-        ))
+        Err(exact_arity_error("clamp_min", 2, 1))
     }
 
     fn apply_args(
@@ -313,22 +310,13 @@ impl PromQLFunction for ClampMinFunction {
         args: Vec<PromQLArg>,
         _eval_timestamp_ms: i64,
     ) -> EvalResult<Vec<EvalSample>> {
-        let mut args_iter = args.into_iter();
-        let Some(first_arg) = args_iter.next() else {
-            return Err(EvaluationError::InternalError(
-                "clamp_min requires two arguments".to_string(),
-            ));
-        };
-        let Some(second_arg) = args_iter.next() else {
-            return Err(EvaluationError::InternalError(
-                "clamp_min requires two arguments".to_string(),
-            ));
-        };
-        if args_iter.next().is_some() {
-            return Err(EvaluationError::InternalError(
-                "clamp_min accepts at most two arguments".to_string(),
-            ));
+        if args.len() != 2 {
+            return Err(exact_arity_error("clamp_min", 2, args.len()));
         }
+
+        let mut args_iter = args.into_iter();
+        let first_arg = args_iter.next().expect("validated args.len() == 2");
+        let second_arg = args_iter.next().expect("validated args.len() == 2");
 
         let mut samples = first_arg.into_instant_vector()?;
         let min = second_arg.into_scalar()?;
@@ -343,9 +331,7 @@ struct ClampFunction;
 
 impl PromQLFunction for ClampFunction {
     fn apply(&self, _arg: PromQLArg, _eval_timestamp_ms: i64) -> EvalResult<Vec<EvalSample>> {
-        Err(EvaluationError::InternalError(
-            "clamp requires three arguments".to_string(),
-        ))
+        Err(exact_arity_error("clamp", 3, 1))
     }
 
     fn apply_args(
@@ -353,27 +339,14 @@ impl PromQLFunction for ClampFunction {
         args: Vec<PromQLArg>,
         _eval_timestamp_ms: i64,
     ) -> EvalResult<Vec<EvalSample>> {
-        let mut args_iter = args.into_iter();
-        let Some(first_arg) = args_iter.next() else {
-            return Err(EvaluationError::InternalError(
-                "clamp requires three arguments".to_string(),
-            ));
-        };
-        let Some(second_arg) = args_iter.next() else {
-            return Err(EvaluationError::InternalError(
-                "clamp requires three arguments".to_string(),
-            ));
-        };
-        let Some(third_arg) = args_iter.next() else {
-            return Err(EvaluationError::InternalError(
-                "clamp requires three arguments".to_string(),
-            ));
-        };
-        if args_iter.next().is_some() {
-            return Err(EvaluationError::InternalError(
-                "clamp accepts at most three arguments".to_string(),
-            ));
+        if args.len() != 3 {
+            return Err(exact_arity_error("clamp", 3, args.len()));
         }
+
+        let mut args_iter = args.into_iter();
+        let first_arg = args_iter.next().expect("validated args.len() == 3");
+        let second_arg = args_iter.next().expect("validated args.len() == 3");
+        let third_arg = args_iter.next().expect("validated args.len() == 3");
 
         let mut samples = first_arg.into_instant_vector()?;
         let min = second_arg.into_scalar()?;
@@ -1004,6 +977,100 @@ mod tests {
 
         assert!(
             err.to_string().contains("expected instant vector"),
+            "unexpected error: {}",
+            err
+        );
+    }
+
+    #[test]
+    fn should_include_actual_argument_count_in_clamp_errors() {
+        let registry = FunctionRegistry::new();
+        let clamp = registry.get("clamp").unwrap();
+
+        let err = clamp.apply_args(vec![], 1000).unwrap_err();
+        assert!(
+            err.to_string()
+                .contains("clamp requires exactly 3 argument(s), got 0"),
+            "unexpected error: {}",
+            err
+        );
+
+        let err = clamp
+            .apply_args(
+                vec![
+                    PromQLArg::InstantVector(vec![create_sample(1.0)]),
+                    PromQLArg::Scalar(0.0),
+                    PromQLArg::Scalar(1.0),
+                    PromQLArg::Scalar(2.0),
+                ],
+                1000,
+            )
+            .unwrap_err();
+        assert!(
+            err.to_string()
+                .contains("clamp requires exactly 3 argument(s), got 4"),
+            "unexpected error: {}",
+            err
+        );
+    }
+
+    #[test]
+    fn should_include_actual_argument_count_in_clamp_min_errors() {
+        let registry = FunctionRegistry::new();
+        let clamp_min = registry.get("clamp_min").unwrap();
+
+        let err = clamp_min.apply_args(vec![], 1000).unwrap_err();
+        assert!(
+            err.to_string()
+                .contains("clamp_min requires exactly 2 argument(s), got 0"),
+            "unexpected error: {}",
+            err
+        );
+
+        let err = clamp_min
+            .apply_args(
+                vec![
+                    PromQLArg::InstantVector(vec![create_sample(1.0)]),
+                    PromQLArg::Scalar(0.0),
+                    PromQLArg::Scalar(1.0),
+                ],
+                1000,
+            )
+            .unwrap_err();
+        assert!(
+            err.to_string()
+                .contains("clamp_min requires exactly 2 argument(s), got 3"),
+            "unexpected error: {}",
+            err
+        );
+    }
+
+    #[test]
+    fn should_include_actual_argument_count_in_clamp_max_errors() {
+        let registry = FunctionRegistry::new();
+        let clamp_max = registry.get("clamp_max").unwrap();
+
+        let err = clamp_max.apply_args(vec![], 1000).unwrap_err();
+        assert!(
+            err.to_string()
+                .contains("clamp_max requires exactly 2 argument(s), got 0"),
+            "unexpected error: {}",
+            err
+        );
+
+        let err = clamp_max
+            .apply_args(
+                vec![
+                    PromQLArg::InstantVector(vec![create_sample(1.0)]),
+                    PromQLArg::Scalar(0.0),
+                    PromQLArg::Scalar(1.0),
+                ],
+                1000,
+            )
+            .unwrap_err();
+        assert!(
+            err.to_string()
+                .contains("clamp_max requires exactly 2 argument(s), got 3"),
             "unexpected error: {}",
             err
         );


timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,398 @@
...
+Module responsibilities today:
+
+- `functions.rs`: function trait definitions, function registry, and
+  implementation bodies.
+- `evaluator.rs`: function-call orchestration (arity gate, argument evaluation,
+  dispatch, and result wrapping).
+
+After this RFC:
+
+- `functions.rs`: generalized function interfaces, function implementations, and
```
agavra, line 88: nice, I like moving the function implementations out of the evaluator.

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,398 @@
...
+- Prometheus promql package docs (`FunctionCalls`, `FunctionCall`):
+  <https://pkg.go.dev/github.com/prometheus/prometheus/promql>
+- Prometheus function docs (default/variadic argument behavior):
+  <https://prometheus.io/docs/prometheus/latest/querying/functions>
+
+### Alignment with Rust `promql-parser` (0.6.x)
+
+The parser used in this crate (`promql-parser = "0.6"`) is mostly aligned with
+Prometheus at signature/type-check level, with one important representation
+difference:
```
agavra: what's the one important difference? or are these all differences?

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,398 @@
...
+- Type-checking for variadic overflow reuses the final declared arg type.
+
+Implication for this RFC:
+
+- AST metadata on `Call.func` is reliable for baseline type expectations.
+- Rust parser `variadic: bool` is not rich enough to represent Prometheus
+  `Variadic int` semantics exactly.
+- Evaluator must not encode arity behavior through scattered function-name
+  exceptions.
+- Add a Prometheus-derived internal signature layer for arity semantics and a
```
agavra: I can't say I understand what you mean by a "internal signature layer for arity semantics" but I think it might be OK for us at the moment to not support variadic functions if we need to do some hacky stuff to fix it. I'm sure the folk at greptime would appreciate a contribution to `promql-parser` if we wanted to implement this at the parser layer. we can start by opening a ticket there

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,398 @@
...
+        &self,
+        evaluated_args: Vec<Option<FunctionArgValue>>,
+        ctx: &FunctionCallContext<'_>,
+    ) -> EvalResult<ExprResult>;
+}
+```
+
+Notes:
+
+- Return type becomes `ExprResult` (not always instant vector).
```
agavra: I wonder if it makes sense to have different `PromQLFunction` traits depending on what the return type is. Is it helpful to make them all generic in the sense that there are places where you could use them interchangably? I imagine most places that require a vector don't let you pass in a scalar (for example)

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,398 @@
...
+- String arguments are read from `ctx.raw_args`; corresponding
+  `evaluated_args[idx]` is `None` (Prometheus-compatible shape).
+- Existing one-arg function implementations remain usable through compatibility
+  adapters during migration.
```
agavra: what does this mean? we shouldn't need any adapters since this is all internal, we should support things natively

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,398 @@
...
+### Signature Source of Truth
+
+We use two signature sources, each for a different job:
+
+- Parser metadata on `Call.func` (`arg_types`, `return_type`) for type checking:
+  argument type expectations and return-type expectations.
+- Internal Prometheus-derived `FunctionSignature` table for arity semantics:
+  exact arity, bounded optional arity, and unbounded variadic arity.
+
+Why two layers:
```
agavra: see above, I think we should just contribute this back to promql-parser, I suspect the variadics are not as commonly used

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,398 @@
...
+For initial rollout, implement the subset needed by first target functions:
+
+- Optional second arg: `round(v, to_nearest=1)`
+- Unbounded variadic strings: `label_join(...src_labels)`
+- Optional instant-vector defaults: time-component functions (`year`, `month`,
+  `day_of_month`, etc.) in later phase.
+- Default argument filling remains inside function implementations (for example:
+  `round`, date/time helpers), matching Prometheus behavior.
+
+### Evaluator Call Flow
```
agavra: I appreciate the detail put into this RFC, but this section and some of the others are probably more details than we need in an RFC, anything that is better just looked at as code should probably remain code only so that we can focus only on the higher level architectural decisions in the RFC.

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,398 @@
...
+
+Replace split registries with one registry keyed by function name:
+
+```rust
+pub(crate) struct FunctionRegistry {
+    functions: HashMap<String, Box<dyn PromQLFunction>>,
+}
+```
+
+### Migration Plan (Low Technical Debt)
```
agavra: I think we can remove this since it's all internal, there's no external facing compatibility concerns

timeseries/rfcs/0005-promql-function-signatures.md
samsond: The primary difference I meant is the variadic representation: Prometheus uses Variadic int (0 exact, >0 bounded optional args, <0 unbounded), while promql-parser 0.6 exposes only variadic: bool.

There is also a secondary metadata gap: the Rust Function metadata does not carry Prometheus’ Experimental flag.

I’ll reword the RFC section to say two key differences instead of "one important difference" so this is explicit.

timeseries/rfcs/0005-promql-function-signatures.md
samsond: By "internal signature layer for arity semantics," I meant a small evaluator-side FunctionSignature map keyed by function name, so we can model Prometheus arity rules (variadic as int) without scattering name-based checks. It is not intended as a long-term fork of parser behavior.

I’m good with reducing scope for now: we can land fixed-arity and non-hacky multi-arg support first, and defer unbounded variadics (for example label_join) until parser support improves.

I’ll open an upstream promql-parser ticket for richer function metadata (Variadic cardinality and Experimental), and we can align/remove local bridging once that lands.

timeseries/rfcs/0005-promql-function-signatures.md
samsond: Agreed. I removed the internal signature-layer plan from this RFC.

This phase now uses promql-parser metadata as the runtime source of truth and limits support to parser-supported arity/variadics.

Prometheus-exact variadic cardinality and Experimental metadata are deferred to upstream promql-parser, and we’ll track that via an upstream issue/PR.

timeseries/rfcs/0005-promql-function-signatures.md
samsond: Agree, good feedback.

timeseries/rfcs/0005-promql-function-signatures.md
samsond: I opened an upstream `promql-parser` issue to track richer variadic signature metadata: [promql-parser-issue-link](https://github.com/GreptimeTeam/promql-parser/issues/129). For this PR, we intentionally support fixed and bounded optional arity, and defer unbounded variadic tails until upstream metadata is available.

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,342 @@
...
+
+- Introduce typed argument values and call context for unified dispatch,
+  especially for string-argument functions.
+
+```rust
+pub(crate) enum FunctionArgValue {
+    Scalar(f64),
+    InstantVector(Vec<EvalSample>),
+    RangeVector(Vec<EvalSamples>),
+}
```
cadonna: Why can we not re-use `ExprResult` here?
It is exactly the same enum.

samsond: Great point. I agree we should reuse existing types here.

`FunctionArgValue` was introduced for role separation (function input args vs evaluator result), not because we needed a different data shape. Since it currently matches `ExprResult`, we can simplify the RFC/code path by reusing `ExprResult` for evaluated non-string args, and keep string literals in `raw_args` with `None` slots.

That avoids a duplicate enum and keeps Phase 2 focused on call-context and dispatch unification.

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,336 @@
...
+
+    fn apply_args(
+        &self,
+        args: Vec<PromQLArg>,
+        eval_timestamp_ms: i64,
+    ) -> EvalResult<Vec<EvalSample>>;
+}
+```
+
+Phase 2 (target):
```
agavra: curious why we're splitting the RFC into  two phases? I think an RFC should document the target state, if the implementation gets unweildly we can split PRs that implement them into two.

To me, the point of an RFC is mostly to serve as documentation of decisions. If there are implementation specific details that are challenging, we can call them out in brief.

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,336 @@
...
+References:
+
+- Prometheus parser package docs (function signatures and metadata):
+  <https://pkg.go.dev/github.com/prometheus/prometheus/promql/parser>
+- Prometheus promql package docs (`FunctionCalls`, `FunctionCall`):
+  <https://pkg.go.dev/github.com/prometheus/prometheus/promql>
+- Prometheus function docs (default/variadic argument behavior):
+  <https://prometheus.io/docs/prometheus/latest/querying/functions>
+
+### Alignment with Rust `promql-parser` (0.6.x)
```
agavra: nit: probably worth referencing the ticket you submitted here

timeseries/src/promql/evaluator.rs
```diff
@@ -1455,6 +1455,74 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
         hasher.finish() as SeriesFingerprint
     }
 
+    fn call_arity_bounds(call: &Call) -> (usize, usize) {
+        let expected_args_len = call.func.arg_types.len();
+        if call.func.variadic {
+            // Phase 1 intentionally keeps variadic arity bounded to parser arg_types length.
```
agavra: "Phase 1" in this context is unclear, we should reference RFC-5 explicitly

timeseries/src/promql/evaluator.rs
```diff
@@ -1455,6 +1455,74 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
+            // Unbounded variadics (for example label_join string tails) are deferred until
+            // string-argument function support lands.
+            (expected_args_len.saturating_sub(1), expected_args_len)
```
agavra: just to make sure I understand this correctly, we allow exactly 0 or 1 of the variadic parameter? IIUC only the last parameter to a function can be variadic

timeseries/src/promql/evaluator.rs
```diff
@@ -1455,6 +1455,74 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
+
+    fn expected_call_arg_type(
+        call: &Call,
+        idx: usize,
+    ) -> EvalResult<promql_parser::parser::value::ValueType> {
+        if idx < call.func.arg_types.len() {
+            return Ok(call.func.arg_types[idx]);
+        }
+
+        call.func.arg_types.last().copied().ok_or_else(|| {
```
agavra: we should probably confirm its variadic before returning the last one if `idx >= call.func.arg_types.len()`

timeseries/src/promql/functions.rs
```diff
@@ -164,6 +182,211 @@ impl PromQLFunction for UnaryFunction {
...
+            ));
+        };
+        let mut samples = first_arg.into_instant_vector()?;
+        let to_nearest = match args_iter.next() {
+            None => 1.0,
+            Some(second_arg) => {
+                if args_iter.next().is_some() {
+                    return Err(EvaluationError::InternalError(
+                        "round accepts at most two arguments".to_string(),
+                    ));
```
agavra, line 227: just to make sure I understood, shouldn't `validate_call_signature` catch this? (defensive is OK, so we can keep it, but I wanted to check my udnerstanding)

agavra (commented):
Thanks @samsond - this LGTM! Honestly, I think the code is easier to review than the RFC, I would consider dropping the RFC as it doesn't really help clarify future documentation and instead file issues for the phase 2 follow-ups.

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,336 @@
...
+- Non-variadic functions require exact arity.
+- Variadic functions allow at least `len(arg_types) - 1`.
+- For `label_join`, `sort_by_label`, and `sort_by_label_desc`, max arity is
+  unbounded.
+- Type-checking for variadic overflow reuses the final declared arg type.
+
+Implication for this RFC:
+
+- AST metadata on `Call.func` is reliable for baseline type expectations.
+- Rust parser `variadic: bool` is not rich enough to represent Prometheus
```
rodesai: I found this a bit confusing. Why does the parser need to tell us about the style of variadic the fn uses? Isn't that a static property of the function tracked in the function registry?

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,336 @@
...
+```rust
+pub(crate) struct FunctionCallContext<'a> {
+    pub eval_timestamp_ms: i64,
+    pub raw_args: &'a [Box<Expr>],
+}
+
+pub(crate) trait PromQLFunction {
+    fn apply(
+        &self,
+        evaluated_args: Vec<Option<ExprResult>>,
```
rodesai: what's the convention here? `evaluated_args` has the values for all non-static arguments? Why not just model the non-static arguments as `ExprResult`?

timeseries/rfcs/0005-promql-function-signatures.md
rodesai: Nvm I think I get it. The promql-parser crate has a registry of function metadata that we want to treat as the source of truth for the set of functions, and the `func` field is populated from this set. That makes sense to me.

timeseries/rfcs/0005-promql-function-signatures.md
samsond: Make sense. I will update the RFC.

timeseries/rfcs/0005-promql-function-signatures.md
samsond: Good point. I will include.

timeseries/src/promql/evaluator.rs
samsond: cool, I will reference as suggested.

timeseries/src/promql/evaluator.rs
samsond: Yes, your understanding is correct for the current RFC-5 scope.

In our evaluator, `variadic: true` is treated as an **optional trailing argument**, so allowed arity is:

- `min = len(arg_types) - 1`
- `max = len(arg_types)`

So effectively we allow 0 or 1 occurrence of that trailing variadic slot.

And yes, only the last parameter is variadic. We are intentionally not modeling unbounded variadic tails yet (for example `label_join(...src_labels)`), pending parser metadata improvements.

timeseries/src/promql/evaluator.rs
samsond: Good call, I will update.

timeseries/src/promql/functions.rs
samsond: Yes, that’s correct: `validate_call_signature` should catch over-arity on the evaluator query path before dispatch.

I kept this check intentionally as a defensive guard for direct/internal `apply_args` calls (including unit tests) that can bypass evaluator-level validation.

timeseries/rfcs/0005-promql-function-signatures.md
samsond: Great question. The convention here is positional, not non-static only:

- evaluated_args.len() == raw_args.len()
- Some(ExprResult) for arguments we evaluate
- None for AST-only/static arguments (for example string literals), which are read from raw_args[idx]

So evaluated_args is really a slot-aligned view of the original call arguments.

If this were just Vec<ExprResult> for evaluatable args, mixed signatures would lose index alignment with raw_args, and handlers would need extra bookkeeping to map evaluated values back to original argument positions.

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,309 @@
...
+- Use parser metadata as the runtime signature source, and defer
+  Prometheus-exact variadic cardinality and experimental metadata parity to an
+  upstream `promql-parser` contribution.
+  Tracking issue: <https://github.com/GreptimeTeam/promql-parser/issues/129>.
+
+### Proposed Function Argument Model
+
+Use one unified function-call interface at the function boundary.
+
+- Reuse `ExprResult` directly for evaluated non-string arguments.
```
cadonna: Wouldn't it be simpler to add the `String` type to `ExprResult`? In the end, `String` is a valid type to which an expression can evaluate (see https://prometheus.io/docs/prometheus/latest/querying/basics/#expression-language-data-types).

timeseries/src/promql/evaluator.rs
cadonna: Thanks for the explanation!
That is indeed a bit confusing. IMO, different concepts are merged together.
For me variadic means that the last argument can be repeated zero or more times. There is no default value involved.
Optional arguments can be passed to the function or not (0 or 1 time). If they are not passed the default value is used.
It seems that in Prometheus both of these concepts are called variadic.

timeseries/src/promql/evaluator.rs
```diff
@@ -1455,6 +1455,80 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
...
+            ))
+        })
+    }
+
+    fn validate_call_signature(&self, call: &Call) -> EvalResult<()> {
+        let actual_args_len = call.args.args.len();
+        let (min_args, max_args) = Self::call_arity_bounds(call);
+
+        if min_args == max_args {
+            if actual_args_len != min_args {
```
cadonna: Isn't this already verified in the parser?
I found this test: https://github.com/GreptimeTeam/promql-parser/blob/1fb5fc40502a132268fb0155b46ece19c2c0a31d/src/parser/parse.rs#L1301

cadonna: It also seems that the parser tests the cases below: https://github.com/GreptimeTeam/promql-parser/blob/345cd679887f48d1c6d6684da06f8368b6709f61/src/parser/ast.rs#L1561

cadonna: Is there a good reason, we verify the number of args again?

samsond: Great question. Yes, parser validates arity for parsed queries. We keep evaluator-side validation as defense-in depth for non-parser AST entry paths (internal/tests/manual `Call` construction), so malformed calls fail consistently before dispatch.

samsond: Now that `promql-parser` v0.8.0 includes richer function metadata, we can simplify this logic to mirror parser semantics more directly. But we still want a thin evaluator guard for non-parser entry paths.

cadonna: Could you give examples? I am a bit worried that we make the evaluator more complex than needed.

samsond: Fair concern.

You’re right that for normal query strings, parser-side checks already protect the happy path. The evaluator guard is mainly for bypass/invariant paths.

Concrete non-parser paths we have today are tests that construct `Call` AST nodes directly and invoke `evaluate_call(...)` (without `parse(...)`), for example:

- [test_evaluate_call_scalar_argument](https://github.com/opendata-oss/opendata/blob/main/timeseries/src/promql/evaluator.rs#L3730)
- [test_evaluate_call_scalar_expression_argument](https://github.com/opendata-oss/opendata/blob/main/timeseries/src/promql/evaluator.rs#L3767)
- and in this change: `should_error_when_call_has_too_many_arguments`

Without evaluator validation, invalid calls in those paths still fail, but later and inconsistently (function-specific errors like `round accepts at most two arguments`) instead of a normalized call-signature boundary error.

So this is intended as a thin defense-in-depth guard, not a second custom rule system.

timeseries/rfcs/0005-promql-function-signatures.md
samsond: Good question. I considered `ExprResult::String`, but kept it out intentionally for now.

In this design, `ExprResult` is the runtime value channel used by evaluator execution paths (scalar/vector/matrix). String literals are currently treated as AST-only function arguments (read from `raw_args`), which matches the Prometheus call model where string args are not passed as evaluated value slots.

Adding `ExprResult::String` now would widen evaluator/result-handling paths (dispatch, boundary checks, response handling) without immediate runtime use, since we do not yet support standalone string-valued evaluation in this engine.

If we later need first-class string evaluation/results, we can add `ExprResult::String` as a follow-up without changing the function-call model.

timeseries/src/promql/evaluator.rs
cadonna: Thanks for the examples!

I looked into them and I have the following comments:

- `validate_call_signature()` has basically the same content as `check_ast_for_call()` in the promql_parser (except for the inf and NaN checks for log and exp functions). So I would not call it a thin guard since it does equally much as the parser.
- As far as I understand, we have a check (`validate_call_signature()`) in the production code that is only needed by tests. In addition, the main benefit of the check is that it does error out earlier. If we skip `validate_call_signature()`, we would get also an error, but during evaluation. Note that the check would have the same content, i.e., the name of the function and that the number of arguments is wrong.
- `should_error_when_call_has_too_many_arguments()` is not really an example since it tests `validate_call_signature()`. Test `should_error_when_call_has_too_many_arguments()` only exists because `validate_call_signature()` exists, not the other way around.

My conclusion is that `validate_call_signature()` does not add any value. If the parser is used, the parser validates the call. If `Call` is directly constructed, the error is thrown during evaluation.

I apologize for being difficult on this. I just do not see any value in `validate_call_signature()`. It is more code to maintain, it is always called on the evaluation path, but only needed for tests, and it adds a validation that partly needs to be done in each function evaluation anyways. I say "partly" because in theory in function evaluations you do not need to check for too many arguments. I would rather prefer to completely validate the arguments in each function evaluation than introducing `validate_call_signature()`.

timeseries/rfcs/0005-promql-function-signatures.md
cadonna: I am not sure, I can follow.

I see that PromQL does not have string operators or functions that return a string. However, there are some functions that take strings as arguments (e.g., `label_join()`). So we need a way to pass a string to functions. If `ExprResult` does not contain a string, we need some other way to pass a string to a function. I can see the argument not adding a string to `ExprResult`, because the overall result of a PromQL expression cannot be a string, currently. Maybe a different type (e.g. `PromQLArg` as it is used now) with a string could be used for function arguments instead of `ExprResult`.

timeseries/src/promql/evaluator.rs
samsond: Thanks, agreed. I will remove validate_call_signature() and related helper logic from evaluator so parser validation remains the single source of truth on normal query paths. That will allow us to rely on function-level checks for direct/internal call paths, and I will updated tests accordingly to assert function-path errors rather than evaluator pre-validation errors.

cadonna: @samsond great that we found an agreement!
Thanks a ton for being open to my comments!

timeseries/rfcs/0005-promql-function-signatures.md
samsond: Thanks, this is a good callout.

I will updated the RFC/code direction to separate function inputs from expression outputs:

- `ExprResult` stays the expression result channel (`Scalar | InstantVector | RangeVector`).
- Function inputs use `PromQLArg` (it will include range vectors), plus `raw_args` in `FunctionCallContext` for AST string literals.
- In evaluator we will materialize positional argument slots (`Vec<Option<PromQLArg>>`), where string literal positions are `None` and are read from `raw_args` by string-capable functions.

So I agree with your point: string handling belongs in function-argument modeling, not in `ExprResult` right now.  If we later support first-class string expression results, we can add `ExprResult::String` as a separate follow-up.

timeseries/src/promql/evaluator.rs
samsond: Thanks for the thoughtful review. Glad we aligned.

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,305 @@
...
+## Design
+
+### Decisions
+
+- Use parser metadata on `Call.func` (`arg_types`, `return_type`, `variadic`,
+  `experimental`) as the signature source of truth.
+- Use one runtime-dispatched `PromQLFunction` interface at the function
+  boundary.
+- Keep a thin evaluator runtime signature guard for non-parser AST entry paths
+  (for example manually constructed calls in tests/internal paths).
```
cadonna: I guess this is an oversight since we decided not to do this in a dramatic discussion 😉.

timeseries/rfcs/0005-promql-function-signatures.md
```diff
@@ -0,0 +1,305 @@
...
+We considered matching it exactly. We chose a typed enum argument list because
+it is easier to reason about in Rust and maps naturally to current evaluator
+results.
+
+### Parser-Only Validation
+
+Relying exclusively on parser validation was considered. We still keep runtime
+checks for robustness in tests and internal call paths that construct AST nodes
+directly.
+
```
cadonna: We do now parser-only validation, right? 
(Except for validating too many arguments in function evaluation).

timeseries/src/promql/evaluator.rs
```diff
@@ -3797,6 +3836,302 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn should_evaluate_call_with_optional_second_argument() {
```
cadonna: Shouldn't this be now specific to `round()` and go into `function.rs` where `apply_args()` can be called directly?
I am fine with moving and changing this in a future PR.

Same applies to all added tests.

timeseries/src/promql/functions.rs
```diff
@@ -164,6 +182,213 @@ impl PromQLFunction for UnaryFunction {
...
+
+    fn apply_args(
+        &self,
+        args: Vec<PromQLArg>,
+        _eval_timestamp_ms: i64,
+    ) -> EvalResult<Vec<EvalSample>> {
+        let mut args_iter = args.into_iter();
+        let Some(first_arg) = args_iter.next() else {
+            return Err(EvaluationError::InternalError(
+                "clamp requires three arguments".to_string(),
```
cadonna: nit: This error and the following ones could also contain the actual number of arguments passed. Same applies to the other clamp functions.

Can go into a future PR.

timeseries/src/promql/functions.rs
```diff
@@ -681,6 +906,133 @@ mod tests {
...
+                ],
+                1000,
+            )
+            .unwrap();
+
+        assert!(result.is_empty());
+    }
+
+    #[test]
+    fn should_propagate_nan_when_clamp_bound_is_nan() {
```
cadonna: IMO, we need to test all three cases for `clamp()`:
1. min is NaN, max and sample are floats
2. max is NaN, min and sample are floats
3. sample is NaN, min and max are floats

Similar applies to `clamp_max()` and `clamp_min()` but with just two cases to consider.

Can also go into a future PR.

timeseries/rfcs/0005-promql-function-signatures.md
samsond: Correct this is a stale RFC wording before removing the evaluator-side validation. I will update it.

timeseries/rfcs/0005-promql-function-signatures.md
samsond: Correct, I will update this as well.

timeseries/src/promql/evaluator.rs
samsond: Good point. Addressed in the current PR

timeseries/src/promql/functions.rs
samsond: Addressed.

timeseries/src/promql/functions.rs
samsond: Good point. Addressed in the current PR.

cadonna:
🥳

PR #288, opened 2026-03-05
Add `close()` to timeseries and vector

### LINKED ISSUE #287: Add close APIs for TimeSeriesDb/VectorDb
The close() APIs should implicitly flush as done in LogDb. We should have integration tests which verify that the flush occurs.

ryandielhenn:
## Summary

Adds `close()` to timeseries and vector with implicit flush.

## Related Issues

Fixes #287

## Test Plan

This PR adds integration tests that write data to timeseries/vector databases, call close on them, then reopen them and check that the written data survived the close.

## Checklist

- [X] Tests added/updated

### DIFF
--- a/vector/src/db.rs
+++ b/vector/src/db.rs
@@ -595,7 +595,7 @@ impl VectorDb {
         self.write_coordinator
             .stop()
             .await
-            .map_err(|e| anyhow::anyhow!(e))?;
+            .map_err(Error::Internal)?;
         self.storage.close().await?;
         Ok(())
     }


timeseries/src/timeseries.rs
```diff
@@ -219,4 +219,74 @@ impl TimeSeriesDb {
...
         self.tsdb.flush().await
     }
+
+    /// Closes the time series database, flushing any pending data and releasing
+    /// resources.
+    ///
+    /// All written data is flushed to durable storage before the database is
+    /// closed. For SlateDB-backed storage, this also releases the database
+    /// fence.
+    pub async fn close(self) -> Result<()> {
```
agavra, line 229: we should consider implementing `Drop` to call close on drop for timeseries, vector and log (if we haven't already).

agavra (approved):
LGTM! Thanks @ryandielhenn - one suggestion

timeseries/src/timeseries.rs
ryandielhenn: Can you clarify what you mean by 'call close on drop'. I figure you mean just abort the background tasks. `close` and `flush` are both async so actually calling them from drop would require `block_in_place`.

agavra: Ah, that makes sense to me. It'd be nice to have a way to avoid drop without close, but for now this is fine. Thanks for digging into that! (I did mean flush + close, didn't think about the fact they were async)

ryandielhenn: Ok, now that I'm clear on the requirement I'll think through how to implement.

agavra: Nah let's just leave it for now. I think what we have is good, SLateDB doesn't do this either so I think it might be standard. I appreciate it though!

PR #290, opened 2026-03-05
[log] Implement read_durable semantics

hachikuji:
Now that we have slatedb 0.11, we have access to the durable sequence number. This patch uses this to implement `read_durable` semantics for `LogDb`. We update `Storage` to expose sequence numbers and the subscription API. We then implement the bookkeeping in `LogDb` to expose a view corresponding to the durable sequence number.

### DIFF
--- a/log/src/log.rs
+++ b/log/src/log.rs
@@ -7,7 +7,6 @@
 
 use std::ops::RangeBounds;
 use std::sync::Arc;
-use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::Duration;
 
 use async_trait::async_trait;
@@ -105,7 +104,6 @@ pub struct LogDb {
     epoch_watcher: EpochWatcher,
     read_subscriber_task: JoinHandle<()>,
     read_visibility: ReadVisibility,
-    required_visible_durable_epoch: AtomicU64,
 }
 
 impl LogDb {
@@ -284,23 +282,15 @@ impl LogDb {
     /// This method ensures that all acknowledged writes are durably persisted
     /// to storage.
     pub async fn flush(&self) -> Result<()> {
-        let durable_epoch = self.handle.flush().await?;
-        if self.read_visibility.is_remote() {
-            self.required_visible_durable_epoch
-                .fetch_max(durable_epoch, Ordering::Relaxed);
-        }
+        self.handle.flush().await?;
         Ok(())
     }
 
     /// Waits for read-side visibility to reach the current requirement.
     async fn sync_reads(&self) -> Result<()> {
-        let (target, durability) = if self.read_visibility.is_remote() {
-            (
-                self.required_visible_durable_epoch.load(Ordering::Relaxed),
-                Durability::Durable,
-            )
-        } else {
-            (self.handle.flushed_epoch(), Durability::Written)
+        let (target, durability) = match self.read_visibility {
+            ReadVisibility::Remote => (self.handle.durable_epoch(), Durability::Durable),
+            ReadVisibility::Memory => (self.handle.written_epoch(), Durability::Written),
         };
         self.epoch_watcher
             .clone()
@@ -398,7 +388,6 @@ impl LogDb {
             epoch_watcher,
             read_subscriber_task,
             read_visibility,
-            required_visible_durable_epoch: AtomicU64::new(0),
         })
     }
 }
--- a/log/src/writer.rs
+++ b/log/src/writer.rs
@@ -385,8 +385,8 @@ impl LogWriteHandle {
         Self::recv_epoch(result_rx).await
     }
 
-    /// The highest epoch flushed to storage (but not necessarily durable).
-    pub(crate) fn flushed_epoch(&self) -> u64 {
+    /// The highest epoch that has reached written visibility.
+    pub(crate) fn written_epoch(&self) -> u64 {
         *self.watcher.written_rx.borrow()
     }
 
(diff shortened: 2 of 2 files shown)


common/src/storage/slate.rs
```diff
@@ -282,7 +291,24 @@ impl Storage for SlateDbStorage {
...
+        })
+    }
+
+    fn subscribe_durable(&self) -> watch::Receiver<u64> {
+        let slate_rx = self.db.subscribe();
+        let (tx, rx) = watch::channel(slate_rx.borrow().durable_seq);
+        tokio::spawn(async move {
+            let mut slate_rx = slate_rx;
+            while slate_rx.changed().await.is_ok() {
+                let durable_seq = slate_rx.borrow_and_update().durable_seq;
```
agavra: we may want to consider whether the DB was closed and break in that scenario as well. otherwise you may get into weird situations where the database doesn't drop the main tx because it notices a rx is still alive (I had some weird behaviors like that in unit tests)

log/src/config.rs
```diff
@@ -44,6 +45,18 @@ pub struct Config {
...
+    ///
+    /// When `true`, the read view advances only after the storage engine reports
+    /// that written data has been durably persisted (e.g., flushed to the WAL and
+    /// object store). This provides stronger consistency guarantees at the cost
+    /// of higher read latency.
+    ///
+    /// When `false` (the default), reads see data as soon as it hits the
+    /// memtable, which is faster but means visible data may not yet be durable.
+    #[serde(default)]
+    pub read_durable: bool,
```
agavra: I'd consider making this an enum like in SlateDB so eventually there can be different levels of durability (`Memory`, `Remote` and eventually `Local` if disk durability is sufficient). Basically the difference between acks=none and acks=all

log/src/view_tracker.rs
```diff
@@ -0,0 +1,162 @@
...
+//! advances the watermark immediately after each push.
+
+use std::collections::VecDeque;
+use std::sync::Arc;
+
+use crate::model::SegmentId;
+use common::storage::StorageSnapshot;
+
+/// A pending write waiting for durable confirmation.
+pub(crate) struct PendingEntry {
```
agavra: why do we have two different structs for this? they look identical.

log/src/log.rs
```diff
@@ -49,23 +52,17 @@ use crate::writer::{LogWrite, LogWriteHandle, LogWriter, LogWriterConfig, Writte
...
 ///
-/// Appended records are not immediately visible to reads. The
-/// [`flush()`](LogDb::flush) method ensures all pending data is durably
-/// persisted to storage, but does not block on reader synchronization.
-/// Instead, read operations ([`scan`](LogRead::scan),
-/// [`list_keys`](LogRead::list_keys),
-/// [`list_segments`](LogRead::list_segments)) synchronize themselves by
-/// waiting for the reader view to catch up to the coordinator's current
-/// flushed watermark before returning results. This means:
+/// Visibility depends on `Config::read_durable`:
```
agavra: should this be a top-level config or a per read request config?

log/src/log.rs
```diff
@@ -286,11 +287,24 @@ impl LogDb {
     /// This method ensures that all acknowledged writes are durably persisted
     /// to storage.
     pub async fn flush(&self) -> Result<()> {
-        self.handle.flush().await
+        self.handle.flush().await?;
+        if self.read_durable {
+            let durable_seq = *self.storage.subscribe_durable().borrow();
```
agavra: note that this won't necessarily immediately give you the sequence that was just flushed. there's some async stuff that happens and the `borrow` will not wait for a new value, it'll just use the current value (which is initialized and then updated later on each flush), probably makes sense to have a background thread subscribe and update `self.required_visible_durable_seq` (or better yet, just hold a reference to the `self.storage.subscribe_durable()` watch and just call `borrow` on it every time and get rid of `required_visible_durable_seq`)

EDIT: I may have lost track of all the different watch mechanisms, but I feel like we should try to have as few of them as possible (e.g. there's both the storage watch and the notifier on the atomic ints for visible_durable_seq)

log/src/log.rs
```diff
@@ -300,6 +314,27 @@ impl LogDb {
         Ok(())
     }
 
+    /// In read-durable mode, waits until the read view has advanced to at
+    /// least the required durable sequence (set by flush barriers).
+    async fn wait_for_required_visible_durable(&self) -> Result<()> {
```
agavra: if you apply my previous comment we could just use the subscription watch `wait_for` function

log/src/log.rs
hachikuji: Yeah, this was kind of nasty. I cleaned this up so that we're just using EpochWatchers. Let me know if that addresses the concern.

log/src/log.rs
hachikuji: The implementation is a bit simpler with a global read durability. I thought we would start there and see if anyone needs per-read config.

log/src/view_tracker.rs
hachikuji: Agreed. I consolidated both as `ViewEntry`

log/src/config.rs
hachikuji: Sounds good. Would `ReadVisibility` be a reasonable name?

common/src/storage/slate.rs
hachikuji: I implemented a persistent task which can be closed directly when Storage is closed.

log/src/config.rs
agavra: Yup, that sounds good to me

PR #294, opened 2026-03-06
add stateless ingest ingestor

cadonna:
## Summary

Adds the ingestor for the stateless ingest. The ingestor writes the ingested data in batches to object storage and enqueues the locations of the data batches to a queue. In addition to the location, the ingestor also writes metadata to the queue and the ingestion time.   

## Test Plan

Unit tests

## Checklist

- [x] Tests added/updated

### DIFF
--- a/ingest/src/ingestor.rs
+++ b/ingest/src/ingestor.rs
@@ -69,8 +69,13 @@ impl DataAndNotifiers {
         metadata: Bytes,
         ingestion_time_ms: i64,
         notifier: Notifier,
-    ) {
-        let start_index = self.entries.len() as u32;
+    ) -> Result<()> {
+        let start_index: u32 = self.entries.len().try_into().map_err(|_| {
+            Error::InvalidInput(format!(
+                "batch entry count {} exceeds u32::MAX",
+                self.entries.len()
+            ))
+        })?;
         self.entries.extend(entries);
         self.metadata.push(Metadata {
             start_index,
@@ -78,6 +83,7 @@ impl DataAndNotifiers {
             payload: metadata,
         });
         self.notifiers.push(notifier);
+        Ok(())
     }
 
     fn is_empty(&self) -> bool {
@@ -107,13 +113,14 @@ impl Batch {
         ingestion_time_ms: i64,
         notifier: Notifier,
         now: SystemTime,
-    ) {
+    ) -> Result<()> {
         self.size_bytes += entries.iter().map(|e| e.len()).sum::<usize>() + metadata.len();
         self.data_and_notifiers
-            .add(entries, metadata, ingestion_time_ms, notifier);
+            .add(entries, metadata, ingestion_time_ms, notifier)?;
         if self.started_at.is_none() {
             self.started_at = Some(now);
         }
+        Ok(())
     }
 
     fn take(&mut self) -> DataAndNotifiers {
@@ -174,8 +181,9 @@ impl BatchWriterTask {
                 msg = rx.recv() => {
                     match msg {
                         Some(IngestMessage::Add { entries, metadata, ingestion_time_ms, notifier }) => {
-                            self.batch.add(entries, metadata, ingestion_time_ms, notifier, self.clock.now());
-                            if self.batch.size_bytes >= self.flush_size_bytes {
+                            if let Err(e) = self.batch.add(entries, metadata, ingestion_time_ms, notifier.clone(), self.clock.now()) {
+                                let _ = notifier.send(Some(Err(e)));
+                            } else if self.batch.size_bytes >= self.flush_size_bytes {
                                 let _ = self.write_batch().await;
                             }
                         }
--- a/ingest/src/queue.rs
+++ b/ingest/src/queue.rs
@@ -558,7 +558,8 @@ impl ConflictCounter {
             return 0.0;
         }
         let conflicts = self.conflict_count.load(Ordering::Relaxed);
-        (conflicts as f64 / writes as f64) * 100.0
+        let rate = (conflicts as f64 / writes as f64) * 100.0;
+        rate.min(100.0)
     }
 }
 


ingest/src/ingestor.rs
```diff
@@ -0,0 +1,699 @@
...
+        &mut self,
+        entries: Vec<Bytes>,
+        metadata: Bytes,
+        ingestion_time_ms: i64,
+        notifier: Notifier,
+        now: SystemTime,
+    ) {
+        self.size_bytes += entries.iter().map(|b| b.len()).sum::<usize>();
+        self.entries.extend(entries);
+        self.metadata = metadata;
```
hachikuji: Do we not need to accumulate metadata?

ingest/src/ingestor.rs
```diff
@@ -0,0 +1,699 @@
...
+    }
+
+    fn with_object_store(
+        config: IngestorConfig,
+        object_store: Arc<dyn ObjectStore>,
+        clock: Arc<dyn Clock>,
+    ) -> Result<Self> {
+        let producer =
+            QueueProducer::with_object_store(config.manifest_path.clone(), object_store.clone());
+        let (tx, rx) = mpsc::unbounded_channel();
```
hachikuji: Should we use a bounded channel?

ingest/src/ingestor.rs
```diff
@@ -0,0 +1,699 @@
...
+            cancellation_token,
+            handle,
+            producer,
+            clock,
+            pending_bytes,
+            max_unflushed_bytes: config.max_unflushed_bytes,
+        })
+    }
+
+    pub async fn ingest(&self, entries: Vec<Bytes>, metadata: Bytes) -> Result<WriteHandle> {
```
hachikuji, line 356: Can we add some docs for these?

ingest/src/ingestor.rs
```diff
@@ -0,0 +1,699 @@
...
+            })
+            .map_err(|_| Error::Storage("ingestor shut down".to_string()))?;
+        self.pending_bytes
+            .fetch_add(incoming_size, Ordering::Release);
+        Ok(WriteHandle { rx: notifier_rx })
+    }
+
+    async fn maybe_apply_backpressure(&self, incoming_size: usize) -> Result<()> {
+        let threshold = self.max_unflushed_bytes;
+        if incoming_size > threshold {
```
hachikuji: This check is a little puzzling to me. Is this just a max size check for the incoming data? How is a user expected to handle the error?

ingest/src/ingestor.rs
```diff
@@ -0,0 +1,699 @@
...
+        let (notifier_tx, notifier_rx) = tokio::sync::watch::channel(None);
+        self.tx
+            .send(IngestMessage::Write {
+                entries,
+                metadata,
+                ingestion_time_ms,
+                notifier: notifier_tx,
+            })
+            .map_err(|_| Error::Storage("ingestor shut down".to_string()))?;
+        self.pending_bytes
```
hachikuji: Not sure how rigid this needs to be, but since we update the pending bytes after adding to the channel, it is possible for the bytes on the channel to exceed the limit.

ingest/src/ingestor.rs
cadonna: Oh man, what a mistake! Bad Claude! No treat for you today 😁 !
Of course, we need to accumulate metadata.

ingest/src/ingestor.rs
cadonna: Yes, of course!

ingest/src/ingestor.rs
cadonna: This is for when the user adds a batch that is larger than the backpressure threshold. 
Since the batch might be much larger than the threshold, I thought we cannot do anything else than throw an error to not exceed the configured memory limit of the ingestor. Now, that I look again at the code, maybe it is not needed to be that strict, we could just block until the entries are written to object storage. This is my thinking: The entries are already in memory. Merely the ownership of the entries is moved into the ingestor. If the entries are copied as part of the move (rust does not guarantee that the data is not copied when moved), we already have an issue before we reach this point in the code. So it would be better to avoid that more entries are passed to the ingestor which we can do by blocking the call.

ingest/src/ingestor.rs
cadonna: In `maybe_apply_backpressure()` I have the following condition that should avoid that:

```rust
if self.pending_bytes.load(Ordering::Acquire) + incoming_size >= threshold {
    self.flush().await?;
}
```

Did I misunderstand something?

ingest/src/ingestor.rs
cadonna: I am not against it. What would be a good limit, though? I would like to avoid adding a config.

ingest/src/ingestor.rs
cadonna: Done in d1e7147d5cb83cb9ec7d2f31a1f96292e3aabd0d

ingest/src/ingestor.rs
cadonna: I removed the check and the error in 128db67

ingest/src/ingestor.rs
cadonna: I made a logical error with the metadata. To resolve that, I came up with the following:
- `Ingestor::ingest()` takes one metadata for the vector of entries it ingests.
-  A data batch can consists of multiple concatenated vectors of entries. Thus, the queue entry for a batch  needs to contain the corresponding vector of metadata.
- The format of the queue now stores a vector of metadata if at least one metadata in the batch is non-empty, otherwise the metadata does not use up any space in the queue entry.

I updated the RFC and the code.

I also added an open question to the RFC:
```
Should we ingest metadata per entry or metadata per vector of entries, i.e., per call to `Ingestor::ingest()`?     
```
IMO, ingesting metadata per call to `ingest()` is quite flexible, because the caller pass metadata per call to `ingest()`, but they could also structure the passed metadata in such a way that it contains metadata per entry.
WDYT?

cadonna: BTW, Thank you for spotting this rather bad bug.

cadonna: Done in  876c9df

cadonna: I just realized that metadata per call to `ingest()` does not work since this concept does not exist on the collector side. Consequently, the metadata cannot be mapped to the ingested entries. 🤦

cadonna: Done in dc409c60a459cd285c6529f886e07ce7b546edef

ingest/src/ingestor.rs
hachikuji: Adding to the channel first creates a number of races. For example, multiples entries may see the same pending bytes. They may all be added to the channel before pending bytes is updated. What is worse, the item on the channel may be read before `pending_bytes` is incremented, which could lead to underflow when we try to decrement. It might be simpler to use a bounded queue as the backpressure mechanism for this first pass.

ingest/src/ingestor.rs
```diff
@@ -0,0 +1,668 @@
...
+
+    fn add(
+        &mut self,
+        entries: Vec<IngestEntry>,
+        ingestion_time_ms: i64,
+        notifier: Notifier,
+        now: SystemTime,
+    ) {
+        for entry in entries {
+            self.size_bytes += entry.data.len();
```
hachikuji: Do we not care about the metadata size? Perhaps we should call it `data_size_bytes` to be explicit?

ingest/src/ingestor.rs
```diff
@@ -0,0 +1,668 @@
...
+        if self.ingestion_time_ms == 0 {
+            self.ingestion_time_ms = ingestion_time_ms;
+        }
+        self.notifiers.push(notifier);
+        if self.started_at.is_none() {
+            self.started_at = Some(now);
+        }
+    }
+
+    fn take(&mut self) -> (Vec<Bytes>, Vec<Bytes>, i64, Vec<Notifier>) {
```
hachikuji: Can we create a struct for the return type? That would give us names which makes the code easier to follow.

ingest/src/ingestor.rs
```diff
@@ -0,0 +1,668 @@
...
+        notifier: Notifier,
+    },
+    Flush {
+        done: tokio::sync::oneshot::Sender<Result<()>>,
+    },
+}
+
+struct Batch {
+    entries: Vec<Bytes>,
+    metadata: Vec<Bytes>,
```
hachikuji: Why not reuse IngestEntry instead of using separate vectors? Perhaps we could even have a single struct to add in the Notifier.
```rust
struct BatchEntry {
  entry: IngestEntry,
  notifier: Notifier,
}
```

ingest/src/ingestor.rs
```diff
@@ -0,0 +1,668 @@
...
+    ingestion_time_ms: i64,
```
hachikuji: Can you clarify the usage of this? Is it the same thing as the timestamp of the first entry? Another name might be clearer.

ingest/src/ingestor.rs
```diff
@@ -0,0 +1,668 @@
...
+}
+
+enum IngestMessage {
+    Write {
+        entries: Vec<IngestEntry>,
+        ingestion_time_ms: i64,
+        notifier: Notifier,
+    },
+    Flush {
+        done: tokio::sync::oneshot::Sender<Result<()>>,
```
hachikuji: The name threw me off a little. How about just `result_sender`?

ingest/src/ingestor.rs
```diff
@@ -0,0 +1,668 @@
...
+            ingestion_time_ms: 0,
+            notifiers: Vec::new(),
+            size_bytes: 0,
+            started_at: None,
+        }
+    }
+
+    fn add(
+        &mut self,
+        entries: Vec<IngestEntry>,
```
hachikuji: If entries is empty, the notifier may be left hanging because `write_batch` returns early when the batch is empty. I can't think of a case where a user would try to add an empty `entries`. Perhaps it can be an error?

ingest/src/ingestor.rs
```diff
@@ -0,0 +1,668 @@
...
+        self.entries.is_empty() && self.metadata.iter().all(|m| m.is_empty())
+    }
+}
+
+struct BatchWriter {
+    object_store: Arc<dyn ObjectStore>,
+    producer: Arc<QueueProducer>,
+    path_prefix: String,
+    batch_interval: Duration,
+    batch_max_bytes: usize,
```
hachikuji: I think the way this is implemented, it is more of a `flush_min_bytes`. We trigger `write_batch` when the threshold is crossed. I think `flush_interval` might be clearer as well.

ingest/src/ingestor.rs
```diff
@@ -0,0 +1,668 @@
...
+
+    async fn write_and_enqueue(
+        &self,
+        entries: Vec<Bytes>,
+        metadata: Vec<Bytes>,
+        ingestion_time_ms: i64,
+    ) -> Result<()> {
+        let non_empty: Vec<Bytes> = entries.into_iter().filter(|e| !e.is_empty()).collect();
+        let location = if non_empty.is_empty() {
+            String::new()
```
hachikuji: Can you explain what is going on here?

ingest/src/ingestor.rs
```diff
@@ -0,0 +1,668 @@
...
+
+    /// Flush the current batch, blocking until all pending entries are durably written.
+    pub async fn flush(&self) -> Result<()> {
+        let (done_tx, done_rx) = tokio::sync::oneshot::channel();
+        self.tx
+            .send(IngestMessage::Flush { done: done_tx })
+            .map_err(|_| Error::Storage("ingestor shut down".to_string()))?;
+        done_rx
+            .await
+            .map_err(|_| Error::Storage("ingestor shut down".to_string()))?
```
hachikuji: Does the error make sense in all cases?

hachikuji:
@cadonna Side thought. I'm considering how stateless ingest fits within opendata. Could we treat it as a first-class system in its own right rather than just an ingest source? It is similar to log for example, but its semantics are closer to queue. It may amount to the same thing, but for example, we could move it out of "ingest" and into "queue-like-thingy" (better name tbd). I can see this system being used on its own for any kind of simple write pipeline into another system.

ingest/src/ingestor.rs
cadonna: That is again a miss on my side. It is similar to the metadata situation. I need to store the ingestion time for each entry instead of for each call to `ingest()`.

ingest/src/ingestor.rs
cadonna: I agree on the `flush_interval`. IMO, also `flush_min_bytes` is not a good name since it suggests that at least that amount of bytes is flushed, which is not true because due `flush_interval` less bytes than `flush_min_bytes` can be flushed. I renamed `flush_max_bytes` to `flush_size_bytes` as the public config. for the ingestor.

Done in f6e23ce

cadonna: I agree on the `flush_interval`. IMO, also `flush_min_bytes` is not a good name since it suggests that at least that amount of bytes is flushed, which is not true because due `flush_interval` less bytes than `flush_min_bytes` can be flushed. I renamed `flush_max_bytes` to `flush_size_bytes` as the public config. for the ingestor.

Done in https://github.com/opendata-oss/opendata/commit/f6e23ce16b99ee88356a722e93fd7adac7b020f7

ingest/src/ingestor.rs
cadonna: I changed this to a bounded channel for now.

ingest/src/ingestor.rs
cadonna: Created struct `DataAndNotifiers` in d9a31832f33030b6ad80b8a500d829f32b9cbf87.

ingest/src/ingestor.rs
cadonna: True. Added a check and related error in d9a31832f33030b6ad80b8a500d829f32b9cbf87.

ingest/src/ingestor.rs
cadonna: OK, I changed it again.
`Ingestor::ingest()` takes a vector of entries and a single metadata payload. In the queue, the metadata and the ingestion time are mapped to a start index in the entries batch. I chose this approach because it seemed the most flexible. It reduces the size used up by metadata when the metadata is the same for each ingest call, but it still allows to provide metadata per entry by using an appropriate format for the metadata payload or by ingesting entry-by-entry.

Done in d9a31832f33030b6ad80b8a500d829f32b9cbf87.

ingest/src/ingestor.rs
cadonna: I changed the mapping between entry and metadata. See https://github.com/opendata-oss/opendata/pull/294/changes#r2918597847

ingest/src/ingestor.rs
cadonna: Done

ingest/src/ingestor.rs
cadonna: I disallow an empty entries vector, but I allow a non-empty vector of empty elements.

ingest/src/ingestor.rs
cadonna: No, but it is wrong with the current design. Removed it.

ingest/src/ingestor.rs
cadonna: I added also the size of the metadata to the check. In future, we can handle them separately, if needed.

ingest/src/ingestor.rs
cadonna: What do you mean exactly?
Do you mean that 
1. `Error::Storage` does not make sense in all cases or 
2. that the message does not make sense?

Regarding 1, I can see that it might make sense to introduce a new error. 
What could cause the sender to drop except for a stopped background task?

cadonna:
> @cadonna Side thought. I'm considering how stateless ingest fits within opendata. Could we treat it as a first-class system in its own right rather than just an ingest source? It is similar to log for example, but its semantics are closer to queue. It may amount to the same thing, but for example, we could move it out of "ingest" and into "queue-like-thingy" (better name tbd). I can see this system being used on its own for any kind of simple write pipeline into another system.

Yeah, I agree! That might be a likely and great development.

ingest/src/ingestor.rs
```diff
@@ -0,0 +1,739 @@
...
+            .put(&path, PutPayload::from(payload))
+            .await
+            .map_err(|e| Error::Storage(e.to_string()))?;
+
+        self.producer.enqueue(path.to_string(), metadata).await?;
+
+        Ok(())
+    }
+
+    async fn close(&mut self, rx: &mut mpsc::Receiver<IngestMessage>) {
```
hachikuji: This method duplicates a lot of work that the run loop is already doing. Feels like there ought to be a cleaner approach. I think it might simplify the structure if BatchWriter owned the tx channel. Rather than sending through the channel directly, callers would invoke an API on BatchWriter which would send to the channel. Then `close` could be rewritten to send an explicit Flush rather than duplicating the run logic. Since BatchWriter would own the tx channel, no additional writes would be accepted. The rust ownership model would give us that for free. Also just better encapsulation.

ingest/src/ingestor.rs
hachikuji: What I was wondering if underlying storage errors should be forwarded through.

ingest/src/ingestor.rs
cadonna: Thanks for this comment!

I did what you said (at least I hope I got it right) and learned better rust.

ingest/src/ingestor.rs
cadonna: Do you mean since this receiver contains the error from the flush, we can pass it through with `unwrap()`?
I have changed it that way now.

Issue #272, opened 2026-03-03
switch the implementations of `MergeOperator` to implement `merge_batch` and remove `merge`


rodesai:
## Summary

The opendata `MergeOperator` trait currently has fns for merging a pair of rows and a batch of multiple rows. Merging batches of rows is generally preferable since it only has to deserialize, compute the merge, and serialize the result one time for the whole batch. Slatedb will always prefer `merge_batch` over `merge`. Internally it always calls `merge_batch` and has a default impl which implements it using `merge` to support merge operators that only implement `merge`.

We should switch the implementations of  the opendata `MergeOperator` to all implement `merge_batch`. Once we've done that we can remove `merge` from the `MergeOperator` in opendata, and have the `SlateDbMergeOperatorAdapter` in `common::slate` implement `merge` by calling `merge_batch`.

## Additional Context

See https://github.com/opendata-oss/opendata/pull/263 for some relevant discussion and an example of implementing `merge_batch`

micllam:
Hi, I'll try implementing this for the timeseries case as a first step.

micllam:
I'll next try implementing this for the remaining vector cases.

PR #295, opened 2026-03-06
[timeseries] Implement merge_batch

micllam:
## Summary

Implements `common::storage::MergeOperator::merge_batch()` for `OpenTsdbMergeOperator`.

I added some `merge_batch_` helper functions and modified the existing `merge_` helpers to call their new batch counterparts.

## Related Issues

Relates to #272. I have only tackled part of that issue.

## Test Plan

Added unit tests.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/storage/merge_operator.rs
+++ b/timeseries/src/storage/merge_operator.rs
@@ -3,8 +3,9 @@ use bytes::Bytes;
 use crate::model::RecordTag;
 use crate::serde::bucket_list::BucketListValue;
 use crate::serde::inverted_index::InvertedIndexValue;
-use crate::serde::timeseries::merge_time_series;
+use crate::serde::timeseries::merge_batch_time_series;
 use crate::serde::{EncodingError, RecordType, record_type_from_tag};
+use common::storage::default_merge_batch;
 
 /// Merge operator for OpenTSDB that handles merging of different record types.
 ///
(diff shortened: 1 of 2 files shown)


timeseries/src/storage/merge_operator.rs
agavra, line 17: we should have this implementation just delegate to merge_batch. if we do that we can remove all the non-batch implementations and keep just the production batch merging code

agavra (approved):
This LGTM! Thanks @micllam very clean PR. Just one comment that I think will help clean up a lot of code. Thanks for your patience waiting for a review, I think this one slipped past me. In the future if reviews are sitting un-acknowledged for more than a business day feel free to ping us on discord (and if you aren't already hanging out there you totally should)

timeseries/src/storage/merge_operator.rs
micllam: I've made this change now, but I kept the following check at the start of `merge()`:
```
if existing_value.is_none() {
    return new_value;
}
```
Otherwise, I had the following test failures and wasn't quite sure how to fix them:
```
---- storage::merge_operator::tests::should_return_new_value_when_no_existing_value stdout ----

thread 'storage::merge_operator::tests::should_return_new_value_when_no_existing_value' (414781) panicked at timeseries/src/storage/merge_operator.rs:35:18:
Failed to batch merge inverted index: EncodingError { message: "Failed to deserialize RoaringBitmap: unknown cookie value" }

---- reader::tests::slatedb_writer_and_reader_coexist_no_fencing stdout ----

thread 'reader::tests::slatedb_writer_and_reader_coexist_no_fencing' (414708) panicked at timeseries/src/reader.rs:595:14:
called `Result::unwrap()` on an `Err` value: Execution("Storage error: Data error: invalid sequence number ordering during merge. expected sequence numbers in descending order, but found 1 followed by 2")
```

micllam:
Thanks for your review, and no problem about the delay. I'll make sure to join the discord server :)

timeseries/src/storage/merge_operator.rs
agavra: Interesting, I'll take a look at those test failures today. I hadn't ever seen the "cookie" one, that looks internal to the roaring bitmap crate

micllam: Sorry for not investigating this further yesterday, but I think the first failing test was just because it couldn't decode "new_value" as a roaring bitmap, and had previously been passing due to the early exit in `merge()`. I've now modified the test to use a proper roaring bitmap and removed the early exit, and the test passes.

Regarding the other failing test, it's no longer failing for me after repeated runs, so for now I think it's fine.

Issue #101, opened 2026-01-17
[tsdb] support parallel ingest to tsdb


rodesai:
As a follow-up to #95 we should support parallel ingest to a single tsdb bucket. With #95 we will serialize ingest to a given bucket. I expect this will bottleneck ingestion, as the work to maintain the current delta is computationally expensive. To allow for parallel ingest, we'd need to shard a given tsdb bucket over multiple tasks (proportional to the number of cores). This would look something like this:
- A tsdb server runs some number of ingest shards (proportional to the number of cores)
- Request handler tasks group scrapes and remote write requests by shard into shard writes, mapping a given time series to a shard by hashing the labels. Each request writes shard writes into a per-shard channel.
- There is a single writer task per channel that consumes shard writes and writes them into a pending delta for the shard. The only shared state is the series id counter for each mini tsdb.
- There is a single writer task that pulls pending deltas and writes them into slatedb as new write batches. If/when SlateDB supports direct SST writes then this could just write new SSTs (rather than going through the memtable write path)

PR #297, opened 2026-03-06
add stateless ingest collector

cadonna:
## Summary

Adds the collector for the stateless ingest. The collector gets the locations of the ingested batches from the queue and reads the batches from object storage. The collector also helps to maintain the sequence number of the last acked batch. The caller of the collector is responsible for persisting the last acked sequence number.

## Test Plan

Unit tests

## Checklist

- [x] Tests added/updated

### DIFF
--- a/ingest/src/collector.rs
+++ b/ingest/src/collector.rs
@@ -353,7 +353,7 @@ mod tests {
 
         // Acking sequence 5 when last acked was 0 should fail
         let result = collector.ack(5).await;
-        assert!(result.is_err());
+        assert!(matches!(result, Err(Error::Storage(msg)) if msg.contains("out-of-order ack")));
     }
 
     #[tokio::test]


ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -316,38 +330,52 @@ pub struct CollectorConfig {
...
+On the first call (when no batch has been acked yet), it peeks the earliest entry in the queue.
+On subsequent calls, it reads the entry with sequence `last_acked_sequence + 1`.
+If no entries are available, `None` is returned.
 
-By calling `ack()` the caller confirms that the batch has been fully processed.
-The method calls `dequeue()` on the queue consumer to remove the acknowledged entries from the
-queue manifest.
+By calling `ack(sequence)` the caller confirms that the batch with the given sequence number
+has been fully processed. Acks must be in order — acking a sequence that is not immediately
+after the last acked sequence returns an error. To amortize the cost of manifest writes,
```
apurvam, line 351: Not sure how to parse to reconcile: "acking a sequence that's not immediately after the last acked sequence returns an error" with "the collector batches acks and calls dequeue every 100 acks". Does this imply the ack-batch must start with the next sequence after the last ack, and that all acks in the batch must have a sequential sequence number?

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -316,38 +330,52 @@ pub struct CollectorConfig {
...
+
+Since the batch and the corresponding sequence number are atomically written to the
+OpenData subsystem, the sequence number always represents the last written batch.
+After a failure, the ingest can continue at the first batch that was not ingested.
+No data is ingested twice.
 
-The delivery guarantees of the entries that were not confirmed to be durable depends on the progress tracking of the
-caller of the stateless ingest.
+The delivery guarantees of the entries on the ingestor-side that were not confirmed to be 
+durable before a failure depends on the progress tracking of the caller of the ingestor.
```
apurvam, line 378: I'm having a hard time parsing this sentence. What is the ingestor in this diagram? the service writing data that's picked up by the collector? Or the service that picks up the output of the collector. Does the stateless ingest library help this ingestor with tracking progress at all? Basically, I assume that whatever happens, the ingestor needs to commit the data + sequence number atomically to guarantee exactly once ingestion, right? Would be worth spelling this out.

ingest/src/collector.rs
```diff
@@ -0,0 +1,557 @@
...
+        producer
+            .enqueue("batches/first".to_string(), vec![])
+            .await
+            .unwrap();
+
+        let batch = collector.next_batch().await.unwrap().unwrap();
+        collector.ack(batch.sequence).await.unwrap();
+        collector.close().await.unwrap();
+
+        // After close, entries should be dequeued
```
apurvam, line 439: Not sure I follow this test. After the close, entries are flushed. the new collector initializing with the 'None' last acked sequence should basically initialize to the batch of the next sequence after the previous collectors final ack, which is empty?

ingest/src/collector.rs
```diff
@@ -0,0 +1,557 @@
...
+            .enqueue("batches/second".to_string(), vec![])
+            .await
+            .unwrap();
+
+        let first = collector.next_batch().await.unwrap().unwrap();
+        collector.ack(first.sequence).await.unwrap();
+
+        // Acking sequence 5 when last acked was 0 should fail
+        let result = collector.ack(5).await;
+        assert!(result.is_err());
```
apurvam: Is there an error code to assert?

cadonna:
> Thanks bruno. this looks good to me. I left a few comments for clarification.
> 
> My bigger question is how this fixes #101 ?

I was under the impression that stateless ingest is a generalization of #101 . I can also create a new issue if you like.

ingest/rfcs/0001-stateless-ingest.md
cadonna: What do you mean with "ack-batch"?
`ack()` takes the sequence number of the data batch to ack. The sequence number of a data batch is part of the queue entry of the data batch. Batches can only be acked in sequence number order. The guarantee is that the order of the data batches ingested by the same ingestor is preserved.

To reduce contention on the queue manifest due to clean-up, the queue entry is not removed/dequeued for every ack. Every 100 calls to `ack()`, the ack'ed queue entries with a sequence number less than the current ack'ed sequence number are removed/dequeued from the queue. This is correct because the queue entries are strictly processed and ack'ed in sequence number order.

ingest/rfcs/0001-stateless-ingest.md
cadonna: The diagram in the RFC shows the ingestor and the collector.
The ingestor is called by the component that wants to ingest data to a opendata subsystem. In terms of responsive's metrics pipeline that is the ingest-gateway. The ingestor accepts data, batches the data, writes the data batches to object storage, and enqueues the queue entry consisting of the location of the batch, the metadata and the sequence number of the batch to the queue.
The collector is called by the writer that writes data to the opendata subsystem (a single writer is enforced). The collector reads the queue and passes data batches ordered by the sequence number to the writer. It also acks the written the batches and cleans up processed queue entries. In terms of responsive's metrics pipeline that is the shared consumer in the metrics-backend.
If the writer atomically persists the sequence number of a batch when writing the batch, exactly-once delivery can be guaranteed.
On the ingestor-side, it is left to the caller of the ingestor to track what data was successfully ingested. Stateless ingest guarantees that the data was written to object storage and the location was enqueued to the queue if the durability watcher returns success. If the caller of the ingestor fails after the location was enqueued, but the durability watcher has not yet returned success, the caller needs to decide if it wants to re-ingest the unconfirmed data (at-least-once) or to skip it (at-most-once).

The previous two paragraphs describe the exactly-once delivery on the collector side also mentioning the atomic write to the subsystem. 
This paragraph should describe the possibility of at-least-once on the ingestor side and calls out that stateless ingest does not provide any progress tracking.

ingest/src/collector.rs
cadonna: Yeah, that is a good point. I will do that! Thanks!

cadonna: Done in 8dc648a

ingest/src/collector.rs
cadonna: The important thing here is that when the collector closes it will dequeue/flush the ack'ed queue entries so that a possible new collector that does not have a last acked sequence number does not process data that has been already processed. Normally, the second collector has the last ack'ed sequence number since it can read it from the opendata subsystem if it was atomically written with the data. This is more an edge case.
I could also add a second entry to the queue and verify that second batch is returned in `collector2.next_batch()` if that is better to understand.

ingest/rfcs/0001-stateless-ingest.md
apurvam: Got it, so the consumer in the metrics backed creates a collector, initializes it, calls `next_batch()` continuosly (with some back off I assume, if `next_batch`) returns `None`. For a batch, it remote writes to the timeseries server. Once the batch is written, it calls `ack(sequence)` with the sequence of the batch that was written. 

That makes sense. 

But going back to my previous comment on the ack-batch: if the metrcis backend calls `ack(sequence)` 88 times, but then the metrics backend dies. the data would be in the timeseries, but not dequened. Assuming that the metrics backend is also writing the sequeences atomically with data in timeseries, then the expectation is that when the metrics backend comes back online, it will read old sequences from the collector again, but because these sequences have been tracked by the metrics backend, the metrics backend would just skip the duplicates. 

And if the metrics backend didn't store the sequences, it would rewrite the batches that were not successfully dequeued. 

Is that right?

ingest/rfcs/0001-stateless-ingest.md
```diff
@@ -316,38 +330,52 @@ pub struct CollectorConfig {
...
+On subsequent calls, it reads the entry with sequence `last_acked_sequence + 1`.
+If no entries are available, `None` is returned.
 
-By calling `ack()` the caller confirms that the batch has been fully processed.
-The method calls `dequeue()` on the queue consumer to remove the acknowledged entries from the
-queue manifest.
+By calling `ack(sequence)` the caller confirms that the batch with the given sequence number
+has been fully processed. Acks must be in order — acking a sequence that is not immediately
+after the last acked sequence returns an error. To amortize the cost of manifest writes,
+the collector batches acks and only calls `dequeue()` on the queue consumer every 100 acks.
```
apurvam, line 352: Should we consider a `CollectorConfig` and sholud have the `flushInterval` be a config there. This would align with the `Config` paradigm of other components too. i'm fine with postponing this too, especially since it adds to the API and there is no use case for configuring this yet. So it's up to you.

apurvam (approved):
Thanks Bruno, this looks good to me for now. We can iterate forward as we go. I don't think this fixes 101 though. that is about parallel ingestion into TSDB, which is work that needs to be done independently of whether we have a collector at the stateless ingest layer or not.

ingest/rfcs/0001-stateless-ingest.md
cadonna: I wanted to leave such a config to when we see the need for it.

ingest/rfcs/0001-stateless-ingest.md
cadonna: Yes, this is correct. 
Just one note, not the metrics backend would skip the already-processed queue entries, but the collector would do that. At start-up, the collector is initialized with the last ack'ed sequence number:

```rust
collector.initialize(last_acked_sequence)
```

PR #302, opened 2026-03-07
add reader client for vector

rodesai:
## Summary

- moves vector query logic to new QueryEngine type
- adds a VectorDbReader client that uses a StorageReader with QueryEngine for read-only access to vector

## Test Plan

unit tests added

## Checklist

- [x] Tests added/updated

### DIFF
--- a/vector/src/query_engine.rs
+++ b/vector/src/query_engine.rs
@@ -117,14 +117,11 @@ impl QueryEngine {
 
         let centroid_ids = self.prune_centroids(&centroid_ids, query);
 
-        let sorted_lists = self
-            .load_and_score(&centroid_ids, query, self.storage.clone())
-            .await?;
+        let sorted_lists = self.load_and_score(&centroid_ids, query).await?;
         if sorted_lists.is_empty() {
             return Ok(Vec::new());
         }
-        self.resolve_top_k(sorted_lists, k, self.storage.as_ref())
-            .await
+        self.resolve_top_k(sorted_lists, k).await
     }
 
     pub(crate) async fn search_with_nprobe(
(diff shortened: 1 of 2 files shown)


vector/src/query_engine.rs
```diff
@@ -0,0 +1,662 @@
...
+        storage: Arc<dyn StorageRead>,
+    ) -> Self {
+        Self {
+            options,
+            centroid_graph,
+            storage,
+        }
+    }
+
+    pub async fn search(&self, query: &[f32], k: usize) -> Result<Vec<SearchResult>> {
```
hachikuji: nit: should be pub(crate)?

vector/src/query_engine.rs
```diff
@@ -0,0 +1,662 @@
...
+            .collect()
+    }
+
+    /// Spawn a task per centroid to load its posting list and score all candidates
+    /// against the query vector. Returns per-centroid sorted candidate lists.
+    async fn load_and_score(
+        &self,
+        centroid_ids: &[u64],
+        query: &[f32],
+        storage: Arc<dyn StorageRead>,
```
hachikuji: nit: we have access to `self.storage` already? Similar for `resolve_top_k`.

vector/src/reader.rs
```diff
@@ -0,0 +1,190 @@
...
+        };
+
+        Ok(Self {
+            options,
+            storage,
+            centroid_graph: Arc::from(centroid_graph),
+        })
+    }
+
+    fn query_engine(&self) -> QueryEngine {
```
hachikuji: Could we move this to a field?

vector/Cargo.toml
```diff
@@ -27,6 +27,7 @@ rstest.workspace = true
 tokio.workspace = true
 serde.workspace = true
 serde_yaml.workspace = true
+slatedb.workspace = true
```
hachikuji, line 30: Why do we need the direct dependency?

rodesai: this is left around from a different change. I'll clean it up

rodesai: ah wait no it's because we pass the slatedb reader options directly to `create_storage_read` in common

vector/src/reader.rs
rodesai: Yeah will do. At some point I want the ability to take a snapshot from the db reader so we can do snapshot-aligned reads with a reader that keeps itself fresh. But we need support from slate, plus we need some way to keep the centroid graph fresh. I'll change it for now

PR #303, opened 2026-03-08
add initial limited support for attribute filtering to vector db

rodesai:
## Summary

- Introduce Query struct from RFC-00003 with support for a subset of filters. Specifically we support eq, neq, in, and, or.
- change the db query interfaces to use Query struct rather than just a vector.
- during ingest, write inverted index posting updates to delta
- at query time, evaluate filters on the postings returned from nearby centroids.

## Test Plan

unit tests added

## Checklist

- [x] Tests added/updated


rodesai:
I went ahead and merged this one to unblock my server PR. I'll look out for any post-merge feedback

PR #304, opened 2026-03-08
Create 0006-graph-database.md

StevenBtw:
## Summary

Defines the storage model for a labeled property graph database built on SlateDB. Maps all graph primitives (nodes, edges, labels, properties, adjacency) to SlateDB records using the standard 3-byte key prefix (subsystem `0x05`), with index structures for label lookups, property searches and traversal. Integrates Grafeo (v0.5.28) as the GQL query engine by implementing its `GraphStore` and `GraphStoreMut` traits over SlateDB's ordered key-value interface. Concurrency control is delegated to SlateDB's built-in transaction support with snapshot isolation.

## Related Issues

Fixes #

## Test Plan

- Record encoding/decoding verified against implementation in `graph/src/serde/`
- Key layout consistency verified against RFC 0001 (record key prefix) and RFC 0004 (common encodings)
- Record tag allocations (`0x10`-`0xF0`) verified to have no collisions with other subsystems
- NodeRecordValue encoding uses `FixedElementArray<u32>` per RFC 0004
- SequenceAllocator block-based ID allocation matches common crate implementation
- Merge operator pattern matches existing timeseries precedent
- Catalog bidirectional storage (6 sub-types) aligned with test implementation

## Checklist

- [ ] Tests added/updated


rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,1074 @@
...
+
+**Authors**:
+
+- [Steven Grond](https://github.com/StevenBtw)
+
+## Summary
+
+This RFC defines the storage model for a labeled property graph (LPG) database built on
+[SlateDB](https://github.com/slatedb/slatedb). The graph database integrates
+[Grafeo](https://github.com/grafeo-db/grafeo) (v0.5.16) as its query engine, implementing Grafeo's
```
cadonna: Gives me a 404. Probably a typo.
The correct link is https://github.com/GrafeoDB/grafeo

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,1074 @@
...
+│  label_ids:   FixedElementArray<u32 LE>                │
+│               (label_count elements)                   │
+└────────────────────────────────────────────────────────┘
+```
+
+**Flags:**
+
+| Bit  | Name      | Description                              |
+|------|-----------|------------------------------------------|
+| 0    | `DELETED` | Node was deleted at this epoch           |
```
cadonna: Could slateDB tombstones and snapshots also be used instead of the DELETED flag and the MVCC epoch?

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,1074 @@
...
+
+**Value Schema:** Empty (all information is in the key).
+
+**Structure:**
+
+- The key is designed for efficient prefix scans:
+  - `[prefix, src_node_id]`, all outgoing edges from a node (implements `neighbors(node, Outgoing)`)
+  - `[prefix, src_node_id, edge_type_id]`, outgoing edges of a specific type (type-filtered
+    traversal)
+  - `[prefix, src_node_id, edge_type_id, dst_node_id]`, check if a specific edge exists
```
cadonna: The trade-off here is that finding any edges between nodes is more expensive. Is that not as common as finding a given type of edge between nodes?

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,1074 @@
...
+**Cost and justification:** Each backward adjacency key is 30 bytes with an empty value, adding
+one extra PUT per edge creation. For N edges this adds ~30N bytes (e.g., ~3 GB for 100M edges).
+This is small relative to the total record set (each edge already produces an EdgeRecord +
+ForwardAdj + property records), and the benefit is first-class `Incoming`/`Both` traversal without
+scanning all ForwardAdj records.
+
+### `LabelIndex` (`0x70`)
+
+Maps a label to the set of nodes that carry it. Stored as individual keys (one per label-node
+pair) to avoid read-modify-write on high-cardinality labels.
```
cadonna: This sounds similar to the mapping between labels and timeseries in the timeseries DB. Maybe similar trade-offs make also sense here. (See https://github.com/opendata-oss/opendata/blob/main/timeseries/rfcs/0001-tsdb-storage.md#invertedindex-recordtypeinvertedindex--0x04)

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,1074 @@
...
+2. Prefix scan `[0x01, 0x70, label_id_be]`
+3. Each key suffix is a `node_id`
+4. Return `Vec<NodeId>`
+
+**Cardinality Estimation:**
+
+To implement `estimate_label_cardinality(label)`, the `Metadata` record stores per-label node
+counts maintained by merge operators (see Metadata section).
+
+### `PropertyIndex` (`0x80`)
```
cadonna: Out of curiosity: Do edges have properties, too?

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,1074 @@
...
+**Range Scan:**
+
+To implement `find_nodes_in_range(property, min, max, min_inclusive, max_inclusive)`:
+
+1. Resolve `property` → `PropertyKeyId`
+2. Encode `min` and `max` as `SortableValue`
+3. Range scan from `[prefix, property_key_id, min_encoded]` to `[prefix, property_key_id, max_encoded]`
+4. Apply inclusivity bounds
+
+### Catalog Records (`0x90` – `0x95`)
```
cadonna: In timeseriesDB, we decided to not use a catalog (a.k.a. dictionary). Maybe the same arguments also apply here. See (https://github.com/opendata-oss/opendata/blob/main/timeseries/rfcs/0001-tsdb-storage.md#dictionary-encoding-for-labels)

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,1074 @@
...
+```
+
+Total: 6 bytes. Value is the name as raw UTF-8 bytes.
+
+#### By-Name Keys (`0x91`, `0x93`, `0x95`)
+
+```text
+┌─────────┬─────────────┬────────────────────┐
+│ version │ record_tag  │      name          │
+│ 1 byte  │   1 byte    │  TerminatedBytes   │
```
cadonna: I believe, we do not need `TerminatedBytes` here, since the length of the name is implicitly given by the length of the full key.

cadonna (commented):
@StevenBtw, Thanks for the RFC!
That is really exciting!
I reviewed half of the RFC and wanted to share my first comments.

rfcs/0006-graph-database.md
StevenBtw: Indeed, this was and old url. It's fixed!

rfcs/0006-graph-database.md
StevenBtw: I don't think so to be honest. This is the same pattern used by [Neo4j's transaction log](https://neo4j.com/docs/operations-manual/current/database-internals/transaction-management/) and [DGraph's MVCC](https://docs.dgraph.io/design-concepts/transaction-mutation-concept): graph-level multi-version concurrency requires application-managed version chains, not storage-level snapshots.

rfcs/0006-graph-database.md
StevenBtw: Correct! Finding all edges between two specific nodes regardless of type requires scanning the full [prefix, src_node_id] range and filtering by dst_node_id, since edge_type_id sits between them in sort order.

This was very intentional. In GQL, edge patterns almost always specify a type: MATCH (a)-[:KNOWS]-> (b) is far more common than MATCH (a) -[]-> (b). The optimizer also benefits from type-filtered scans for cardinality estimation. JanusGraph makes the [same design choice](https://docs.janusgraph.org/schema/advschema/#edge-and-vertex-ordering-via-sort-keys), placing the edge label before the adjacent vertex for the same reason.

For the untyped case the scan is still bounded to one node's adjacency, not a global scan; the cost is proportional to that node's out-degree. And for checking connectivity between two nodes in either direction, the BackwardAdj index means you can pick the lower-degree endpoint and scan from there rather than needing to scan both sides.

rfcs/0006-graph-database.md
StevenBtw: Good catch, I see the similarity. If I'm reading the TSDB RFC correctly, the InvertedIndex stores a RoaringBitmap per label/value pair, so one key holds all matching series IDs. That works well there because time bucketing keeps each bitmap bounded and series membership within a bucket is mostly append-only (I'm guessing).

On the graph side i don't have that kind of natural partitioning. A label like "Person" could apply to millions of nodes, and nodes gain or lose labels all the time. Storing all those node IDs in a single value would mean reading and rewriting the bitmap on every label change, which seems like it would create hot keys for common labels.

Individual keys keep it simple: one PUT to add a label, one DELETE to remove it, no need to read current state. The trade-off is more keys in the tree, but since they all share a long common prefix ([0x01, 0x70, label_id]) I'm expecting SlateDB's prefix compression to handle that reasonably well.

But if we ever need posting-list-style set operations (like intersecting labels for query planning), bucketed bitmaps similar to your TSDB design could be worth looking into further.

rfcs/0006-graph-database.md
StevenBtw: Yes! 
Edges can have properties too. They use EdgeProperty records (0x40) which work the same as NodeProperty (0x30) but keyed by edge ID instead of node ID. The PropertyIndex currently only covers nodes though; we could add edge property indexing later if query patterns call for it.

rfcs/0006-graph-database.md
StevenBtw: To be honest I'm not sure how well those arguments translate to the graph side. From the graph perspective, the catalog feels natural for me: label types and edge types are usually a small, stable set (dozens, not millions) and other graph engines like Neo4j or JanusGraph use integer IDs internally for the same reason. Adjacency keys include the edge type on every entry, so a 4-byte integer instead of a variable-length string adds up when you have billions of edges. But I'd need to look at the TSDB trade-offs more carefully to say whether the same concerns apply here.

rfcs/0006-graph-database.md
StevenBtw: Absolutely right, name is the last field in the key so the length is implicit. I'll change it to raw UTF-8, same as the TSDB InvertedIndex does.

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,1074 @@
...
+### MVCC Model
+
+The graph database uses epoch-based MVCC for snapshot isolation. Each write operation (create
+node, delete edge, etc.) is associated with an epoch. Readers observe a consistent snapshot at
+a specific epoch.
+
+**Epoch Lifecycle:**
+
+1. The global epoch starts at 1 and is stored in `Metadata(Epoch)`.
+2. Each write transaction advances the epoch by 1 on commit.
```
apurvam: What is the `transaction` being referred to here? Later on you just mention using SlateDB `WriteBatch` , which only provides atomicity across write operations. 

More specifically, it looks like the assumption is that the `CurrentEpoch` is read, updated, and written as part of a `WriteBatch`. But the read isn't guarded in a `WriteBatch`. It _seems_ that two concurrent writes could use the same `CurrentEpoch` as a result. Is that true? Should it be allowed in the correctness model?

apurvam (commented):
Thanks for this great RFC! It's super thorough! I'll need multiple passes to digest it. But for now, I left one comment on the MVCC model. Let me know if it's valid. 

More generally, I'm wondering if we can leverage SlateDB Transactions and snapshot isolation more directly. Having the MVCC model managed by the database seems a bit redundant.

rfcs/0006-graph-database.md
StevenBtw: You're correct, two concurrent writers could potentially read the same CurrentEpoch and stamp their records with the same epoch, which violates the one-epoch-per-write invariant. Under the current single-writer assumption this can't happen in practice, but the RFC should have stated that constraint explicitly in the MVCC section. I'll fix that.

I'll look into using SlateDB's DbTransaction to protect the epoch read-increment-write. It solves the atomicity problem cleanly, but introduces some coupling (the graph storage layer would depend on SlateDB's transaction API rather than just Write Batch), which makes the GraphStore implementation less portable. The application-level epoch itself still needs to stay, since Grafeo's query engine reasons about epochs directly (e.g., get_node_versioned, current_epoch). Thanks for the review BTW!

rfcs/0006-graph-database.md
cadonna: I would like to understand this better. SlateDB offers transactions with (serializable) snapshot isolation (https://slatedb.io/rfcs/0011-transaction/). Why is this not enough for a graph database?
Do you maybe know some examples that disqualify slateDB transactions for the MVCC that is needed for graph databases?

rfcs/0006-graph-database.md
apurvam: >  Under the current single-writer assumption this can't happen in practice, but the RFC should have stated that constraint explicitly in the MVCC section.

The system still needs to enforce the single writer, which means it ultimately needs an atomic compare-and-set on the epoch to ensure that you have only one writer for an epoch. Without a compare and set, you could have zombie writers writing with the same epoch. 

For instance, a node loses the network connection to the k8s controller, is marked dead, another node spins up for the same partitions, but the first node is still connected to clients and accepts writes. In this case, you will have two writers for the same partition, and one needs to be fenced out.

rfcs/0006-graph-database.md
StevenBtw: It is probably enough indeed. I think for the initial version we can definitely lean on SlateDB's transaction support for snapshot isolation and not build a separate MVCC layer. The epoch based versioning adds a lot of complexity (version chains, GC, the race condition apurvam found) and the main use case it enables is time travel queries which aren't needed for a first implementation. I'll simplify the design to use SlateDB transactions directly and keep the key layout forward compatible so we can add versioning later if needed.

This does mean I'll need to rework some of Grafeo's GraphStore trait to thread SlateDB transactions through, and drop the epoch based versioned read methods since they won't apply anymore.

rfcs/0006-graph-database.md
StevenBtw: I will revise the RFC, might take a few days.

rfcs/0006-graph-database.md
cadonna: That sounds great!
BTW, this RFC is super interesting!

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,971 @@
...
+- `node_id` (u64): Big-endian node identifier
+
+**Value Schema:**
+
+```text
+┌────────────────────────────────────────────────────────┐
+│                    NodeRecordValue                     │
+├────────────────────────────────────────────────────────┤
+│  label_count: u16 (LE)                                 │
+│  label_ids:   [u32 LE; label_count]                    │
```
cadonna: Could this also be encoded as a `FixedElementArray<T>` from the [common encodings RFC](https://github.com/opendata-oss/opendata/blob/main/rfcs/0004-common-encodings.md#fixedelementarrayt)? Do you still need the count if you use `FixedElementArray<T>`?

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,971 @@
...
+│                     SeqBlockValue                      │
+├────────────────────────────────────────────────────────┤
+│  base_sequence:  u64 (LE), start of allocated block    │
+│  block_size:     u64 (LE), number of IDs in block      │
+└────────────────────────────────────────────────────────┘
+```
+
+Follows the same allocation pattern as the common crate's `SequenceAllocator`.
+
+### Value Serialization
```
cadonna, line 377: Could you add some more info about Grafeo's native values format?
You do not need to list all formats. A link to the format specification or code in the Grafeo crate would be great. Some examples and some details would also be interesting like are the formats backwards compatible.

rfcs/0006-graph-database.md
cadonna: Fair enough!
Great that you also added the reasoning below!

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,971 @@
...
+  - `[prefix, src_node_id]`, all outgoing edges from a node (implements `neighbors(node, Outgoing)`)
+  - `[prefix, src_node_id, edge_type_id]`, outgoing edges of a specific type (type-filtered
+    traversal)
+  - `[prefix, src_node_id, edge_type_id, dst_node_id]`, check if a specific edge exists
+- Edge type is placed before destination to enable type-filtered traversal without loading all
+  edge types. In GQL, edge patterns almost always specify a type: `MATCH (a)-[:KNOWS]->(b)` is
+  far more common than `MATCH (a)-[]->(b)`. For the untyped case the scan is still bounded to
+  one node's adjacency, not a global scan.
+- The `edge_id` suffix ensures uniqueness when multiple edges of the same type connect the same
+  pair of nodes (multi-edges).
```
cadonna: Is this more of an edge case or is this common?
I ask because I am wondering if it would be better to move `edge_id`s into the value. I assume prefix scans are more expensive than point lookups. If multiple edges of the same type is more an edge case, we do also not need to worry too much about read-modify-write operations.

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,971 @@
...
+The adapter uses two patterns depending on the operation type:
+
+1. **Write-only operations** (`create_node`, `create_edge`, `set_*_property`): Use `WriteBatch`
+   via `Storage::apply()`. All records for a single logical operation (entity record, indexes,
+   counter merges) are written atomically in one batch.
+
+2. **Read-then-write operations** (`delete_node`, `delete_edge`, `add_label`, `remove_label`,
+   `remove_*_property`): Use `StorageTransaction` for snapshot-isolated atomicity. The operation
+   begins a transaction, reads the current state within that transaction's snapshot, buffers
+   writes, and commits. This prevents TOCTOU races (e.g., reading a node's labels, then writing
```
cadonna: nit: I have to admit that I had to look up TOCTOU. Could you add a link or/and add the full version of the acronym?

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,971 @@
...
+This was rejected for several reasons:
+
+- **Redundant**: SlateDB already provides snapshot isolation via `DbTransaction`, making a
+  custom MVCC layer unnecessary.
+- **Race condition**: The epoch-based design required reading and incrementing a global epoch
+  counter, which was not protected by a transaction. Two concurrent writers could read the same
+  epoch and stamp their records with duplicate epoch values.
+- **Complexity**: Version chains required scanning and filtering for every point lookup (O(v)
+  where v is the number of versions), garbage collection of old versions, and special handling
+  of deletion markers. Without these, point lookups are O(1) direct `get()` calls.
```
cadonna: nit: I agree that in the context of versions point lookup are O(1) since you only need to get one single key-value pair. However in the context of data in the database a `get()` in slateDB is probably more like a O(log(n)) where n is the number of keys in a block. I did not consider the complexity to find the right block which differs between L0 and sorted runs. 
You probably meant that in the context of versions and your point is still valid, but I just wanted to point this out for completeness. Maybe to avoid misunderstandings you can just say that by using slateDB's MVCC you just need one point lookup instead of a point lookup for each version.

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,971 @@
...
+```text
+Key:   [0x01, 0x50, src_node_id]
+Value: [(edge_type_id, dst_node_id, edge_id), ...]
+```
+
+**Rejected because:**
+
+1. **Read-modify-write**: Adding or removing one edge requires loading, deserializing, modifying,
+   re-serializing, and writing the entire list. For high-degree nodes (thousands of edges), this
+   is prohibitively expensive.
```
cadonna: For adding, you actually do not need to deserialize and re-serialize. Since each element has constant size you could use the `FixedElementArray<T>` format and just append the new element at the end of the list without deserializing the whole list.
For removing, you could look up the edge ID in the serialized list by jumping from one edge ID offset to the next without deserializing the whole list.
It is not for free, but it can be reduced to offset arithmetic and byte level copying.

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,971 @@
...
+Total: 14 bytes.
+
+**Value Schema:** Empty.
+
+**Label Scan:**
+
+To implement `nodes_by_label(label)`:
+
+1. Resolve `label` → `LabelId` via catalog
+2. Prefix scan `[0x01, 0x70, label_id_be]`
```
cadonna: For a high-cardinality label like `Person`, I am not sure if the trade-off of avoiding read-modify-write with using a prefix scan is worthwhile. Scans need to look into all levels, i.e., current write batch, memtable, L0, and sorted runs and potentially need to load many blocks. A point look-up can stop once the most recent key-value pair is found and needs to load one block.
I agree that writing is simpler with the described model, but reading might be overly expensive with that model.
If we have labels mapped to lists of node_ids, we might be able to optimize the read-modify-write operation since the node_ids are simple integers.
This is just a feeling. Experiments might proof me wrong.

cadonna (commented):
@StevenBtw Thanks for the updates!

Here my comments.

rfcs/0006-graph-database.md
cadonna: I discussed this matter with @rodesai and he pointed me to a draft RFC for slate that will make scans more performant and consequently your trade-off worthwhile:
https://github.com/slatedb/slatedb/pull/1380

StevenBtw: I am set to leave for a short ski holidays, so unfortunately won't have a lot of spare time until monday next week, so my reponse might take a week. Thanks for the review, and I will evaluate and compare both options.

cadonna: Have fun!

cadonna:
Note that there is a PR that updates the common key prefix in opendata. It looses the restriction on the record type in the key and adds a subsystem byte to the prefix.
https://github.com/opendata-oss/opendata/pull/326

StevenBtw:
During the last weekend of my holiday Grafeo started trending and was posted on several channels (not by me) so I have seen a huge increase in real users (that encountered real bugs) so I have been very busy bugfixing and implementing community requested features. 

Good news is that the project is in a much better state now, than a week ago, but I have not had the time to properly look at this RFC. The goal is to update the rfc today. Thank you for the patience.

rfcs/0006-graph-database.md
StevenBtw: Yes definitely, `FixedElementArray<u32>` fits here since the value contains only label IDs. I'll drop the label_count: u16 prefix and derive the count from value_length / 4. Will update the RFC

rfcs/0006-graph-database.md
StevenBtw: For sure! I will add a link to the Value::serialize()/Value::deserialize() implementation in grafeo-common.

rfcs/0006-graph-database.md
StevenBtw: Multi-edges are common in several domains, knowledge graphs (multiple "authored" edges between a person and a paper with different roles), financial networks (multiple transactions between accounts), and social graphs (multiple interaction types). Grafeo's data model explicitly supports them, so the storage layer needs to handle them efficiently. All my own usecases have this also.

rfcs/0006-graph-database.md
StevenBtw: Will update to "Time-of-check to time-of-use (TOCTOU)" with a link.

rfcs/0006-graph-database.md
StevenBtw: Fair point. I will replace the O(1) claims with "single point lookup" to avoid implying anything about SlateDB's internal complexity.

rfcs/0006-graph-database.md
StevenBtw: My reasoning was that even with offset arithmetic it's still a read-modify-write on a single key, which under concurrent edge creation means contention and individual keys sidestep that entirely. What do you think?

rfcs/0006-graph-database.md
StevenBtw: Thanks for digging into this with @rodesai. The SlateDB scan optimization RFC (slatedb/slatedb#1380) kind of thing that tips the balance here. With that in place, the individual-key approach should do well even for high-cardinality labels.

That said, I'll be honest, I'm less confident about the performance characteristics at the storage engine level than the graph model level. If you or @rodesai see anything concerning once we start benchmarking, I'd appreciate the feedback. This is an area where your expertise with SlateDB internals is really valuable.

rfcs/0006-graph-database.md
StevenBtw: @apurvam The RFC has been revised: custom MVCC is gone entirely, concurrency is now delegated to SlateDB's snapshot isolation (see the "Concurrency Model" section). So the epoch race condition you identified no longer applies.

Your broader point about single-writer enforcement still stands though. The RFC assumes single-writer semantics but doesn't describe how that's enforced. I know SlateDB has writer fencing via the manifest but I'm not deep enough on the details to document it confidently; particularly around the edge cases you raised (zombie writers, k8s failover). I've listed "Write coordination" in Future Considerations but your concern deserves more than that. Would you be open to helping sketch out what the fencing story should look like here?

StevenBtw:
Updated the RFC and tested a draft implementation!

apurvam:
Thanks for the updates @StevenBtw. I’ll take another pass this week.

StevenBtw:
Addressed all outstanding review comments from cadonna's 3/17 review:

- **FixedElementArray for NodeRecordValue (L296):** adopted, label count derived from `value_length / 4`
- **Value format details (L706):** added link to `grafeo-common` source, documented bincode delegation. Clarified variant indices are bincode-assigned, not explicit discriminants.
- **Multi-edges / edge_id in key vs value (L445):** kept in key — multi-edges are uncommon but the layout avoids read-modify-write entirely
- **TOCTOU (L734):** added full expansion and Wikipedia link
- **Label index scan vs bitmap (L512):** kept individual keys with justification. The SlateDB scan optimization RFC (slatedb/slatedb#1380) will further improve this path.
- **Adjacency list append (L863):** individual keys avoid serialization and contention on high-degree nodes

Also bumped Grafeo to v0.5.30 and clarified Value variant index encoding.

---

One thing worth discussing: since all of Grafeo's LPG query languages (GQL, Cypher, Gremlin, GraphQL, SQL/PGQ) translate to the same `LogicalPlan` and execute against the `GraphStore`/`GraphStoreMut` traits defined in this RFC, adding them after this lands is a feature flag, no storage changes needed. Graph algorithms (PageRank, shortest path, community detection, etc.) also operate on `&dyn GraphStore` and are gated behind the `algos` feature. RDF/SPARQL is the exception: Grafeo maintains a separate `RdfStore` for the triple model, which would need its own record types in a future RFC.

The current design ships GQL as the only mandatory interface. Would you prefer to stay opinionated with GQL-only for now, or expose additional languages behind feature flags from the start (the Cargo features are already wired up in the graph crate)? The trade-off is surface area to test and document vs. giving users familiar entry points, someone coming from Neo4j might expect Cypher, someone from a knowledge graph background might expect SPARQL.

apurvam:
Hi Steve,

Thanks for the updates. The RFC is really clear about the storage layout and concurrency model, and also has a nice set of examples for how GrahpStore operations will translate to SlateDB operations. 

My main concern is about performance. In particular, the label index (lables to nodes), property index (properties to nodes), and edges (src, dest, plus edge type) are all fully encoded in the key. Since nodes and edges are usually very high cardinality, this could really blow out the key space. In some sense, the current proposal is basically a fully materialized secondary index for properties, labels, and bi-directional edges.

In other words, The part that worries me is not the raw key length by itself, but the row fan-out from materializing labels, properties, and both adjacency directions as separate records. 
 
I think the RFC would benefit from an explicit keyspace/storage analysis. 

Codex whipped up a simple live-row model: 

  `K_live ≈ N(1 + λ + p_n + i_n) + E(3 + p_e) + 2(L + T + P) + O(1)`

  where:

  - `N` = nodes
  - `E` = edges
  - `λ` = avg labels per node
  - `p_n` = avg node properties per node
  - `i_n` = avg indexed node properties per node
  - `p_e` = avg edge properties per edge
  - `L/T/P` = distinct label / edge-type / property-key vocab sizes

That is:

  - each node contributes `1` `NodeRecord` + `λ` `LabelIndex` rows + `p_n` `NodeProperty` rows + `i_n` `PropertyIndex` rows
  - each edge contributes `1` `EdgeRecord` + `2` adjacency rows + `p_e` `EdgeProperty` rows
  - catalog size is only `2(L + T + P)`, so vocabulary cardinality is probably not the main driver

 If you plug in a fairly small graph with 10M nodes and 100M edges:

  N   = 10M
  E   = 100M
  λ   = 2
  p_n = 10
  i_n = 4
  p_e = 2

then:

K_live ≈ 10M*(1+2+10+4) + 100M*(3+2)
       ≈ 10M*17 + 100M*5
       ≈ 170M + 500M
       ≈ 670M keys

So even at 10M nodes / 100M edges, you are already in the neighborhood of 670 million live KV entries.

With 100M Nodes and 1B edges, you have:

  N   = 100M
  E   = 1B
  λ   = 2
  p_n = 10
  i_n = 4
  p_e = 2

and you get:

K_live ≈ 100M*17 + 1B*5
       ≈ 1.7B + 5.0B
       ≈ 6.7B keys

Breakdown for that 100M / 1B case:

 - 100M NodeRecord
 - 200M LabelIndex
 - 1.0B NodeProperty
 - 400M PropertyIndex
 - 1.0B EdgeRecord
 - 2.0B adjacency keys
 - 2.0B EdgeProperty

Total: 6.7B keys.

 * * End codex model. 

The Property indices (EdgeProperty, NodeProperty) are particularly large, followed by the adjacency index (5B rows on their own).

I buy that SlateDB prefix compression will reduce repeated bytes in sorted SST keys, but it does not reduce the number of rows, WAL/memtable pressure, compaction work, or delete/update fan-out. That matters a lot here because the schema is intentionally highly materialized.

How do existing storage engines for graph handle these problems? Do we want the graph to scale to 10's of millions of nodes, 100's of millions of edges. What read and write throughput do we want to target?

We should consider materializing less information in the key, and instead relying more on on read-time or compaction time computation to reduce the write amplification of the current model .

Thoughts?

StevenBtw:
>........
> We should consider materializing less information in the key, and instead relying more on on read-time or compaction time computation to reduce the write amplification of the current model .
> 
> Thoughts?

Good point, in Grafeo's storage layer I avoid this by using columnar property storage (one column per property key, not one row per entity×property) and chunked adjacency lists with delta-encoded blocks instead of one KV row per neighbor.

I will run some benchmarks on packed property values vs chunked adjecency blocks on a benchmark I already have. I will share the results here and then we can decide which direction is best.

StevenBtw:
I added a benchmark table to the RFC. This is using my draft implementation which can be found here: https://github.com/StevenBtw/opendata/tree/graph/rfc-0006

StevenBtw:
Thanks for the detailed review @apurvam. The keyspace concern is helpfull; the Merged layout was designed to address exactly that. Appendix A has the numbers: fewer keys at scale (6.7B → 2.0B), which proportionally reduces WAL/memtable pressure and compaction work, and traversal is faster at 1000 edges in-memory since one packed value replaces a 1000-key prefix scan.

I have five design decisions where I would appreciate input from you and @cadonna before moving forward. Listed options for each.

### 1. PropertyIndex staleness in Merged mode

Merged mode does blind property index writes; stale entries are filtered at read time by checking candidates against actual values. Staleness is bounded by distinct indexed values per (node, property) not by write count, but it can still accumulate before compaction.

**A:** Add a staleness counter now so we have observability before designing vacuuming. Cheap to implement.
**B:** Defer entirely; the bound is tight enough and compaction vacuuming will handle it at scale.

I lean towards A. Would you want this kind of metric from the start or is it premature?

### 2. Statistics granularity for the optimizer

Only aggregate NodeCount and EdgeCount are persisted today. Per-label cardinality needs a full LabelIndex scan and per-edge-type degree falls back to total_edges / total_nodes regardless of type. This can mislead the optimizer by orders of magnitude (see the KNOWS/OWNS example in RFC).

**A:** Add per-label and per-edge-type Metadata counters now via merge operator. The sub-type space has room; cost is one extra merge operand per mutation.
**B:** Ship aggregate-only, add per-type stats as a fast follow-up. Acceptable if we are okay with suboptimal plans on heterogeneous graphs initially.
**C:** Keep aggregate counters but add a background sampling task that caches per-type estimates. Avoids write-path overhead, adds background task complexity.

I lean towards A since the optimizer ships with cost-based planning and bad degree estimates could hurt from day one. What do you think is the right tradeoff here?

### 3. Single-writer fencing (Open Question #6)

SlateDB has manifest-based fencing but the adapter does not surface fencing errors yet. The zombie-writer scenario @apurvam raised could be a real issue in production settings.

**A:** Define the fencing contract in this RFC: adapter propagates ManifestFencedError as a fatal shutdown signal, server layer drains. Implementation follows in a separate PR.
**B:** Resolve in a separate RFC before production deployment. Keeps this RFC scoped to the storage model.
**C:** Punt; single-writer is enforced by deployment topology.

I lean towards A; the contract should be documented even if the implementation follows later. Is that enough or would you prefer a dedicated RFC for this?

### 4. backward_edges config flag

Config has `backward_edges: bool` (default true) but the writer always creates BackwardAdj entries. The reader checks the flag but the writer does not.

**A:** Remove the flag; always maintain backward adjacency. One extra write per edge, simpler code, no half-implemented config.
**B:** Honor the flag in the writer; skip BackwardAdj writes when false. Saves one PUT per edge for outgoing-only workloads.

I lean towards A. A partially ignored flag is worse than not having it, and the cost is modest. Do you see a use case where skipping backward edges matters enough to justify the complexity?

### 5. Sync-async bridge

Grafeo's query traits (GraphStore/GraphStoreMut) are synchronous by design; query execution is CPU-bound and in-memory. SlateDB is async. The adapter bridges via `block_in_place` which holds a Tokio worker thread per operation.

Grafeo 0.5.32 adds async support at the persistence layer (AsyncStorageBackend, async WAL) but the query traits remain sync. This means the bridge pattern is architecturally aligned with how Grafeo itself works.

**A:** Keep the current `block_in_place` bridge as-is. It matches Grafeo's own architecture and SlateDB I/O will be the bottleneck long before thread pool saturation.
**B:** Move to a dedicated blocking thread pool (`spawn_blocking`) to avoid holding Tokio worker threads. Higher implementation cost, marginal benefit until we have real concurrency numbers.

I lean towards A. The bridge pattern is the same one Grafeo uses internally and the current approach is the simplest correct option. Worth revisiting if concurrency benchmarks show thread pool contention, but I don't expect that to happen before other bottlenecks surface. Curious to hear your thoughts. 

ps. sorry for all the back and forth, I am still learning a lot in all of this.

apurvam:
Thanks for the quick response @StevenBtw. Those benchmark results look promising. I’m traveling thru Wednesday next week, so my responses on this are likely going to be delayed. 

But I think focusing on leveraging the merge operator more would make sense.

rfcs/0006-graph-database.md
```diff
@@ -0,0 +1,377 @@
...
+Adj Add:         [0x01][peer_id:u64 LE][edge_id:u64 LE]
+Adj Remove:      [0x02][peer_id:u64 LE][edge_id:u64 LE]
+```
+
+**PropertyIndex in Merged mode:** Writes use blind puts (no read-before-write) to avoid O(N)
+merge-operand resolution. Stale index entries are filtered at read time in
+`find_nodes_by_property` / `find_nodes_in_range` by verifying candidates against actual values.
+Staleness is bounded by the number of *distinct indexed values* per (node, property) pair —
+not by total write count. Setting `age` to 30 then 40 then 30 again produces at most 2 stale
+entries, not 3. Compaction-time vacuuming (see Future Considerations) will reclaim these.
```
cadonna: > Setting `age` to 30 then 40 then 30 again produces at most 2 stale
entries, not 3.

Does this mean that the second setting to 30 does not write a PropertyIndex record?

And should it not say 

> Setting `age` to 30 then 40 then 30 again produces at most 1 stale
entries, not 2.

because one of the three settings is the current value which is not stale by definition.

cadonna (commented):
Thanks for the changes, @StevenBtw  and sorry for the delay in reviewing. I was OOO.

cadonna:
> 1. PropertyIndex staleness in Merged mode

Here I need some more clarifications. See https://github.com/opendata-oss/opendata/pull/304#discussion_r3052962138.

> 2. Statistics granularity for the optimizer

I would go for option A and improve later if needed.

> 3. Single-writer fencing

I would go for option B. Let's focus on the storage model. As with timeseries, you can have a write API RFC next where we can discuss this. We can always come back to this RFC and make changes here, since we are still in 0.x mode.

> 4. backward_edges config flag

I agree with you on option A.

> 5. Sync-async bridge 

We can discuss this also in a the write API RFC, I guess.

apurvam:
Hi @StevenBtw .. Thanks for iterating on this so quickly. And sincere apologies for my delayed reply. I was traveling through thursday last week and couldn't get to this. I should be able to review much quicker going forward. 

First off, the merged layout looks like the right direction to me. I think we should commit to it by dropping the dual-mode design and making merged the only storage layout in the first version. The rest of OpenData DBs use the merge-operator extensively, and there I don't see a good reason not to commit to it here.

Carrying both `Individual` and `Merged` through the implementation, test matrix, and future API design feels like extra surface area without a clear payoff.

 On the open questions, I would mostly optimize for the simplest path for now:

  1. `PropertyIndex` staleness: I think we should defer adding stats to the point where we are actually trying to build custom compaction or trying to tune existing compaction. 
  2. Statistics granularity: Can't we punt this to a future RFC or optimization work? If we need stats for the Grafeo to work efficiently, we should just add the stats when we are doing the optimization work. If we need to add stats in the storage, we can update this RFC at that time. 
  3. Single-writer fencing: I would go with B and defer this to the write API / deployment discussion.
  4. `backward_edges`: I would go with A and remove the flag.
  5. Sync/async bridge: I would defer this for now as well. I think getting this correct is going to be very important for write performance. But it can be designed in a subsequent write RFC.

The main thing I would still ask to strengthen in this RFC is the value-side stroage design. Right now `packed props blob` and `packed  adj blob` are placeholders. That is the core of the merged design, so I think the RFC should be more concrete there.

The timeseries storage RFC is a good precedent. It does not stop at “compressed samples” or “posting-list value.” It
recommends Gorilla encoding for time series samples and RoaringBitmap for posting-list values, and it explains why those choices fit the workload. I think this RFC should do the same for the new mergeable blobs.

apurvam:
Hi @StevenBtw . Checking in. Let me know if you have any questions on my comments. As I mentioned, specing out the format of the blobs and commiting to using merges is the main items left. Then we can merge this and think about the write implementation. I'm excited for this to land!

StevenBtw:
Thanks for the patience @apurvam @cadonna!

Two main changes based on your feedback:

- **Went with the merged layout.** Dropped the dual Individual/Merged design; merged is now the only layout, matching the rest of OpenData's merge-operator usage.
- **Specced the blob formats** in a new Value Encoding section: byte-level layouts for packed properties and packed adjacency, plus rationale for each choice (why unsorted, why not whole-map bincode, why no delta plus bit-packing yet). High-degree overflow is acknowledged honestly (16MB at 1M edges is a real cliff), with a forward-compatible upgrade path: chunked adjacency, then delta plus bit-pack, then zone-map sidecar.

Also:

- **Per-label and per-edge-type Metadata counters** (`LabelNodeCount`, `EdgeTypeCount`) via merge operator: the KNOWS/OWNS optimizer example from before now resolves correctly. Cost is one extra merge operand per mutation.
- Dropped unused `prop_count:u16` from EdgeRecord, clarified that operand tags (`0x01`/`0x02`) are record-type-scoped, documented missing-counter fallback, bumped Grafeo to 0.5.40.

For the previous five design questions: went with A on stats granularity (the day-one optimizer argument won out) and on removing `backward_edges`. PropertyIndex staleness, single-writer fencing, and sync/async bridge all moved to Future Considerations for the write API RFC per your guidance.

Implementation is on [graph/rfc-0006](https://github.com/StevenBtw/opendata/tree/graph/rfc-0006) with 65 storage and 59 lib tests green. Implementing alongside the RFC caught a few spec gaps (the dead `prop_count` field, optimizer fallback behavior, `sub_type` layout convention); all fixed ofcourse.

StevenBtw:
Regarding the staleness math @cadonna. Walking `age = 30 → 40 → 30` again: two distinct values leave two live PropertyIndex entries, and one always matches the current value, so the bound is `distinct_indexed_values - 1`. I'll update L310-311 to "produces at most 1 stale entry, not 2" and state the minus-one explicitly. Also wiring up the `Value` format note: a link to `Value::serialize` / `Value::deserialize` in `grafeo-common` next to the bincode reference at L191. The format is bincode-stable within a Grafeo minor; any breaking change rides a prefix-version bump per RFC 0001.

On subsystem `0x05`: with #326 merged the 3-byte prefix is final, so I can open a one-line PR against `rfcs/0001-record-key-prefix.md` adding `0x05 → Graph` and link it here before merge. For the two Open Questions at L406-407, I'd like to close both by moving them to Future Considerations. Edge property indexing is a clean defer: no MVP query pattern needs it and the key layout is forward-compatible, a new `0x81 EdgePropertyIndex` tag mirrors `0x80` when demand shows up. 

For multi-graph I'd pick one SlateDB instance per graph: it matches how every other OpenData subsystem is scoped, avoids stamping a graph-id across the ~2B keys in the scale analysis, and a namespace prefix can be added later under a version bump without breaking single-graph deployments.

Updating the rfc with the wording fix, the link, and the Future Considerations move in one commit. @cadonna @apurvam, anything else before a final pass?

apurvam:
Thanks @StevenBtw . I'll take a look in the next couple of days.

apurvam (approved):
Hi @StevenBtw . Sorry for the delay here. This looks good to me, with one procedural change. 

I've found it really helpful to spell out the on disk format fully for the other DBs. So I went ahead and did it for your PR: https://github.com/StevenBtw/opendata/pull/5/changes

My patch basically enumerates every the exact key and value for key type in your RFC. For things like the `Catalog` type, the RFC was a bit underspecified, so I went and specified the byte values for each catalog sub type. But otherwise, it's purely an elaboration of your RFC, which is great! 

If my PR looks good and you merge it to your branch, I'll merge this. 

Thanks again for the patch!

PR #305, opened 2026-03-10
support consistent snapshots of the centroid graph

rodesai:
## Summary

This patch extends the centroid graph to support a snapshotting mechanism that allows for snapshot-consistent views of the centroids:
- splits the centroid graph trait into reader and writer traits
- adds a snapshot() fn to the writer trait (CentroidGraph)
- implements snapshot consistency in UsearchCentroidGraph. The mechanism works by tracking additions and deletions of centroids. When a centroid is added, the epoch it was added is tracked in the in-memory metadata about the centroids. When a centroid is deleted, it is deleted immediately and the deletions are tracked in a deletion log. Searching now specifies an epoch. When searching at the latest epoch, the search simply searches as before. When searching at earlier epochs, the search filters out centroids added after the epoch and post-merges centroids deleted after the epoch. The deletion log is annotated with the epoch the centroid was deleted. UsearchCentroidGraph tracks the set of outstanding snapshots and prunes the log when the oldest snapshot's epoch advances.
- this change requires a couple tweaks to the write coordinator. First, apply now includes the epoch. Second, the creator of the coordinator can supply an optional initial view.
- when a delta is frozen, vector takes a snapshot of the centroid graph and writes it to the frozen view. Reads use the view associated with the slatedb snapshot they are reading from.

## Test Plan

- unit tests for centroid graph changes
- cohere1M benchmark shows equivalent ingest/query performance

## Checklist

- [x] Tests added/updated

### DIFF
--- a/vector/src/lire/rebalancer.rs
+++ b/vector/src/lire/rebalancer.rs
@@ -759,17 +759,6 @@ impl IndexRebalancerTask {
             .await
             .map_err(|e| e.to_string())?;
 
-        // Safety check: if c has < 1 vector (e.g. drained by concurrent ops), abort.
-        if posting_list.is_empty() {
-            warn!(
-                c_from,
-                num_vectors = posting_list.len(),
-                "cannot merge empty centroid",
-            );
-            self.send_finish_merge(c_to).await?;
-            return Ok(());
-        }
-
         // Track original vector IDs for sweep phase, and save (id, vector)
         // pairs for reassignment in Phase 3.
         let original_vector_ids: HashSet<u64> = posting_list.iter().map(|p| p.id()).collect();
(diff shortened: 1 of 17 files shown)


vector/src/lire/rebalancer.rs
```diff
@@ -759,17 +759,6 @@ impl IndexRebalancerTask {
             .await
             .map_err(|e| e.to_string())?;
 
-        // Safety check: if c has < 1 vector (e.g. drained by concurrent ops), abort.
-        if posting_list.is_empty() {
```
rodesai, line 763: this was just an unnecessary check that I removed

rodesai:
Closing since the new indexer doesn't use the centroid graph anymore

PR #307, opened 2026-03-11
feat[timeseries]: implement retention for tsdb

### LINKED ISSUE #20: implement retention in timeseries
## Summary

Currently `timeseries` stores data points indeterminately. It probably makes sense to implement retention like other tsdbs (e.g. prometheus). 

## Proposed Solution

We can use TTL feature in SlateDB on writes, wired in with some default configuration for retention. 

## Alternatives Considered

Manual deletions on some interval. I think it's best to rely on TTL instead and then improve the compaction strategy to be something like TWCS in Cassandra.

## Additional Context

Also see: https://github.com/slatedb/slatedb/issues/1178

b3ll9trix:
## Summary

This PR closes #20 
### Server
1. Adds an optional `RetentionConfig` to `PrometheusConfig` with time accepted in Prometheus `tsdb.retention.time` format, eg., 1d, 15m, 6m. When not configured, defaults to "7d".
2. Optional `ttl` is added to `SlateDbStorageConfig`. Setting this overrides `default_ttl` (from other sources).
3. The `RetentionConfig.time` or its default of PrometheusConfig is set as `SlateDbConfig.ttl` for `tsdb`.

### Library
1. Makes use of the existing optional `Config.retention` config (`timseries/src/config.rs`).
2. Optional `ttl` is added to `SlateDbStorageConfig`. Setting this overrides `default_ttl` (from other sources).
3. The `Config.retention` when configured is set as `SlateDbConfig.ttl` for `tsdb`.

### Server
- [x] 1.1: Optional `RetentionConfig` is set correctly when configured.
- [x] 1.2:  Optional `RetentionConfig` defaults to the default value when not configured.
- [x] 2.1: Optional `SlateDbStorageConfig.ttl` overrides `Settings.default_ttl` when configured.
- [x] 2.2: When the optional `SlateDbStorageConfig.ttl` is not configured, no change in functionality (takes the value from path/files).
- [x] 3.1 `Prometheus.retention` is set to `SlateDbStorageConfig.ttl` when creating storage.

### Library
- [x] 1: Optional `Config.retention` is set correctly when configured.
- [x] 2.1: Optional `SlateDbStorageConfig.ttl` overrides `Settings.default_ttl` when configured.
- [x] 2.2: When the optional `SlateDbStorageConfig.ttl` is not configured, no change in functionality (takes the value from path/files).
- [x] 3: `Config.retention` is set to `SlateDbStorageConfig.ttl` when creating storage.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/examples/query_bench.rs
+++ b/timeseries/examples/query_bench.rs
@@ -14,7 +14,10 @@
 use std::time::{Duration, Instant, SystemTime};
 
 use clap::Parser;
-use timeseries::{ReaderConfig, TimeSeriesDbReader};
+use common::StorageConfig;
+use serde::{Deserialize, Deserializer};
+use slatedb::config::DbReaderOptions;
+use timeseries::TimeSeriesDbReader;
 
 /// Dashboard queries to benchmark.
 const QUERIES: &[(&str, &str)] = &[
@@ -65,7 +68,7 @@ const QUERIES: &[(&str, &str)] = &[
 #[derive(Parser)]
 #[command(about = "Benchmark PromQL queries using TimeSeriesDbReader")]
 struct Args {
-    /// Path to ReaderConfig YAML file.
+    /// Path to benchmark config YAML file.
     #[arg(long, default_value = "OpenDataTimeSeries.yaml")]
     config: String,
 
@@ -78,6 +81,58 @@ struct Args {
     step_secs: u64,
 }
 
+#[derive(Deserialize)]
+struct BenchConfig {
+    storage: StorageConfig,
+    #[serde(
+        default = "default_reader_options",
+        deserialize_with = "deserialize_reader_options"
+    )]
+    reader: DbReaderOptions,
+    #[serde(default = "default_cache_capacity")]
+    cache_capacity: u64,
+}
+
+fn default_reader_options() -> DbReaderOptions {
+    DbReaderOptions {
+        skip_wal_replay: true,
+        ..DbReaderOptions::default()
+    }
+}
+
+fn default_cache_capacity() -> u64 {
+    50
+}
+
+fn deserialize_reader_options<'de, D>(
+    deserializer: D,
+) -> std::result::Result<DbReaderOptions, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let overrides = serde_yaml::Value::deserialize(deserializer)?;
+    let mut defaults =
+        serde_yaml::to_value(default_reader_options()).map_err(serde::de::Error::custom)?;
+    merge_yaml_value(&mut defaults, overrides);
+    serde_yaml::from_value(defaults).map_err(serde::de::Error::custom)
+}
+
+fn merge_yaml_value(base: &mut serde_yaml::Value, overrides: serde_yaml::Value) {
+    match (base, overrides) {
+        (serde_yaml::Value::Mapping(base_map), serde_yaml::Value::Mapping(overrides_map)) => {
+            for (key, value) in overrides_map {
+                match base_map.get_mut(&key) {
+                    Some(existing) => merge_yaml_value(existing, value),
+                    None => {
+                        base_map.insert(key, value);
+                    }
+                }
+            }
+        }
+        (base_slot, override_value) => *base_slot = override_value,
+    }
+}
+
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     tracing_subscriber::fmt::init();
@@ -85,9 +140,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let args = Args::parse();
 
     let config_str = std::fs::read_to_string(&args.config)?;
-    let config: ReaderConfig = serde_yaml::from_str(&config_str)?;
+    let config: BenchConfig = serde_yaml::from_str(&config_str)?;
 
-    let reader = TimeSeriesDbReader::open(config).await?;
+    let reader =
+        TimeSeriesDbReader::open(config.storage, config.reader, config.cache_capacity).await?;
 
     let now = SystemTime::now();
     let range_start = now - Duration::from_secs(args.range_secs);
--- a/timeseries/src/config.rs
+++ b/timeseries/src/config.rs
@@ -7,10 +7,8 @@
 use std::time::Duration;
 
 use common::StorageConfig;
-use serde::{Deserialize, Serialize};
-use serde_with::{DurationMilliSeconds, serde_as};
 
-/// Configuration for opening a [`TimeSeriesDb`](crate::TimeSeriesDb) database.
+/// Configuration for opening a [`TimeSeriesDb`](crate::timeseries::TimeSeriesDb) database.
 ///
 /// This struct holds all the settings needed to initialize a time series
 /// instance, including storage backend configuration and operational parameters.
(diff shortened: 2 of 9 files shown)


apurvam:
Hi @b3ll9trix! Thanks for the patch. Let me know when this is ready for review, and I'll take a look.

b3ll9trix:
@apurvam the PR is ready for review. a small headsup. this is my first contribution so appreciate any feedback.

b3ll9trix:
@apurvam, thanks so much for taking the time to review the PR. To address the issues pointed out, I have a few clarifying questions about the implementations:


1. > The first is structural. In particular, I think this should stay narrower in scope. The useful behavior here is timeseries-specific: allow Prometheus/timeseries config to express retention and route that through to SlateDB’s default TTL when the DB is opened. I don’t think that requires adding a new ttl field to shared opendata-common SlateDbStorageConfig.

<details>
<summary> Clarifying Details </summary>
I agree that this has to be scoped more narrowly. I in fact thought this is how the implementation would fare after reading the RFC, however, when creating the storage, only `PrometheusConfig::storage` is passed and propagated further, making the `RetentionConfig` on a level higher completely invisible down the line.
```rust
 let storage = StorageBuilder::new(&prometheus_config.storage)
```
This is already happening on the library side of the implementation, which already has `Config::Retention`, but when creating the storage, the config is omitted entirely, making it obsolete. 

```rust
#[derive(Debug, Clone)]
pub struct Config {
    /// Storage backend configuration.
    ///
    /// Determines where and how time series data is persisted. See [`StorageConfig`]
    /// for available options including in-memory and SlateDB backends.
    pub storage: StorageConfig,

    /// How often to flush data to durable storage.
    ///
    /// Data is buffered in memory and periodically flushed to the storage backend.
    /// Lower values provide better durability at the cost of write performance.
    pub flush_interval: Duration,

    /// Maximum age of data to retain.
    ///
    /// Data older than this duration may be automatically deleted during
    /// compaction. Set to `None` to retain data indefinitely.
    pub retention: Option<Duration>,
}
```
but here, the config never reaches the storage layer. Am I missing something here? If I am, can you please point me to the part of the code?

```rust
pub async fn open(config: Config) -> Result<Self> {
        let storage = StorageBuilder::new(&config.storage)
            .await?
            .with_semantics(
                StorageSemantics::new().with_merge_operator(Arc::new(OpenTsdbMergeOperator)),
            )
            .build()
            .await?;
        let tsdb = Tsdb::new(storage);
        Ok(Self { tsdb })
    }
```

This gave me two ways to maintain abstraction
1. Introduce a new `ttl` which reflects the value configured in `PrometheusConfig::Retention` of `Config::Retention` by applying the retention onto the `StorageConfig` here

// Library
```rust
pub async fn open(config: Config) -> Result<Self> {
        // Apply retention config
        let config = config.apply_retention();
        let storage = StorageBuilder::new(&config.storage)
.
.
```
// Server

```rust
  // Apply retention TTL to storage config
async fn main() {
.
.
    let prometheus_config = prometheus_config.apply_retention();

    // Create storage based on configuration
    tracing::info!(
        "Creating storage with config: {:?}",
        prometheus_config.storage
    );
    let merge_operator = Arc::new(OpenTsdbMergeOperator);
    let storage = StorageBuilder::new(&prometheus_config.storage)
        .await
        .unwrap_or_else(|e| {
            tracing::error!("Failed to create storage: {}", e);
            std::process::exit(1);
        })
.
.
}
```


2. Introduce `RetentionConfig` directly on `SlateDbStorageConfig`. 

I went with 1 because `RetentionConfig` is a property of `tsdb`, so making it a property of `SlateDbStorageConfig` enforces a property at a higher abstraction on the underlying SlateDB. But if you think we can handle this better. I am happy to do it differently.

</details>


   
2. > That last point is probably the most important: the default 7d timeseries retention path needs to be careful not to clobber an explicitly configured serverside storage TTL (eg. a slatedb ttl configured in settings.yaml on the server) .

<details>
<summary> Clarifying Details </summary>
I want to reiterate the order of precedence and confirm the priority of defaults (with 1 being the highest priority). 

1. RetentionConfig in both library/server.
2. `default_ttl` configured through `Settings`
3. Defaults to 7d.  

</details>


I will start with 2 while I wait for some clarification on 1. Once again, thanks for the review.

apurvam:
Hi @b3ll9trix . thanks for the detailed response. I couldn't get to this today, but will try to respond by tomorrow.

apurvam:
Hi @b3ll9trix, thanks for the detailed follow-up. You are not missing anything: as the code is currently structured, Config::retention / PrometheusConfig::retention does not reach the storage layer by itself, because storage creation only sees `config.storage`. So the only way to pass the ttl setting through today is by updating `StorageConfig`, as your PR does.

The more I think about this, I think it's best for OpenData to let slatedb configurations happen directly via the SlateDB settings as much as possible. This is aligned with the spirit of the project: your single storage engine should present a single way to be configured. It's this type of thing that will make OpenData DBs feel similar to operate. 

The tension is that,  for dbs like Timeseries, we are trying to implement established interfaces. In particular, we want our timeseries to look like prometheus, which means we want it to be configurable like prometheus.

I think the cleanest way to do that, is to allow to timeseries configs to set (or override) SlateDB configs as a special case. We can do this with a small refactor in StorageBuilder.

Concretely, I think StorageBuilder should keep the loaded SlateDB Settings around a little longer and expose a hook that lets the caller modify them before the final [DbBuilder::with_settings(...)](https://github.com/opendata-oss/opendata/blob/main/common/src/storage/factory.rs#L87) call. For example, something in the spirit of  `StorageBuilder.map_slatedb_settings(|settings| { ... })`.

The important part is that this would operate on the already-loaded Settings object, not replace it entirely. That gives us a clear merge model and precedence:

  1. DB-level config (RetentionConfig / Config::retention)
  2. SlateDB config loaded from settings_path / Settings
  3. SlateDB defaults

  So the flow would be:

  1. load SlateDB settings from file/env exactly as today
  2. overlay the DB-owned configs like retention setting onto settings.default_ttl if retention is explicitly configured
  3. pass the final merged Settings into SlateDB

That would preserve all existing SlateDB settings from the file, while allowing timeseries to override the one field it owns. 

I like this approach for two reasons:

  - SlateDB settings remain the primary way to configure SlateDB. But individual DBs have a way to override these settings.
  - it also makes the precedence explicit and testable

If this makes sense from a design perspective, I'm happy to work through the details with you. It's notable that we already have a [map_slateb](https://github.com/opendata-oss/opendata/blob/main/common/src/storage/factory.rs#L116) method that allws you to set any SlateDB knob that's not already exposed by the `StorageBuilder`. We just need to generalize that approach to make it more ergonomic.

 On the precedence/default question, I’d suggest testing the following explicitly:

  1. when retention is explicitly set in library/server config, it overrides Settings.default_ttl
  2. when retention is omitted, an existing default_ttl from SlateDB settings is preserved
  3. defaults do not silently clobber an explicitly configured storage TTL
  4. the final settings passed to SlateDB are the merged result of this precedence

b3ll9trix:
@apurvam, thanks for the detailed comment. I agree with the homogeneity of how different DBs build their storage engines; that is, through `Settings`. 

>The more I think about this, I think it's best for OpenData to let slatedb configurations happen directly via the SlateDB settings as much as possible. This is aligned with the spirit of the project: your single storage engine should present a single way to be configured. It's this type of thing that will make OpenData DBs feel similar to operate.

The current implementation already follows the plan you describe: DB-owned configs are set via `Settings` to SlateDB.
Don't you think? 

When I thought about precedence, I converged on the very point you mention: DB's (in this case `tsdb`) config, when explicitly configured (in this case `ttl`), overrides the undpinning SlateDB's config (in this case `default_ ttl`).


```rust
let settings = match &slate_config.settings_path {
    Some(path) => Settings::from_file(path).map_err(|e| {
        StorageError::Storage(format!(
            "Failed to load SlateDB settings from {}: {}",
            path, e
        ))
    })?,
    None => Settings::load().unwrap_or_default(),
};

// Apply `ttl` if configured. Overrides `default_ttl` from other sources.
let settings = if let Some(ttl) = slate_config.ttl {
    Settings {
        default_ttl: Some(ttl),
        ..settings
    }
} else {
    settings
};

info!(
    "create slatedb storage with config: {:?}, settings: {:?}",
    slate_config, settings
);

let mut db_builder =
    DbBuilder::new(slate_config.path.clone(), object_store).with_settings(settings);
```

Additionally, `ttl` still becomes necessary in `StorageConfig` because `StorageBuilder` knows only that.


```rust
impl StorageBuilder {
    /// Creates a new `StorageBuilder` from a [`StorageConfig`].
    ///
    /// For SlateDB configs this creates a [`DbBuilder`] with the configured
    /// path, object store, settings, and block cache (if configured). For
    /// InMemory configs it stores a sentinel so that `build()` returns an
    /// `InMemoryStorage`.
    pub async fn new(config: &StorageConfig) -> StorageResult<Self>
```


OR, are we trying to essentially defer from adding `ttl` (a DB-owned config) onto our `StorageConfig` because it adds an extra (mid-)level of abstraction between `PrometheusConfig` and `slatedb::config::Settings`? Although it is worth noting that the latter reduces the impact of this change (not having to accommodate `StorageConfig::ttl` in every place it is instantiated). 

Precisely,
1. (New Impl) Adding a hook and, by that, directly pipelining retention from `PrometheusConfig` to `Settings`, overlaying the existing `Settings`.  `PrometheusConfig.retention` → hook → `Settings`
OR
2. (Curr Impl) Carrying over the retention from `PrometheusConfig` to `StorageConfig`, which finally overlays the existing SlateDB `Settings`. `PrometheusConfig.retention` → `StorageConfig.ttl` → `Settings`

Either way, the DB-owned configs ultimately land in Settings; it's just a question of where the mapping happens.

apurvam:
Thanks for your detailed note @b3ll9trix . Let's keep the conversation here, rather than discord. 

If we do enable you to configure TTL via prometheus config, I'd much prefer 1). We needn't bloat the `StorageConfig` for a setting which basically already exists like for like in slatedb settings. 

I know we had issue #20 to implement retention in timeseries. But as we've discussed, we already can set retention in timeseries via slatedb. 

So then the question becomes, why do we need set a DB-wide TTL via the prometheus config? Is there some advantage of exposing TTL as a prometheus config to a timeseries user? If there is no strong product reason to expose this TTL config as a prometheus config, I think it's actually better to not add it IMO. OpenData as a project should prefer things configured in SlateDB as much as possible. 

An example of where I think TTL at the prometheus level would be interesting if Prometheus allowed you to set it at a granular level (eg if all metrics with label=foo had a TTL of N minutes or something). Such a granular prometheus setting would need to be converted the SlateDB model to take effect. Such a conversion layer would be truly complementary to what SlateDB provides, and thus would be worth adding according to the rationale laid out in this comment.

@agavra WDYT of that framing?

agavra:
+1 to @apurvam's framing of the problem. I think having a db-wide TTL via prometheus config likely adds cognitive load to users that's unnecessary. If users understand that slate has a TTL config, then they need to understand the relationship between the two and what takes precedence (and whether they are, in fact, the same thing). 

If there's something net new (e.g. a metric-level TTL) then it makes sense to have that in the timeseries database.

apurvam:
Hi @b3ll9trix , it turns out we need to set record level TTL using SlateDB's new (not yet released) [ExpireAt option](https://github.com/slatedb/slatedb/pull/1450).

Basically, we need to keep the expire time for a particular series in a particular bucket constant, because otherwise samples for a series in a bucket can't be merged because every one of them will have a separate expiry time by default (because today, expiry time is based on the time of the write).

This causes massive read amplification and high cold latencies, as documented here: https://github.com/opendata-oss/opendata/issues/338#issuecomment-4145276222

This change is in the same spirit as what you were doing, but will truly need to be timeseries specific and be piped through the timeseries write path.

If you are interested in picking it up, I can work with you to scope it out. 

Let me know.

b3ll9trix:
Hi @apurvam, yes, sure, I would like to pick it up. Can you assign it to me?

apurvam:
Great, thanks! i'll create an issue and assign it to you. WE can discuss the details there. I'll get around to it in an hour or so.

apurvam:
@b3ll9trix , here's the issue: https://github.com/opendata-oss/opendata/issues/342

Once you comment on it, I can assign it to you.

PR #309, opened 2026-03-11
[common,timeseries] Add FoyerHybridCache option for SlateDB Block Cache.

apurvam:
## Summary

Adds support for a hybrid (in-memory + on-disk) block cache using foyer's HybridCache, exposed through SlateDB's FoyerHybridCache. When configured, hot SST blocks are served from RAM and evicted blocks spill to local disk (ideally NVMe), avoiding S3 reads for cached data.

The HybridBlockCache is particularly useful for timeseries since we expect the inverted index and forward index blocks are super hot and thus in memory, with the long tail of series blocks spilling to disk and served quickly on dashboard reloads.

## Test Plan

Added relevant unit tests. Mostly just a config framework update. 

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/config.rs
+++ b/timeseries/src/promql/config.rs
@@ -7,8 +7,6 @@ use std::time::Duration;
 use common::storage::config::StorageConfig;
 use serde::Deserialize;
 
-use crate::config::BlockCacheConfig;
-
 use crate::util::Result;
 
 #[cfg(feature = "http-server")]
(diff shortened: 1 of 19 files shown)


timeseries/src/promql/config.rs
```diff
@@ -42,6 +44,9 @@ pub struct PrometheusConfig {
     /// Defaults to 5 seconds.
     #[serde(default = "default_flush_interval_secs")]
     pub flush_interval_secs: u64,
+    /// Optional hybrid block cache (in-memory + on-disk).
+    #[serde(default)]
+    pub block_cache: Option<BlockCacheConfig>,
```
rodesai: This feels out of place here since it's specific to slatedb storage. I think we should just make this configurable under `SlateDbStorageConfig` via some enum like:

```rust
pub enum BlockCacheConfig {
    FoyerHybrid(FoyerHybridCacheConfig)
}
```

I had added it as a runtime config because I think we'll eventually want to support sharing the cache across db instances, but in retrospect I think the right way to do that would be to have some stateful loader object that can initialize/store some shared stuff computed from config. For now we can just not support that.

apurvam: Updated.

rodesai (approved):
LGTM!

PR #312, opened 2026-03-11
adds vector reader/writer http servers

rodesai:
## Summary

- add RFC-0004 for vector http api
- adds reader and writer servers under vector/src/server
- adds server binaries under vector/src/bin/server.rs
- adds e2e integration test for server

## Test Plan

e2e integration test for the server

## Checklist

- [x] Tests added/updated


vector/rfcs/0004-http-apis.md
```diff
@@ -0,0 +1,380 @@
...
+
+message ErrorResponse {
+  string status = 1;
+  string message = 2;
+}
+```
+
+#### JSON Interoperability Note
+
+The JSON form intentionally flattens attribute values instead of using canonical ProtoJSON wrapper
```
apurvam, line 139: Is the motivation here to adhere to conventional json and maybe even the schemas used by popular vector dbs like turbopuffer? 

Put another way, if we were to use protoJson, then we would not be compatible with the conventions of the JSON APIs of other vectordbs?

rodesai: The intent wasn't really to be compatible with a specific API - more just ergonomics. I wanted it to be simple to interact with the db e.g. over curl.

PR #315, opened 2026-03-12
[timeseries] Optimize merge_batch_time_series with streaming k-way merge

agavra:
DISCLAIMER: this is a fully automated Claude loop asking it to optimize merges and run criterion benchmarks to prove it works, the result seems reasonable.

Replace collect-all → sort → dedup with a streaming k-way merge that exploits the fact that each input source is already sorted by timestamp. This eliminates the intermediate Vec allocation and O(n log n) sort, replacing them with an O(n·k) merge that deduplicates inline.

Benchmarked four scenarios using Criterion with up to 10K-point series and up to 20 operands. Each scenario isolates a
  different merge pattern that occurs during compaction:

```
  ┌──────────────┬────────────────────────────────────────────────────────────┬─────────────┐
  │   Scenario   │                        Description                         │   Result    │
  ├──────────────┼────────────────────────────────────────────────────────────┼─────────────┤
  │              │ Two series with no overlapping timestamps (e.g. series A   │ +4-8%       │
  │ Disjoint     │ covers [0s, 100s), series B covers [100s, 200s)). No dedup │ regression  │
  │              │  needed, pure interleave.                                  │             │
  ├──────────────┼────────────────────────────────────────────────────────────┼─────────────┤
  │ Fully        │ Two series covering the exact same timestamp range — every │             │
  │ overlapping  │  timestamp collides and must be deduped (last-write-wins). │ ~20% faster │
  │              │  Worst case for dedup pressure.                            │             │
  ├──────────────┼────────────────────────────────────────────────────────────┼─────────────┤
  │              │ Two series whose timestamps alternate (A has even seconds, │             │
  │ Interleaved  │  B has odd seconds). All points survive but every pick     │ ~10% faster │
  │              │ requires comparing both heads.                             │             │
  ├──────────────┼────────────────────────────────────────────────────────────┼─────────────┤
  │ Many         │ 2 to 20 sources all covering the same timestamp range,     │ 23-62%      │
  │ operands     │ simulating a compaction that merges many L0 SST files at   │ faster      │
  │              │ once. This is the most realistic compaction workload.      │             │
  └──────────────┴────────────────────────────────────────────────────────────┴─────────────┘
```

### DIFF
--- a/timeseries/src/serde/timeseries.rs
+++ b/timeseries/src/serde/timeseries.rs
@@ -284,11 +284,9 @@ pub(crate) fn merge_batch_time_series(
     loop {
         // Find the minimum timestamp among all heads
         let mut min_ts = i64::MAX;
-        for head in &heads {
-            if let Some(s) = head {
-                if s.timestamp_ms < min_ts {
-                    min_ts = s.timestamp_ms;
-                }
+        for s in heads.iter().flatten() {
+            if s.timestamp_ms < min_ts {
+                min_ts = s.timestamp_ms;
             }
         }
         if min_ts == i64::MAX {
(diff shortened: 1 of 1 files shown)


agavra:
I closed this because there's a pretty bad bug, I confirmed that the heads aren't always in timestamp sorted order.

PR #316, opened 2026-03-12
Expose access to DbBuilder

hachikuji:
For performance testing, it's useful to have access to `DbBuilder` to tweak the runtime behavior of SlateDb. Exposing each knob through opendata Storage constructs (such as `StorageRuntime`) is tedious. To that end, I've created a `StorageBuilder` which wraps `DbBuilder` and exposes `map_slatedb` for overrides. I've also removed `StorageRuntime`.


hachikuji:
@apurvam Thanks. Once we're a bit further along, we can consider dialing this back. For now, it's pretty useful to be able to tune SlateDb without all the forwarding.

PR #319, opened 2026-03-13
[timeseries] [do not review] Miscellaneous optimizations to query path.

apurvam:
## Summary

  This branch is a sequence of mostly-isolated timeseries query-path optimizations. Each perf commit was benchmarked independently so reviewers can decide whether the code complexity is justified patch-by-patch instead of only looking at the final  combined result. Cumulatively, these changes bring an 85% latency reduction to the test queries relative to main. 

Also note: by lines, most of the code is test code. Each perf improvement is its own commit which is summarized in this description. The tests and other miscellaneous improvements are separate commits. 

The passing existing tests are the only validation that the changes don't cause a functional regression. 

  ## Benchmark Queries

The benchmark names used below refer to these exact PromQL queries:

* process_rate: `sum by (applicationId) (diagnoser_latency_expected_node_commit_rate)`
* thread_saturation: `max by (applicationid, host_name) (diagnoser_saturation_blocked_ratio)` 
* origin_events_2d: `sum by (topic) ((max by (topic, partition) (responsive_kafka_streams_source_offset_end{applicationid="app_1"})) - (max by (topic, partition) (responsive_kafka_streams_source_offset_end{applicationid="app_1} offset 2d))) / 2`

Benchmark setup:
* Query time range: 6h
* Step: 30s
* Reader config: FoyerHybrid block cache with warm-cache runs unless otherwise noted

  ## Commit-by-commit perf deltas

  | Commit | Change | Measured improvement | Notes |
  | --- | --- | --- | --- |
  | 08a67b1 | Cache shared labels and precomputed series fingerprints | `process_rate`: `1213.0ms -> 678.0ms` (`-44.1%`) |  Large win; removes repeated label sorting/cloning/fingerprinting |
  | 642decc | Reuse cached sample vectors for instant lookups | `process_rate`: `678ms -> 661ms` (`~ -2.5%`) | Small general cleanup |
  | e4577e7 | Fast-path reduction aggregates over vector selectors | `process_rate`: `661ms -> 325.1ms` (`-50.8%`) | Large win; removes substantial evaluator overhead |
  | c99bcdd | Materialize cached forward and inverted query views | `process_rate`: `325.1ms -> 322.0ms` (`-1.0%`) | Small measurable improvement |
  | cc2ab28 | Avoid set churn for simple reader selectors | `process_rate`: `322.0ms -> 303.0ms` (`-5.9%`) | Modest win |
  | 5a06d36 | Use an integer hasher for fingerprint-keyed evaluator state | `process_rate`: `303.0ms -> 270.4ms` (`-10.8%`) |  General fingerprint-keyed map/set improvement |
  | 64f1073  | Cache requested forward index entries | `process_rate`: `270.4ms -> 273.7ms` (`+1.2%`); `thread_saturation`: `809.3ms -> 761.2ms` (`-5.9%`); `origin_events_2d`: `491.9ms -> 379.0ms` (`-22.9%`) | Worth keeping if we care about higher-  cardinality queries; flat on `process_rate` |
  | 615467e | Use jemalloc as the global allocator | `process_rate`: `295.2ms -> 240.9ms` (`-18.4%`); `thread_saturation`:  `919.9ms -> 687.0ms` (`-25.3%`); `origin_events_2d`: `379.3ms -> 271.6ms` (`-28.4%`) | Allocator A/B comparison on otherwise- equivalent code, not strict previous-commit delta |

  ## Supporting / non-perf-isolated commits

  ### 49ccf66 Update configs to pass refresh interval and checkpoint lifetime through to reader

  - Plumbing change so benchmark runs use the intended long-lived reader config.
  - Important for benchmark correctness and avoiding confounding SlateDB refresh behavior.
  - Not a standalone perf optimization.

  ### 65afe0f timeseries: load query reader buckets in parallel

  - Parallelizes bucket loading in Tsdb::query_reader / query_reader_for_ranges.
  - Intended to reduce latency for cold or partially-cached multi-bucket queries.
  - This change brings `TimeseriesDb` in parity with `TimeseriesDbReader`

  ### 400e635, f98a089, b02cc22

  - Regression fixes and tests for the perf changes above.
  - These are correctness/supporting commits, not new perf claims.

  ## Reviewer guidance

  If you want to evaluate the branch conservatively, I’d suggest reviewing the patches in this order:

  1. 08a67b1 and e4577e7
     These are the highest-impact wins by far.
  2. 5a06d36 and 615467e
     These are broad, general improvements with clear measured benefit.
  3. 64f1073
     Worth keeping if we care about higher-cardinality queries, even though it is flat on process_rate.
  4. 642decc, c99bcdd, cc2ab28, 65afe0f
     These are smaller or more situational wins and should be judged more critically on complexity vs benefit.

  ## Testing

  - Added regression tests covering forward-index cache reuse and multi-bucket query correctness.
  - Benchmarked with heracles/perftest/opendata using warm-cache repeated runs.
  - Detailed benchmark notes are in timeseries/PERF_PROCESS_RATE.md.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -2297,7 +2297,19 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
         let end_ms = adjusted_eval_ts.as_millis();
         let start_ms = end_ms - lookback_delta_ms;
 
-        let mut buckets = self.reader.list_buckets().await?;
+        // Only check buckets whose time range overlaps the lookback window
+        // [start_ms, end_ms]. This is typically 1-2 buckets instead of all
+        // buckets in the query range.
+        let all_buckets = self.reader.list_buckets().await?;
+        let mut buckets: Vec<_> = all_buckets
+            .into_iter()
+            .filter(|b| {
+                let bucket_start_ms = b.start as i64 * 60_000;
+                let bucket_end_ms = (b.start as i64 + b.size_in_mins() as i64) * 60_000;
+                // Overlap: bucket_end > start_ms AND bucket_start <= end_ms
+                bucket_end_ms > start_ms && bucket_start_ms <= end_ms
+            })
+            .collect();
         buckets.sort_by(|a, b| b.start.cmp(&a.start));
 
         let mut series_with_results = FingerprintHashSet::default();
@@ -2320,7 +2332,7 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
                 .await?;
 
             for (series_id, series_spec) in forward_index_entries.iter().cloned() {
-                if !series_with_results.insert(series_spec.fingerprint) {
+                if series_with_results.contains(&series_spec.fingerprint) {
                     continue;
                 }
 
@@ -2331,6 +2343,8 @@ impl<'reader, R: QueryReader> Evaluator<'reader, R> {
                 else {
                     continue;
                 };
+
+                series_with_results.insert(series_spec.fingerprint);
 
                 let group_key = Self::grouping_key_from_labels(
                     &series_spec.labels,
@@ -2623,7 +2637,9 @@ mod tests {
     fn setup_mock_reader(
         data: TestSampleData,
     ) -> (crate::query::test_utils::MockQueryReader, SystemTime) {
-        let bucket = TimeBucket::hour(1000);
+        // Bucket must cover the sample timestamps (~300,001ms) so that
+        // bucket-scoped filtering in the reduction aggregate path works.
+        let bucket = TimeBucket::hour(0);
         let mut builder = MockQueryReaderBuilder::new(bucket);
 
         // Base timestamp: 300001ms (ensures samples are > start_ms with 5min lookback)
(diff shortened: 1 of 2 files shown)


ccollie:
One possible perf enhancement: In  [Evaluator::evaluate_binary_expr](https://github.com/opendata-oss/opendata/blob/bf17b2947c7d3a961b8457365b93cde03963e2fe/timeseries/src/promql/evaluator.rs#L1612)   temporary vector values are modified and cloned, even when the values are owned and can be modified in-place. 

Here's a possible refactor:

```rust
    let mut result = vector;
    result.retain_mut(|sample| {
        match self.apply_binary_op(expr.op, scalar, sample.value) {
            Ok(value) => {
                // For comparison ops without bool, filter out false results
                if is_comparison && !return_bool && value == 0.0 {
                    false
                } else {
                    sample.value = value;
                    sample.drop_name |= self.changes_metric_schema(expr.op) || return_bool;
                    true
                }
            }
            Err(_) => false,
        }
    });
    Ok(ExprResult::InstantVector(result))
```

I'm not sure how performant `retain_mut` is, but this at least saves an allocation

apurvam:
Don't review this yet. I'm also optimizing cold query latency on a much more realistic setup. I'll probably issue a combined Pr whose effect is to improve grafana dashboard loading times in a practical way. Specifically: cold loads should be reasonably fast, and refreshes for new data should feel instant. This way you can explore the entire time range without waiting 10s of seconds each time you switch to non overlapping time ranges.

apurvam:
Closing in favor of the work done in the #344 lineage.

Issue #202, opened 2026-02-13
Handle Lagged error in the view subscriber channel


hachikuji:
In the channel between writers and readers, we do not currently handle the `Lagged` error: https://github.com/opendata-oss/opendata/blob/main/common/src/coordinator/subscriber.rs#L122. When we receive this error, we may have missed messages, which means the reader state may become inconsistent. We can handle this by resetting the state using the provided snapshot.

KyleSmith19091:
Was thinking about picking up this issue, cause it seemed like an interesting problem. I just have some questions.

- Note: When referring to 'view' I'm referring to the type defined in common/src/coordinator/handle.rs:26

I'm probably missing something (maybe this is mentioned in the RFC 0005, but I missed it), but handling the error only seems applicable to readers that need to be able to record a strong consistent state of the view. Since even if the channel lags behind (and it's silently ignored) it provides an eventually consistent view for the subscribed readers, by eventually consistent I mean it's acceptable that the data might be a bit stale for a while till it reads the latest snapshot from the channel. Since as I understand it the View is the complete snapshot of the contents in the storage at some epoch. 

However if a reader needs to read *every view sent in the channel there has to be some higher level component that needs to coordinate it and probably needs to be configurable as this is not applicable to all readers I would imagine. Furthermore, if that is the case then you also need to consider messages being lost, for example the scenario where the process crashes for some reason then the channel can't be recovered unless the channel itself is persisted. Almost seems like the write coordinator should instead use something like what is being built in this PR: https://github.com/opendata-oss/opendata/pull/253. That way when the write coordinator starts up again it can replay the views to the readers, the readers would then just need to be idempotent or I guess use the epochs to track the views they have processed.

Sorry if I completely misunderstood what is being proposed

hachikuji:
@KyleSmith19091 Hey, thanks for the interest. My understanding of the `Lagged` error in this context is that it implies messages have already been dropped. The problem is that the view contains a delta of changes. So if a subscriber is applying the delta changes on top of an older state, the result would be inconsistent. For a reader which is ignoring the change delta, and just using the snapshot, no harm done. They just skipped over some changes, but still an (eventually) consistent state as you say. Does that make sense?

We track changes through epochs, so it should be possible for a reader to use the epoch corresponding to the view to tell whether a delta can be safely applied. One way to handle it is to discard in-memory state and rebuild it from the snapshot.

KyleSmith19091:
[...]

PR #324, opened 2026-03-15
Handle lagged error

KyleSmith19091:
## Summary

Handle the lagged error in the view snapshot subscription. If the subscriber channel reaches its configured capacity the `recv` method will return a `SubscribeError::Lagged` which the subscriber can use to determine how it should respond.

## Related Issues

https://github.com/opendata-oss/opendata/issues/202

## Test Plan

- Create a write coordinator instance.
- Subscribe to coordinator
- Write 20 items > 16 configured channel capacity
- Then attempt to call `recv` method
- Expect the lagged error

## Checklist

- [x] Tests added/updated


KyleSmith19091:
@hachikuji I know we originally talked about implementing the logic for recovery(https://github.com/opendata-oss/opendata/issues/202).  I decided to not implement it, since there are currently not any uses of the `subscribe` method so not sure if that logic needs to be implemented by the clients or if it needs to live in the DB implementations. My current implementation assumes that the subscriber can just recover using the view and snapshot.

hachikuji:
@KyleSmith19091 Thanks for the patch! Didn't quite get to it today, but I'll look in the morning.

common/src/coordinator/subscriber.rs
```diff
@@ -113,23 +130,10 @@ impl<D: Delta> ViewSubscriber<D> {
...
-                    // get a fresh initial view to reset state from.
-                    continue;
-                }
-                Err(broadcast::error::RecvError::Closed) => {
-                    return Err(SubscribeError::Shutdown);
-                }
-            }
+        match self.view_rx.recv().await {
+            Ok(view) => Ok(view),
+            Err(broadcast::error::RecvError::Lagged(_)) => Err(SubscribeError::Lagged), // the subscriber is lagging behind the coordinator's progress.
```
hachikuji: Wonder if we should rename Lagged. We can leave it if this is a conventional term in rust, but it surprised me that it implied message loss. Perhaps we could just say MessageLost or something like that.

common/src/coordinator/mod.rs
```diff
@@ -2311,6 +2311,76 @@ mod tests {
...
+                    size: 10,
+                })
+                .await
+                .unwrap();
+            let _ = handle.flush(false).await.unwrap();
+
+            let result = subscriber.recv().await;
+            assert!(result.is_ok());
+        } else {
+            // If the subscriber didn't lag (possible on a fast machine), that's ok
```
hachikuji: We may be able to make this test deterministic with this flow:

1. Loop send the messages and flush(false) as in the current patch
2. Send flush(true) to force the durable watermark to get updated
3. Wait for the durable watermark 
4. Call recv()

I think that would ensure the flushes have all been sent to the broadcast channel.

common/src/coordinator/subscriber.rs
KyleSmith19091: Yeah I think MessageLost is better, don't see a reason to keep the Rust nomenclature in the custom Error.

common/src/coordinator/mod.rs
KyleSmith19091: Yes I agree, I refactored it

common/src/coordinator/subscriber.rs
```diff
@@ -25,9 +25,23 @@
...
-//!     while let Ok(view) = subscriber.recv().await {
-//!         // Process the view (update read state)...
-//!         subscriber.update_durable(epoch);
+//!     loop {
+//!         match subscriber.recv().await {
+//!             Ok(view) => {
+//!                 // Process the view (update read state)...
+//!                 subscriber.update_durable(epoch);
+//!             }
+//!             Err(SubscribeError::Lagged) => {
```
hachikuji:
```suggestion
//!             Err(SubscribeError::MessageLost) => {
```

KyleSmith19091:
Awesome thank you

PR #326, opened 2026-03-18
update common key prefix and add subsystem discriminator

cadonna:
## Summary

This PR adds a subsystem byte to the shared key prefix and standardizes the layout as 
```
subsystem | version | record_tag
```

Updates the common prefix API and adds a subsystem discriminator. 
Adapts the subsystem crates to the new layout and byte offsets. 
Updates "RFC 0001: Record Key Prefix".

## Related Issues

Fixes #

## Test Plan

Added unit tests where needed

## Checklist

- [x] Tests added/updated

🤖 co-authored by Codex (gpt-5.4)

### DIFF
--- a/rfcs/0001-record-key-prefix.md
+++ b/rfcs/0001-record-key-prefix.md
@@ -54,23 +54,22 @@ The 3-byte prefix is followed by record-specific fields that vary by subsystem a
 
 ### Subsystem Byte
 
-The second byte identifies the key format version. Each subsystem manages its version independently—bumping the version in one subsystem does not affect others.
+The first byte identifies the subsystem that owns the key.
 
-| Value  | Description |
-|--------|-------------|
-| `0x00` | Reserved (invalid) |
-| `0x01`–`0xFF` | Subsystem-defined |
+| Value         | Description        |
+|---------------|--------------------|
+| `0x00`        | Reserved (invalid) |
+| `0x01`        | Timeseries         |
+| `0x02`        | Vector             |
+| `0x03`        | Log                |
+| `0x04`        | KeyValue           |
+| `0x01`–`0xFF` | Subsystem-defined  |
 
-Current subsystem allocations:
-
-- `0x01` Timeseries
-- `0x02` Vector
-- `0x03` Log
-- `0x04` KeyValue
 
 ### Version Byte
 
-The first byte identifies the subsystem that owns the key.
+The second byte identifies the key format version.
+Each subsystem manages its version independently—bumping the version in one subsystem does not affect others.
 
 | Value  | Description |
 |--------|-------------|
@@ -78,7 +77,9 @@ The first byte identifies the subsystem that owns the key.
 | `0x01` | Current version (all subsystems) |
 | `0x02`–`0xFF` | Reserved for future versions |
 
-The version byte enables schema migration. When the key format changes in a backwards-incompatible way, the subsystem increments its version. Readers encountering an unknown version can reject the record or apply version-specific parsing logic.
+The version byte enables schema migration.
+When the key format changes in a backwards-incompatible way, the subsystem increments its version.
+Readers encountering an unknown version can reject the record or apply version-specific parsing logic.
 
 **Guidelines for version changes:**
 - Additive changes to value schemas do not require a version bump


log/src/serde.rs
```diff
@@ -55,8 +56,11 @@ impl From<common::serde::DeserializeError> for Error {
 /// Key format version (currently 0x01)
 pub const KEY_VERSION: u8 = 0x01;
 
+/// Subsystem-specific key prefix for log storage.
+pub const SUBSYSTEM: u8 = 0x03;
+
 /// Storage key for the SeqBlock record.
-pub const SEQ_BLOCK_KEY: [u8; 2] = [KEY_VERSION, 0x02]; // RecordType::SeqBlock
+pub const SEQ_BLOCK_KEY: [u8; 3] = [SUBSYSTEM, KEY_VERSION, 0x20]; // RecordType::SeqBlock
```
cadonna, line 63: @hachikuji I think there was a bug in this constant or at least an inconsistency. The `RecordType` implementation moves the record type ID into the higher 4 bits, but this constant keeps it in the lower 4 bits. 
Let me know, if you want to keep `0x02` or change it to `0x20`.

hachikuji: Yep, the change makes sense.

common/src/serde/record_tag.rs
```diff
@@ -0,0 +1,162 @@
...
+/// // Create a tag with type 0x05 and reserved bits 0x03
+/// let tag = RecordTag::new(0x05, 0x03);
+/// assert_eq!(tag.as_byte(), 0x53);
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct RecordTag(u8);
+
+impl RecordTag {
+    /// Creates a new record tag with the given record type and reserved bits.
+    pub fn new(record_type: u8, reserved: u8) -> Self {
```
hachikuji, line 32: nit: we lose some validations when this moved

log/src/serde.rs
```diff
@@ -114,7 +129,7 @@ impl RecordType {
 /// that preserves lexicographic ordering:
 ///
 /// ```text
-/// | version (u8) | type (u8) | segment_id (u32 BE) | terminated_key | relative_seq (var_u64) |
+/// | subsystem (u8) | version (u8) | type (u8) | segment_id (u32 BE) | terminated_key | relative_seq (var_u64) |
```
hachikuji, line 132: I think the only interesting question here is whether we want to allow the same system to coexist with itself in the same keyspace. I can't really think of a situation where we need this, so I'm inclined to defer it.

rfcs/0001-record-key-prefix.md
```diff
@@ -41,17 +44,33 @@ Formalizing this pattern as a common specification provides several benefits:
...
+┌───────────┬─────────┬────────────┬─────────────────────┐
+│ subsystem │ version │ record_tag │  ... record fields  │
+│  1 byte   │ 1 byte  │   1 byte   │    (varies)         │
+└───────────┴─────────┴────────────┴─────────────────────┘
 ```
 
-The 2-byte prefix is followed by record-specific fields that vary by subsystem and record type.
+The 3-byte prefix is followed by record-specific fields that vary by subsystem and record type.
+
+### Subsystem Byte
```
hachikuji, line 55: This section needs to be swapped. Below we're talking about the key format version.

common/src/serde/key_prefix.rs
```diff
@@ -398,55 +298,54 @@ mod tests {
...
-        assert_eq!(prefix.tag().record_type(), 0x02);
-        assert_eq!(prefix.tag().reserved(), 0x05);
+        assert_eq!(prefix.tag(), 0x25);
     }
 
     #[test]
     fn should_reject_wrong_version() {
         // given
-        let data = [0x02, 0x10]; // wrong version
+        let data = [0x10, 0x02, 0x10]; // wrong version
```
hachikuji, line 364: This is modifying the subsystem, not the version. Maybe both cases are useful?

cadonna: This test verifies that the key prefix has the expected version.

What do you mean with "modifying the subsystem"?

I agree that I should add the verification of the subsystem, too. Although, currently we only validate records that we retrieved through scans and gets which specify the complete key prefix including subsystem, version, and record type. That means the validation can never fail. However, since the deserialization code is generic it makes sense to also add this validations as long as they are cheap.

cadonna: Done in 5833f1ede973e8c9536f1625c1126e46ef0fc450

common/src/serde/record_tag.rs
cadonna: Good catch! 
I missed that. Re-added in 1372b3e.

log/src/serde.rs
cadonna: Were you considering a namespace or instance ID when you wrote this comment?

It should be fairly easy to migrate to a namespace or instance ID. So, I would also defer it to later when we have an actual use case for it.

rfcs/0001-record-key-prefix.md
cadonna: You are right! My review was too sloppy 😞 !

Done in 87134d8

PR #327, opened 2026-03-19
delete ingested data batches from object storage after dequeue

cadonna:
## Summary

After the collector dequeues entries from the queue manifest (via ack() every 100 batches or flush()), the batch files remain in object storage forever. Add a fire-and-forget tokio::spawn helper that deletes these files asynchronously after dequeue, logging warnings on failure without blocking batch processing.

## Test Plan

Unit tests

## Checklist

- [x] Tests added/updated

🤖  co-authored by Claude Code Opus 4.6

### DIFF
--- a/ingest/src/collector.rs
+++ b/ingest/src/collector.rs
@@ -2,8 +2,8 @@ use std::cell::Cell;
 use std::sync::Arc;
 
 use bytes::Bytes;
-use futures::stream;
 use futures::StreamExt;
+use futures::stream;
 use slatedb::object_store::ObjectStore;
 use slatedb::object_store::path::Path;
 


ingest/src/collector.rs
```diff
@@ -165,6 +167,24 @@ impl Collector {
...
 }
 
+fn delete_dequeued_batches(object_store: Arc<dyn ObjectStore>, entries: Vec<QueueEntry>) {
+    if entries.is_empty() {
+        return;
+    }
+    tokio::spawn(async move {
+        for entry in entries {
+            let path = Path::from(entry.location.as_str());
+            if let Err(e) = object_store.delete(&path).await {
```
agavra: is there a batch delete API? it seems like S3 has a `DeleteObjectsRequest` which can delete multiple in one call

cadonna:
> This makes sense to me for now, but we may want to open a follow-up ticket to have a proper async GC. There's a race condition where you dequeue and then fail before deleting batches, at which point (if I'm understanding this correctly) the entries will never be deleted.

What do you mean exactly with "proper async GC"? Do you mean an external process that cleans up data batches?

The issue you mention about the batches that are never deleted that can be solved by deleting all batches before a given sequence number which would be cheaper than an external process.

agavra:
> What do you mean exactly with "proper async GC"? Do you mean an external process that cleans up data batches?

It doesn't have to be an external processes, just a thread that's independent from the consumer is fine. Basically the ability to know what's "garbage" without coordination.

> The issue you mention about the batches that are never deleted that can be solved by deleting all batches before a given sequence number which would be cheaper than an external process.

That's effectively exactly what I meant 😄

ingest/src/collector.rs
cadonna: I changed it to use `delete_stream()` instead of delete. Probably not exactly what you were asking.

cadonna:
> That's effectively exactly what I meant 😄

I meant to just do that in the fire-and-forget task that I added in the collector.
I guess, you meant to have a background task that starts with the collector and does the garbage collection.

cadonna:
Issue regarding garbage collection: https://github.com/opendata-oss/opendata/issues/328

PR #329, opened 2026-03-23
implement a decoupled indexer to maintain ann index

rodesai:
## Summary

Implements a decoupled indexer for maintaining the ann index:

The basic idea here is that we add a new write path that changes how we maintain the vector index:
- The write coordinator delta just tracks recently applied writes in a Vec
- When frozen delta just makes a copy of these writes and passes an Arc of that to the views and flusher
- The flusher takes the entire write batch and updates the index in bulk.

Most of the work is now being done during flush to convert the raw batch of vector writes into a batch of updates to storage that maintain a good vector index. This transformation is done by a new component called Indexer.

The Indexer owns the writer's in-memory state, and is responsible for both maintaining it and computing the storage ops for a given write batch. The entry point to Indexer is update_index. update_index takes a batch of writes and the last applied storage snapshot, and is responsible for both returning the resulting storage ops and making any required mutations to the in-memory state (e.g. the centroid graph, dictionary, sequence tracker, etc). It works by initializing a indexer-specific delta called VectorIndexDelta, and then passing both the delta and the data (e.g. the write batch or resulting intermediate data) through various ops. Each op makes changes to the delta to represent the changes required for indexing like adding forward index entries, updating postings, adding/removing centroids, etc. Finally, once the round of indexing has completed, the delta is frozen which commits all the changes to the in-memory state and computes the resulting storage ops. This approach ensures we can gracefully handle i/o failures on the indexing path, as no mutations are actually applied until the full set of mutations is computed.

The Indexer runs through the following ops:
- WriteVectors - this op is responsible for taking the vectors in the write batch and making sure storage references them correctly in the forward index, postings, and inverted index. It's also responsible for deleting the state of updated vectors.
- MergeCentroids - this op is responsible for finding centroids with small clusters, deleting them, and returning their postings for reassignment.
- SplitCentroids - this op finds centroids with large clusters and splits them by running kmeans to find new centroids. It also finds vectors that need to be reassigned and returns them.
- ReassignVectors - this op takes vectors that may need to be reassigned and updates their postings if required.

For each batch, the indexer first runs WriteVectors to add the new vectors to the db. It then runs MergeCentroids to clear any small centroids. Finally, it runs SplitCentroids and ReassignCentroids in a loop until all centroids are below the threshold. We don't run merge after the initial merge for a batch as there is no guarantee that the process will converge. This is consistent with the proof from spfresh which assumes merges happen decoupled from chains of splits. It does diverge from the spfresh implementation which merges when reads detect small centroids, but I don't think this is a very important distinction.

Internally, each op may need to do i/o to read the state of the db. For example, the SplitCentroids op reads posting lists for centroids that are to be split and their neighbours. The basic pattern is to do all these i/os in batch first, and then schedule the compute work required for the op (e.g. for SplitCentroids running kmeans and figuring out what vectors need to be reassigned).

Also includes a couple tweaks to common modules
- allows freezing the sequence tracker so its latest state can be extracted/cloned to create a new in-progress instance when initializing a new index delta
- changes Flusher to take &mut self so it can update its internal state when applying flush

## Test Plan

- unit tests added
- ran the sift100K integration test with a db that uses this write path
- ran the cohere1M benchmark with a db that uses this write path. Ingest throughput is ~1K vectors/second on a 12-core node - equivalent to the previous write path.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/vector/src/batched/indexer/mod.rs
+++ b/vector/src/batched/indexer/mod.rs
@@ -1,0 +1,104 @@
+mod drivers;
+mod merge;
+mod split;
+pub(crate) mod state;
+#[cfg(test)]
+pub(crate) mod test_utils;
+mod vector;
+
+use crate::DistanceMetric;
+use crate::Result;
+use crate::batched::indexer::merge::MergeCentroids;
+use crate::batched::indexer::split::SplitCentroids;
+use crate::batched::indexer::state::{CentroidChunkManager, VectorIndexDelta, VectorIndexState};
+use crate::batched::indexer::vector::{ReassignVectors, WriteVectors};
+use crate::delta::VectorWrite;
+use crate::hnsw::CentroidGraph;
+use bytes::Bytes;
+use common::StorageRead;
+use common::sequence::AllocatedSeqBlock;
+use common::storage::RecordOp;
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
+use tracing::debug;
+
+const INDEXING_ROUNDS: usize = 1;
+
+pub(crate) struct IndexerOpts {
+    pub(crate) dimensions: usize,
+    pub(crate) distance_metric: DistanceMetric,
+    pub(crate) merge_threshold_vectors: usize,
+    pub(crate) split_threshold_vectors: usize,
+    pub(crate) split_search_neighbourhood: usize,
+    pub(crate) indexed_fields: HashSet<String>,
+    pub(crate) chunk_target: usize,
+}
+
+pub(crate) struct Indexer {
+    opts: Arc<IndexerOpts>,
+    state: VectorIndexState,
+}
+
+impl Indexer {
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn new(
+        opts: IndexerOpts,
+        dictionary: HashMap<String, u64>,
+        centroid_counts: HashMap<u64, u64>,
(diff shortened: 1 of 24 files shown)


vector/src/batched/indexer/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+use crate::hnsw::CentroidGraph;
+use bytes::Bytes;
+use common::StorageRead;
+use common::sequence::AllocatedSeqBlock;
+use common::storage::RecordOp;
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
+use tracing::debug;
+
+const INDEXING_ROUNDS: usize = 1;
```
rodesai, line 25: I'll probably clean this up at some point. I'm still playing with if it makes any difference on index quality if we force a few rounds of split-merge

vector/src/lib.rs
```diff
@@ -47,6 +47,8 @@ pub mod server;
 pub(crate) mod storage;
 pub(crate) mod view_reader;
 
+#[allow(dead_code)]
+mod batched;
```
rodesai, line 51: this is not a great name. I'll just rename it to `write` in a follow-up once this is the only write path

rodesai: https://github.com/opendata-oss/opendata/pull/332

PR #333, opened 2026-03-23
RFC 0006: Native Histogram Support

### LINKED ISSUE #102: [tsdb] add histogram support to tsdb
Prometheus treats histograms as a first-class citizen in its data model. The details of the semantics and implementation are laid out in the Prometheus docs here: https://prometheus.io/docs/specs/native_histograms/?utm_source=chatgpt.com. OpenData TSDB should support the same semantics for ingesting and querying histograms, and we can likely borrow quite a bit from the Prometheus implementation. This issue merits its own RFC, but I'm just creating this ticket for now to track that it needs to be done.

samsond:
Add RFC proposing Phase 1 native histogram support in the PromQL engine.

Phase 1 approves a narrow slice:
- Carry native histogram payloads through evaluator/function boundary
- Implement histogram_count, histogram_sum, histogram_avg projection functions
- Validate with unit tests and promqltest

Storage encoding (RecordType::Histogram), range semantics, HTTP serialization, and full quantile/fraction support are deferred to Phases 2-5.

Defines cross-RFC dependencies with RFC 0001 (storage), RFC 0003 (public types), and RFC 0005 (PromQL functions). Aligns with Prometheus evaluator FloatHistogram model for exponential-bucket native histograms.

Fixes #102 

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/timeseries/rfcs/0006-native-histogram-promql.md
+++ b/timeseries/rfcs/0006-native-histogram-promql.md
@@ -1,0 +1,188 @@
+# RFC 0006: Native Histogram Support in PromQL
+
+**Status**: Draft
+
+**Authors**:
+- [Samson Hailu](https://github.com/samsond)
+
+## Summary
+
+This RFC recommends approving a narrow Phase 1 of native histogram support in the PromQL engine: carry native histogram payloads through the evaluator/function boundary, support `histogram_count`, `histogram_sum`, and `histogram_avg`, and validate that slice with unit tests and `promqltest`.
+
+This RFC does **not** propose end-to-end native histogram support in the same change. Storage, ingestion, HTTP response serialization, broad range-function behavior, and full histogram-related annotation parity remain later phases.
+
+The intent is pragmatic: start with the smallest slice that is both useful and testable, prove the architecture at the evaluator boundary, and make the follow-on work explicit rather than letting scope expand implicitly.
+
+## Motivation
+
+OpenData's current PromQL evaluator and storage are float-only. Prometheus treats native histograms as first-class sample types with dedicated functions and HTTP API behaviors. This RFC adopts the smallest slice that proves the evaluator architecture before committing to storage and API changes: carry native histogram payloads through the evaluator/function boundary, support three projection functions, and validate with unit tests and promqltest. Prometheus already defines the reference semantics (native histogram spec, `histogram_count`/`histogram_sum`/`histogram_avg`, HTTP `"histogram"` fields), so OpenData avoids reinventing behavior. Phase 1 is narrow to reduce risk; storage, ingestion, HTTP serialization, and broader histogram semantics remain later phases. See [native histogram spec](https://prometheus.io/docs/specs/native_histograms/), [PromQL functions](https://prometheus.io/docs/prometheus/latest/querying/functions/), and [promtool unit testing](https://prometheus.io/docs/prometheus/latest/configuration/unit_testing_rules/).
+
+## Goals
+
+- Add a native histogram payload to the PromQL evaluator.
+- Support the smallest useful set of native histogram-aware functions: `histogram_count`, `histogram_sum`, and `histogram_avg`.
+- Keep existing float semantics unchanged for non-histogram queries.
+- Add a native histogram test path that proves the feature works at the current evaluator/function boundary.
+- Define a migration path from this minimal slice to broader native histogram support later.
+
+## Non-Goals
+
+- Classic histogram support or reconstruction.
+- Parity with Prometheus for histogram-aware functions beyond `histogram_count`, `histogram_sum`, and `histogram_avg` in one change (e.g., `rate`, `delta`, `increase`, aggregation behavior on histograms).
+- Full histogram-related annotation parity in the first implementation.
+- Finalizing a storage encoding for native histogram samples in the prototype slice.
+
+## Design
+
+### Native Histogram Data Shapes
+
+Phase 1 introduces a minimal native histogram payload sufficient for projection functions (`histogram_count`, `histogram_sum`, `histogram_avg`). The structure aligns with Prometheus's evaluator/runtime `FloatHistogram` model for exponential-bucket native histograms.
+
+**Minimal Phase 1 Shape (Runtime Representation)**:
+
+```rust
+// Phase 1: Minimal payload for instant functions
+pub struct NativeHistogram {
+    /// Schema parameter that determines bucket boundaries (-4..=8).
+    /// Larger schema = smaller buckets (higher resolution).
+    pub schema: i8,
+    
+    /// Total count of all observations across all buckets.
+    /// For float histograms, may contain fractional values.
+    pub count: f64,
+    
+    /// Sum of all observed values. Used for calculating average.
+    pub sum: f64,
+    
+    /// Zero threshold: observations within [-zero_threshold, +zero_threshold]
+    /// are placed in the zero bucket rather than positive/negative buckets.
+    pub zero_threshold: f64,
+    
+    /// Count of observations that fall within the zero threshold.
+    pub zero_count: f64,
+    
+    /// Positive bucket spans and counts (see Span definition below).
+    pub positive_spans: Vec<Span>,
+    pub positive_buckets: Vec<f64>,  // Phase 1: float counts; future: i64 deltas for integer histograms
+    
+    /// Negative bucket spans and counts. Negative buckets store absolute value
+    /// observations; bucket boundaries are expressed as absolute values.
+    pub negative_spans: Vec<Span>,
+    pub negative_buckets: Vec<f64>,  // Phase 1: float counts; future: i64 deltas for integer histograms
+}
+
+/// Span represents a range of consecutive non-empty buckets.
+/// Used to compactly encode sparse bucket layouts.
+pub struct Span {
+    /// For the first span: the starting bucket index (can be negative).
+    /// For subsequent spans: the gap to the previous span (must be positive).
+    pub offset: i32,
+    
+    /// Number of consecutive buckets in this span.
+    pub length: u32,
+}
+```
+
+Bucket boundaries are computed from `schema` and bucket index per Prometheus spec; the `Span` structure encodes which buckets are non-empty, avoiding storage cost of empty buckets.
+
+**Phase 1 Constraints**:
+
+- Both `positive_buckets` and `negative_buckets` store absolute counts (f64), not deltas.
+- No `counter_reset_hint` or reset tracking in Phase 1.
+- Buckets are sparse; empty buckets are not stored.
+- Integer histogram variant (with i64 delta encoding) is deferred to Phase 2+ storage design.
+
+### Decision and Scope
+
+**Approved for Phase 1**:
+- Native histogram payload support at the evaluator/function boundary
+- `histogram_count`, `histogram_sum`, and `histogram_avg` functions
+- Native histogram test coverage via promqltest
+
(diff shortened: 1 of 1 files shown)


timeseries/rfcs/0006-native-histogram-promql.md
```diff
@@ -0,0 +1,188 @@
...
+**Status**: Draft
+
+**Authors**:
+- [Samson Hailu](https://github.com/samsond)
+
+## Summary
+
+This RFC recommends approving a narrow Phase 1 of native histogram support in the PromQL engine: carry native histogram payloads through the evaluator/function boundary, support `histogram_count`, `histogram_sum`, and `histogram_avg`, and validate that slice with unit tests and `promqltest`.
+
+This RFC does **not** propose end-to-end native histogram support in the same change. Storage, ingestion, HTTP response serialization, broad range-function behavior, and full histogram-related annotation parity remain later phases.
```
agavra, line 12: I think it's OK to have a phased implementation for Histogram support, but I'd like to see an RFC that outlines the entire plan so we have the picture of what supporting it in full is going to be.

timeseries/rfcs/0006-native-histogram-promql.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Motivation
+
+OpenData's current PromQL evaluator and storage are float-only. Prometheus treats native histograms as first-class sample types with dedicated functions and HTTP API behaviors. This RFC adopts the smallest slice that proves the evaluator architecture before committing to storage and API changes: carry native histogram payloads through the evaluator/function boundary, support three projection functions, and validate with unit tests and promqltest. Prometheus already defines the reference semantics (native histogram spec, `histogram_count`/`histogram_sum`/`histogram_avg`, HTTP `"histogram"` fields), so OpenData avoids reinventing behavior. Phase 1 is narrow to reduce risk; storage, ingestion, HTTP serialization, and broader histogram semantics remain later phases. See [native histogram spec](https://prometheus.io/docs/specs/native_histograms/), [PromQL functions](https://prometheus.io/docs/prometheus/latest/querying/functions/), and [promtool unit testing](https://prometheus.io/docs/prometheus/latest/configuration/unit_testing_rules/).
```
agavra, line 18: as mentioned on the ticket, we should understand the different options available here. it seems like OTEL has a slightly different representation of histograms, and so does VictoriaMetrics. They all have compatibility layers between them IIUC.

timeseries/rfcs/0006-native-histogram-promql.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Goals
+
+- Add a native histogram payload to the PromQL evaluator.
+- Support the smallest useful set of native histogram-aware functions: `histogram_count`, `histogram_sum`, and `histogram_avg`.
```
agavra, line 23: curious why these are helpful on their own, they're pretty easy to compute with just count/sum individually

agavra (commented):
A few comments on scope before I review the impl details.

apurvam (commented):
Hi @samsond 

Thanks for the RFC. Reading through this, I'm wondering if it makes sense to try to prototype an end-to-end solution here to make sure we are addressing the real problems with histograms in the best possible way. 

Specifically, the [Prometheus design doc](https://docs.google.com/document/d/1cLNv3aufPZb3fNfaJgdaRBZsInZKKIHo9E6HinJVbpM/edit?tab=t.0#heading=h.2ywoyp13c0bx) states the problems native histograms are solving as follows: 

> The problems result in the following practices:
>
>* To prevent Histograms from taking most of the cardinality budget, users tend to
> have very few buckets per Histogram (3–10), leading to huge errors in quantile
> estimation (unless the quantile value falls into a carefully selected value range
> with a few narrow buckets).
> * For the same reason, users shy away from partitioning Histograms. It is common 
> to have a counter for HTTP requests that is partitioned along many label 
> dimensions (status code, method, path, …) and then one latency Histogram for all
> HTTP requests combined, or only partitioned by a subset of those dimensions. 
> This is not only tedious during instrumentation, it also restricts the questions that
> a user can later ask with PromQL (e.g. what’s the latency distribution of 2xx vs. 
> 5xx).
> * Users have to know in advance, during instrumenting their code, what the 
> “interesting” bucket boundaries will be (SLO thresholds, Apdex target latencies, 
> etc.). That alone is problematic, but it also requires a change of the bucket layout 
> if the boundaries of interest ever change. Furthermore, a bucket layout change is 
> not only tedious to conduct, it also makes meaningful queries spanning Histogram 
> metrics from before and after the transition impossible (with the exception of a 
> compatible bucket layout change, which can only be done under specific 
> circumstances with a lot of care).
>
> A solution addressing the problems needs to meet the following requirements:
>
> * No (or minimal) configuration during instrumentation, in particular no 
> configuration that would require knowledge about the expected value ranges of 
> observations and what bucket boundaries will be of particular interest later. 
> * Quantile and percentage estimations with a small error margin across the entire 
> range of observations, including the long tail of latency distributions.
> * Enable partitioning of Histograms similar to what is currently done with plain 
> counters. Histograms will always be more expensive than plain counters, but the 
> gap needs to be much smaller than it is now.

I think that's a very specific set of goals. Evaluating our existing system to find it's limitations on that spectrum would be a good place to start. Then prototyping native histograms will allow us to evaluate a design far more robustly. 

My big worry about a series of large RFCs which only have a payoff in the distant future is that a) we may never get there, and b) if we do we will just call it a win just because we reached the end of the journey. 

That dynamic plays out very often in open source projects. It was probably necessary to do it that way when code was so expensive to write. But testing the limits of current histograms and then prototyping solutions that address those limits is much cheaper now. doing so will allow us to compare ourselves against victoria metrics / prometheus / other implementations, and then choose a path more wisely. 

I'd love your thoughts on such an approach.

timeseries/rfcs/0006-native-histogram-promql.md
samsond: Good point. I’ll do a quick comparison of the OTEL, Prometheus, and VictoriaMetrics histogram models and map the main compatibility/tradeoff points. I think it would help to separate interoperability at the edges from the internal representation we want to commit to.

timeseries/rfcs/0006-native-histogram-promql.md
samsond: The point of starting with those three is not that they are hard to compute in isolation, but that they are the smallest useful set that proves the evaluator can carry histogram-valued samples through function dispatch and return meaningful float projections with Prometheus-compatible semantics.

samsond:
> Hi @samsond
> 
> Thanks for the RFC. Reading through this, I'm wondering if it makes sense to try to prototype an end-to-end solution here to make sure we are addressing the real problems with histograms in the best possible way.
> 
> Specifically, the [Prometheus design doc](https://docs.google.com/document/d/1cLNv3aufPZb3fNfaJgdaRBZsInZKKIHo9E6HinJVbpM/edit?tab=t.0#heading=h.2ywoyp13c0bx) states the problems native histograms are solving as follows:
> 
> > The problems result in the following practices:
> > 
> > * To prevent Histograms from taking most of the cardinality budget, users tend to
> >   have very few buckets per Histogram (3–10), leading to huge errors in quantile
> >   estimation (unless the quantile value falls into a carefully selected value range
> >   with a few narrow buckets).
> > * For the same reason, users shy away from partitioning Histograms. It is common
> >   to have a counter for HTTP requests that is partitioned along many label
> >   dimensions (status code, method, path, …) and then one latency Histogram for all
> >   HTTP requests combined, or only partitioned by a subset of those dimensions.
> >   This is not only tedious during instrumentation, it also restricts the questions that
> >   a user can later ask with PromQL (e.g. what’s the latency distribution of 2xx vs.
> >   5xx).
> > * Users have to know in advance, during instrumenting their code, what the
> >   “interesting” bucket boundaries will be (SLO thresholds, Apdex target latencies,
> >   etc.). That alone is problematic, but it also requires a change of the bucket layout
> >   if the boundaries of interest ever change. Furthermore, a bucket layout change is
> >   not only tedious to conduct, it also makes meaningful queries spanning Histogram
> >   metrics from before and after the transition impossible (with the exception of a
> >   compatible bucket layout change, which can only be done under specific
> >   circumstances with a lot of care).
> > 
> > A solution addressing the problems needs to meet the following requirements:
> > 
> > * No (or minimal) configuration during instrumentation, in particular no
> >   configuration that would require knowledge about the expected value ranges of
> >   observations and what bucket boundaries will be of particular interest later.
> > * Quantile and percentage estimations with a small error margin across the entire
> >   range of observations, including the long tail of latency distributions.
> > * Enable partitioning of Histograms similar to what is currently done with plain
> >   counters. Histograms will always be more expensive than plain counters, but the
> >   gap needs to be much smaller than it is now.
> 
> I think that's a very specific set of goals. Evaluating our existing system to find it's limitations on that spectrum would be a good place to start. Then prototyping native histograms will allow us to evaluate a design far more robustly.
> 
> My big worry about a series of large RFCs which only have a payoff in the distant future is that a) we may never get there, and b) if we do we will just call it a win just because we reached the end of the journey.
> 
> That dynamic plays out very often in open source projects. It was probably necessary to do it that way when code was so expensive to write. But testing the limits of current histograms and then prototyping solutions that address those limits is much cheaper now. doing so will allow us to compare ourselves against victoria metrics / prometheus / other implementations, and then choose a path more wisely.
> 
> I'd love your thoughts on such an approach.

Thanks, this is a helpful framing.

I agree with the underlying concern: if we only discuss native histograms abstractly, it becomes too easy to mistake “we completed the planned phases” for “we solved the right problem.” I also agree that Prometheus’s motivation is quite specific: native histograms are not just “another sample type,” they are meant to address concrete limitations around bucket resolution, partitioning cost, and the need to lock in bucket boundaries too early.

I think your suggested approach is a good one, or at least the right correction to what I wrote so far. Before treating native histograms as the default destination, we should be more explicit about:

- what limitations we see in OpenData’s current float/classic histogram model,
- which of Prometheus’s target problems actually matter for our users and expected workloads,
- how OTEL histogram inputs map onto those needs in practice, and
- what an end-to-end prototype tells us that an evaluator-only slice cannot.

My intent with the phased RFC was to avoid prematurely committing us to full storage/ingest/API work, but I think your point is that there is a risk in the other direction too: we can end up validating architecture in pieces without validating that the overall direction is worth it.

So I think there are really two distinct questions here:

- do we want a narrow architectural slice to understand what supporting histogram-valued samples would entail in the evaluator/runtime, and
- do we want an end-to-end prototype or design spike that tests whether native histograms actually solve an important problem for OpenData better than the alternatives?

I do not think those are in conflict. In fact, the RFC probably needs to make clearer that the architecture slice is not sufficient evidence on its own. We likely need some combination of:

- a gap analysis of current/classic histogram limitations in OpenData,
- a comparison of Prometheus native histograms, OTEL histograms, and other implementations such as VictoriaMetrics,
- and a small prototype that lets us evaluate ingest, storage shape, query ergonomics, and cost profile together.

That would also make the success criteria much healthier.

apurvam:
@samsond You're right in that I don't think doing a prototype / gap analysis / comparison etc is in conflict with what's proposed here. I do think we need to do the former before committing to an incremental plan that builds native histograms though. 

So I would say that we should probably take a step back, do some meaningful benchmarks on the current state. Compare that with what prometheus can do with native histograms, and then agree on whether we want to close the gap. If we do, then we should prototype a solution that gets us to where we want to be. Once we have that prototype, we can create incremental RFCs / PRs like this one.

PR #334, opened 2026-03-24
Complete implementation of `merge_batch` and remove `merge`

### LINKED ISSUE #272: switch the implementations of `MergeOperator` to implement `merge_batch` and remove `merge`

micllam:
## Summary

- Implements `MergeOperator::merge_batch()` in a couple of test structs that only implemented `merge()`
- Removes trait method `MergeOperator::merge()` entirely

## Related Issues

Fixes #272.

## Test Plan

Existing tests updated to call `merge_batch()` instead of `merge()`.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/common/src/storage/mod.rs
+++ b/common/src/storage/mod.rs
@@ -170,33 +170,16 @@ pub struct WriteResult {
 
 /// Trait for merging existing values with new values.
 ///
-/// Merge operators must be associative: `merge(merge(a, b), c) == merge(a, merge(b, c))`.
+/// Merge operators must be associative: `merge_batch(merge_batch(a, [b]), [c]) == merge_batch(a, merge_batch(b, [c]))`.
 /// This ensures consistent merging behavior regardless of the order of operations.
 pub trait MergeOperator: Send + Sync {
-    /// Merges an existing value with a new value to produce a merged result.
-    ///
-    /// # Arguments
-    /// * `key` - The key associated with the values being merged
-    /// * `existing_value` - The current value stored in the database (if any)
-    /// * `new_value` - The new value to merge with the existing value
-    ///
-    /// # Returns
-    /// The merged value.
-    fn merge(&self, key: &Bytes, existing_value: Option<Bytes>, new_value: Bytes) -> Bytes;
-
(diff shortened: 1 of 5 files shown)


micllam:
Please note: I wasn't sure whether `SlateDbMergeOperatorAdapter::merge()` should delegate to `SlateDbMergeOperatorAdapter::merge_batch()` or whether they should each call `self.operator.merge_batch()` as currently implemented.

common/src/storage/mod.rs
```diff
@@ -170,33 +170,16 @@ pub struct WriteResult {
 
 /// Trait for merging existing values with new values.
 ///
-/// Merge operators must be associative: `merge(merge(a, b), c) == merge(a, merge(b, c))`.
+/// Merge operators must be associative: `merge_batch(merge_batch(a, b), c) == merge_batch(a, merge_batch(b, c))`.
```
agavra: nit: probably worth specifying `merge_batch(merge_batch(a, [b]), [c]))` instead to represent that they're batches now

micllam: Thanks for the suggestion. I've changed this now, but if it needs further tweaking, feel free to let me know.

micllam:
Thanks for your review, @agavra. I've now updated the doc comment.

PR #335, opened 2026-03-25
[vector] Add native support for cosine distance

### LINKED ISSUE #301: add native support for cosine distance to opendata-vector
## Summary

opendata-vector only supports L2 distance and dot-product. Many workloads rely on cosine distance to score vector similarity. Currently to use cosine distance users are expected to normalize their vectors to the unit hyper-sphere and then use L2 distance. This works because when vectors are normalized to the unit hyper-sphere both cosine distance and L2 distance yield the same rankings.

## Proposed Solution

We can add support for Cosine distance by simply doing the normalization as vectors are ingested. If the user configures Cosine distance, then we can normalize all vectors to the unit hypersphere at ingest time, and then use L2 distance for clustering and search. See the vector recall benchmark for an example of how to normalize.

micllam:
## Summary

Adds native support for cosine distance by normalizing vectors at ingest time and then using L2 distance, while the original values are still returned on `get()`.

## Related Issues

Fixes #301.

## Test Plan

Added various unit tests.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/vector/bench/src/recall.rs
+++ b/vector/bench/src/recall.rs
@@ -120,14 +120,6 @@ fn read_bvecs(path: &Path, limit: Option<usize>) -> Vec<Vec<f32>> {
     vectors
 }
 
-/// L2-normalize a vector in place.
-fn normalize_vec(v: &mut [f32]) {
-    let norm: f32 = v.iter().map(|x| x * x).sum::<f32>().sqrt();
-    if norm > 0.0 {
-        v.iter_mut().for_each(|x| *x /= norm);
-    }
-}
-
 // -- Recall / percentile helpers ----------------------------------------------
 
 fn recall_at_k(results: &[SearchResult], ground_truth: &[i32], k: usize) -> f64 {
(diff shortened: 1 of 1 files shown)


vector/bench/src/recall.rs
```diff
@@ -237,13 +237,15 @@ impl Dataset {
 fn distance_metric_to_str(m: DistanceMetric) -> &'static str {
     match m {
         DistanceMetric::L2 => "l2",
+        DistanceMetric::Cosine => "cosine",
```
rodesai, line 217: can you update the benchmarks here to no longer normalize from the benchmark but to just use vector's support for cosine distance directly?

micllam: I've now added a commit that removes the `normalize` field from `Dataset` and the corresponding steps to normalize vectors, using the cosine distance directly for `COHERE1M`.

micllam:
Thanks for your review. I've added a commit to address your comment and have rebased on main.

agavra (approved):
Looked at the bencher since @rodesai already reviewed the core implementation. It LGTM! Thanks @micllam for your additional contribution :) also if we miss a review for a few days definitely feel free to ping us on the discord for another round, things sometimes slip through the cracks.

micllam:
Thanks @agavra, and I'll keep that in mind.

PR #337, opened 2026-03-25
add read-only server mode via TsdbEngine enum

agavra:
Support running the HTTP server with either a full read-write Tsdb or a read-only TimeSeriesDbReader. The new TsdbEngine enum dispatches read methods to whichever inner engine is present while rejecting writes and returning empty metadata in read-only mode.

Key changes:
- TsdbEngine enum in tsdb.rs wrapping `Arc<Tsdb>` or `Arc<TimeSeriesDbReader>`
- AppState, build_router, TimeSeriesHttpServer updated to use `Arc<TsdbEngine>`
- Scraper startup guarded (skipped in read-only mode)
- `read_only` and `reader` config section added to PrometheusConfig
- POST /-/flush endpoint for manually triggering storage flush
- storage field made optional on TimeSeriesHttpServer (not needed for readers)

---

I tested this by manually running a write server and read-only server and confirmed that they work in tandem (once flushing the write server).

### DIFF
--- a/timeseries/src/config.rs
+++ b/timeseries/src/config.rs
@@ -8,7 +8,7 @@ use std::time::Duration;
 
 use common::StorageConfig;
 
-/// Configuration for opening a [`TimeSeriesDb`](crate::TimeSeriesDb) database.
+/// Configuration for opening a [`TimeSeriesDb`](crate::timeseries::TimeSeriesDb) database.
 ///
 /// This struct holds all the settings needed to initialize a time series
 /// instance, including storage backend configuration and operational parameters.
--- a/timeseries/src/reader.rs
+++ b/timeseries/src/reader.rs
@@ -115,7 +115,7 @@ impl QueryReader for ReaderQueryReader {
 
 /// A read-only view of a time series database.
 ///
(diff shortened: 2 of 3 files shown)


timeseries/src/promql/config.rs
```diff
@@ -56,6 +62,31 @@ impl Default for PrometheusConfig {
...
+
+/// Reader-specific configuration for read-only server mode.
+#[derive(Debug, Clone, Deserialize)]
+pub struct ServerReaderConfig {
+    /// Manifest poll interval in seconds.
+    /// The reader polls the SlateDB manifest at this interval to discover
+    /// new data written by the writer process.
+    /// Defaults to 300 (5 minutes).
+    #[serde(default = "default_reader_refresh_interval_secs")]
+    pub refresh_interval_secs: u64,
```
apurvam: You may want to consider skipping wal replay setting the checkpoint lifetime too. The settings below allow you to run the reader for a day. Because of filed bugs in slatedb, the compactions from the writer corrupt reader slate everytime it polls a new manifest after a compaction, causing sequence number errors on reads.

```
refresh_interval: 86400000
checkpoint_lifetime: 172800000
skip_wal_replay: true
```

PR #354, opened 2026-03-30
[timeseries] Refactor query engine into explicit execution phases

### LINKED ISSUE #345: [timeseries] Refactor query engine into explicit execution phases
## Summary

The current PromQL evaluator has multiple query paths that all perform broadly the same storage-facing work:

  - compute concrete query/window inputs
  - resolve matching series from indexes
  - load forward-index metadata
  - load sample data
  - shape that data into evaluator inputs
  - evaluate the final expression

Today, that work is duplicated and interleaved differently across multiple evaluator paths. This makes the code harder to reason about, harder to instrument, and harder to optimize consistently.

This issue proposes a structural refactor: move all query paths that execute through the PromQL evaluator onto one explicit phased execution model.

## Goal

Refactor evaluator-backed query execution around explicit phases:

1. Plan
2. ResolveMetadata
3. LoadSamples
4. ShapeSamples
5. Evaluate

The purpose is to create one shared execution model that later optimizations can build on cleanly, specifically:

  - cold-path I/O parallelism
  - warm-path reuse/caching
  - storage-layout improvements
  - consistent use of SlateDB scan vs get.
  - instrumentation (how much time is a query spending where?)

## Scope

This issue covers all query paths that execute through the PromQL evalutaor, including:

  - instant vector selector evaluation 
  - matrix selector evaluation in 
  - range-query execution 
  - subquery vector-selector fast path
  - as much of generic subquery evaluation as can cleanly align to the same model

This refactor does not cover apis that don't go through PromQL, specifically:

  - metadata/discovery endpoints:
  - metadata catalog lookup

## Why this refactor is needed

Today, the same storage-facing traversal is repeated across multiple evaluator paths.

### Matrix selector path

Almost identical to then non-matrix selector range query path:

  - compute [start_ms, end_ms]
  - list overlapping buckets
  - resolve candidate series ids
  - load forward index
  - iterate candidate series ids
  - load samples for series through `CachedQueryReader`
  - append results into a per-series map

### Subquery vector-selector fast path

This does almost the same thing

  - list buckets
  - resolve candidates
  - load forward index
  - iterate candidate series ids
  - call timeseries/src/promql/evaluator.rs:1180
  - merge per-series samples
  - sort/dedup them

  ### Instant vector selector path

  timeseries/src/promql/evaluator.rs:1295 again does:

  - list buckets
  - resolve candidates
  - load forward index
  - iterate candidate series ids
  - call timeseries/src/promql/evaluator.rs:1357
  - choose the latest point in lookback

  So today the expensive storage-facing work is repeated across multiple evaluator paths, with path-specific shaping layered on
  afterward.

  ## Why the current structure makes optimization hard

  The evaluator currently interleaves:

  - time/window planning
  - selector resolution
  - forward-index metadata loading
  - cache checks
  - sample I/O
  - filtering
  - merge/dedup
  - evaluator preparation

  A concrete example is the [samples method](https://github.com/opendata-oss/opendata/blob/main/timeseries/src/promql/evaluator.rs#L393) which currently hides:

  - cache lookup
  - full-bucket load on miss
  - cache population
  - range filtering

inside what looks like a single sample read.

That makes several things harder than they need to be:

  - I/O parallelism: there is no explicit sample-work artifact to schedule
  - instrumentation: timers end up covering mixed work instead of real phases
  - warm-path reuse: reusable artifacts are not first-class outputs
  - consistency: some paths can benefit from optimizations while others keep older traversal logic

## Proposed execution model

All evaluator-backed query paths should align to the same phase model:

1. Plan
  1.1 Compute concrete query/window/bucket inputs
2. ResolveMetadata
  2.1 resolve candidate series ids and load the metadata needed to read samples
3. LoadSamples
  3.1 load samples for explicit (bucket, series) work items
4. ShapeSamples
  4.1 merge/filter/dedup/sample-shape into evaluator-ready per-series structures
5. Evaluate
  5.1 run PromQL logic on prepared in-memory artifacts

The final consumer differs by path:

  - instant vector selector keeps latest-point semantics
  - matrix selector keeps raw window semantics
  - subquery vector-selector fast path buckets samples into step-aligned outputs

But the expensive metadata/sample pipeline should be shared. 

These phases are logical stages, but metadata resolution, sample loading, and sample shaping should be designed so they can pipeline per  bucket rather.

## Why this is a better foundation

This gives one place to attach later optimizations:

  - cold I/O parallelism belongs in ResolveMetadata and LoadSamples
  - warm-path reuse belongs on ResolveMetadata and ShapeSamples outputs
  - metric-prefixed layout and singleton scans affect LoadSamples
  - phase timings become meaningful and comparable across query types

It also resolves an existing structural problem: matrix/vector/subquery paths currently repeat nearly the same storage traversal with slightly different shaping logic.

## Acceptance criteria

This issue is complete when:

  - evaluator-backed query paths share one explicit phase model
  - duplicated metadata/sample traversal logic is consolidated
  - phase outputs are represented by explicit intermediate artifacts
  - current query semantics are preserved
  - aggregate phase-level timings exist
  - later optimization issues can target phases rather than individual evaluator call paths

## Follow-on work enabled by this issue

This refactor is the foundation for later work on:

  - bounded metadata/sample read scheduling
  - bucket-level pipelining with execution hapenning in parallel across buckets.
  - warm-path reuse
  - sample storage layout improvements
  - merge-aware sample read shape
  - improved query-complete instrumentation

apurvam:
## Summary

Refactor evaluator query paths onto shared pipeline

Move all evaluator-backed query paths (instant vector selector, matrix
selector, subquery vector-selector fast path) onto one explicit phased
execution model: Plan → ResolveMetadata → LoadSamples → ShapeSamples.

The new pipeline module (pipeline.rs) provides:
    - QueryPlan with path-specific constructors for planning
    - resolve_bucket_metadata() consolidating duplicated selector+forward-index logic
    - build_bucket_sample_work() making sample I/O explicit
    - load_bucket_samples() centralizing sample reads
    - shape_instant/matrix/subquery_results() isolating path-specific shaping
    - execute_selector_pipeline() as unified orchestrator branching on QueryPathKind
    - Aggregate phase timings logged via tracing::debug

The three evaluator methods become thin wrappers: apply time modifiers,
construct QueryPlan, delegate to execute_selector_pipeline. ~150 lines
net reduction in evaluator.rs with duplicated traversal logic removed.

This sets the basis for future optimizations in the query layer. 

## Related Issues

Fixes #345

## Test Plan

Mainly a refactor. Added unit tests for the pipeline component. All existing tests should continue to pass.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2522,6 +2522,7 @@ dependencies = [
  "tokio-util",
  "tracing",
  "ulid",
+ "zstd",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,7 @@ proptest = "1"
 dashu-float = "0.4"
 snap = "1"
 criterion = "0.5"
+zstd = "0.13"
 tempfile = "3"
 opendata-macros = { path = "./macros" }
 
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -10,7 +10,9 @@ pub use bytes::BytesRange;
 pub use clock::Clock;
 pub use sequence::{DEFAULT_BLOCK_SIZE, SequenceAllocator, SequenceError, SequenceResult};
 pub use serde::seq_block::SeqBlock;
-pub use storage::config::{BlockCacheConfig, FoyerHybridCacheConfig, StorageConfig};
+pub use storage::config::{
+    BlockCacheConfig, FoyerHybridCacheConfig, ObjectStoreConfig, StorageConfig,
+};
 pub use storage::factory::{
     CompactorBuilder, DbBuilder, StorageBuilder, StorageReaderRuntime, StorageSemantics,
     create_object_store, create_storage_read,
--- a/ingest/Cargo.toml
+++ b/ingest/Cargo.toml
@@ -18,6 +18,7 @@ tokio.workspace = true
 tokio-util.workspace = true
 tracing.workspace = true
 ulid.workspace = true
+zstd.workspace = true
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["test-util"] }
--- a/ingest/rfcs/0001-stateless-ingest.md
+++ b/ingest/rfcs/0001-stateless-ingest.md
@@ -220,6 +220,11 @@ pub struct IngestorConfig {
   ///
   /// Defaults to 1000.
   pub max_buffered_inputs: usize,
+
+  /// Compression algorithm applied to the record block in data batches.
+  ///
+  /// Defaults to `None` (uncompressed).
+  pub batch_compression: CompressionType,
 }
 ```
 The queue manifest takes the name specified in `manifest_path`.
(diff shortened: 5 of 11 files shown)


timeseries/src/promql/pipeline.rs
```diff
@@ -0,0 +1,1000 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Phase 3: Sample work generation
+// ---------------------------------------------------------------------------
+
+/// Build explicit sample work for a bucket from resolved metadata.
+///
+/// For instant vector selectors, `skip_fingerprints` enables cross-bucket dedup
+/// (series already found in a newer bucket are excluded from the work list).
```
agavra: is this necessary b/c of lookback? in other words we respect the first found fingerprint and we can skip them in older buckets if we're trying to respect the lookback for that point?

timeseries/src/promql/pipeline.rs
```diff
@@ -0,0 +1,1000 @@
...
+        start_ms: plan.sample_start_ms,
+        end_ms: plan.sample_end_ms,
+    })
+}
+
+/// Build sample work with lenient error handling (skip missing specs).
+///
+/// Matches the existing `fetch_series_samples` behavior where a missing forward
+/// index spec is silently skipped rather than treated as an error.
+pub(crate) fn build_bucket_sample_work_lenient(
```
agavra: is this just the previous method with a "don't throw" flag? can we just add a parameter to the previous one?

timeseries/src/promql/pipeline.rs
apurvam: Yes, exactly. And it only is relevant when the 'eval time - lookback window' straddles bucket boundaries. I was contemplating just deduping at the very end for instant vectors vs keep track of this during loading. But it would mean we could load many unnecessary samples all the time in degenerate cases.

timeseries/src/promql/pipeline.rs
apurvam: Yes, basically. I can merge the two and add a param.

PR #356, opened 2026-03-30
[ingest] Expose metadata on CollectedBatch

agavra:
Add BatchMetadata to the public API so consumers can access the per-range metadata (start_index, ingestion_time_ms, payload) that ingestors attach to each batch.


agavra:
Good call, updated the RFC. I'm certain this was @cadonna's intention since the metadata is user supplied and opaque to the ingestor 😆

PR #363, opened 2026-03-31
[timeseries] Add bounded-parallel bucket pipeline for query execution

### LINKED ISSUE #346: [timeseries] Add parallel reads from storage with bounded concurrency
## Summary

The query pipeline refactor in #345 moved evaluator-backed query paths onto a shared logical structure:

  - Plan
  - ResolveMetadata
  - LoadSamples
  - ShapeSamples
  - Evaluate

This issue adds the next layer on top of that structure:

  - parallel reads from storage with bounded concurrency
  - bucket-level pipelining through the shared phases
  - unified scheduling across evaluator-backed query paths

The goal is to reduce cold-query latency by overlapping storage work safely, while keeping the implementation simple and the pipelinestructure uniform.

## Goal

Use the shared query pipeline to:

  - resolve metadata for multiple buckets in parallel
  - load samples for multiple buckets in parallel
  - allow each bucket to progress independently through the logical phases as its inputs become ready
  - bound concurrency so we do not overwhelm storage / object store / S3

## Scope

This issue applies to the same evaluator-backed query paths covered by #345:

  - instant vector selector evaluation
  - matrix selector evaluation
  - range-query execution over evaluator-backed expressions
  - subquery vector-selector fast path

## Design direction

The pipeline phases remain logical stages, not global barriers.

After Plan, buckets should be able to progress independently:

  - ResolveMetadata(bucket)
  - LoadSamples(bucket)
  - ShapeSamples(bucket)

That means:

  - a bucket that finishes metadata resolution can begin sample loading immediately
  - a bucket that finishes sample loading can be shaped immediately
  - other buckets may still be in earlier phases

This is intentionally a simple unified pipelining model.

For now, we will not preserve path-specific scheduling tricks such as pruning older-bucket sample work for instant selectors before loading. Instead:

  - sample work is scheduled uniformly by bucket
  - instant-selector newest-wins behavior is preserved in shaping/final result semantics

This keeps the concurrency model simple and uniform for the first scheduling PR.

## Expected outcome

This should reduce cold-query latency by:

  - overlapping metadata reads across buckets
  - overlapping sample reads across buckets
  - overlapping shaping work for one bucket with I/O for another bucket

while still keeping the code aligned with the new query architecture.

## Acceptance criteria

  This issue is complete when:

  - metadata resolution is bounded-parallel across buckets
  - sample loading is bounded-parallel across buckets
  - buckets can progress independently through the pipeline
  - evaluator-backed query semantics are preserved
  - aggregate phase timing remains available and becomes meaningful under parallel execution
  - the code remains on one unified scheduling model across the pipeline-backed query paths

apurvam:
## Summary

Replace the serial bucket loop in the query pipeline with per-bucket pipelined tasks bounded by separate metadata and sample concurrency semaphores.

Each bucket independently progresses through three phases:

  1. Acquire metadata semaphore permit → resolve metadata → release
  2. Build sample work (no permit needed, fast)
  3. Acquire sample semaphore permit → load samples → release

All bucket tasks are launched concurrently via buffer_unordered. The two tokio::sync::Semaphore instances (one for metadata, one for samples) bound actual I/O independently, so bucket A can be loading samples while bucket B is still resolving metadata.

Results are reassembled in plan order before shaping to preserve query semantics.

Cross-bucket fingerprint pruning for instant selectors is intentionally dropped — all buckets load all matching series uniformly. Newest-wins semantics are preserved in `shape_instant_results` via plan-order dedup.

The pipelined executor creates one `CachedQueryReader` per bucket task, but all tasks share the same `Arc<QueryReaderEvalCache>/` owned by the Evaluator. This preserves cache reuse across:

  - Concurrent bucket tasks within a single pipeline invocation
  - Successive range-query steps (one Evaluator loops over steps)
  - Separate selector leaves in the same expression (e.g. foo + foo)

To support concurrent access from multiple bucket tasks, `QueryReaderEvalCache` was converted from `HashMap` to `DashMap`. All `CachedQueryReader` methods now take `&self` instead of `&mut self`.

Sample data is cached as `Arc<Vec<Sample>>` so cache hits filter over a shared slice and only allocate the filtered result, avoiding a full-vector clone on the warm path.

Two new fields on `QueryOptions` (default: 4 each):
  - metadata_concurrency
  - sample_concurrency

Plumbed through `Evaluator::with_concurrency()` and `PipelineConcurrency`. Values are clamped to minimum 1 to prevent Semaphore::new(0) from blocking forever.

PipelineTimings tracks per-bucket metrics aggregated into:
  - pipeline_wall_ms (user-visible latency)
  - metadata_resolve_sum_ms / sample_load_sum_ms (total work)
  - metadata_queue_wait_sum_ms / sample_queue_wait_sum_ms (contention)
  - metadata_bucket_count / sample_bucket_count
  - shape_samples_ms

All emitted as numeric tracing::debug fields.

## Related Issues

Fixes #346

## Test Plan

Targeted async integration tests using CountingQueryReader (a QueryReader wrapper with AtomicUsize call counters and per-bucket delay injection):

  - Instant newest-wins when newer bucket finishes last (delayed)
  - Metadata contention timing (concurrency=1, serialized delays)
  - Sample contention timing (concurrency=1, serialized delays)
  - Cache reuse without reloading (call counts stable on 2nd call)
  - End-to-end evaluate_range: 5 steps, all reader methods called exactly once (cache reuse across range-query step loop)

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/evaluator.rs
+++ b/timeseries/src/promql/evaluator.rs
@@ -78,12 +78,10 @@ impl QueryReaderEvalCache {
         &self,
         bucket: &TimeBucket,
     ) -> dashmap::mapref::one::Ref<'_, TimeBucket, QueryReaderBucketEvalCache> {
-        if !self.cache.contains_key(bucket) {
-            self.cache
-                .entry(*bucket)
-                .or_insert_with(QueryReaderBucketEvalCache::new);
-        }
-        self.cache.get(bucket).expect("bucket just inserted")
+        self.cache
+            .entry(*bucket)
+            .or_insert_with(QueryReaderBucketEvalCache::new)
+            .downgrade()
     }
 
     pub(crate) fn cache_forward_index(
@@ -421,7 +419,9 @@ impl<'reader, R: QueryReader> CachedQueryReader<'reader, R> {
             return Ok(filtered);
         }
 
-        // Not in cache, load from underlying reader with wide bounds to cache the whole bucket
+        // Cache the full bucket on a miss so later overlapping range-query
+        // steps and sibling selector evaluations can slice locally without
+        // paying for another storage read.
         let samples = self
             .reader
             .samples(bucket, series_id, i64::MIN, i64::MAX)
--- a/timeseries/src/promql/pipeline.rs
+++ b/timeseries/src/promql/pipeline.rs
@@ -15,12 +15,17 @@ use crate::index::ForwardIndexLookup;
 use crate::model::{Label, Sample, SeriesFingerprint, SeriesId, TimeBucket};
 use crate::promql::evaluator::{
     CachedQueryReader, EvalResult, EvalSample, EvalSamples, EvaluationError, ExprResult,
+    QueryReaderEvalCache,
 };
 use crate::promql::selector::evaluate_selector_with_reader;
 use crate::query::QueryReader;
+use futures::stream;
+use futures::{StreamExt, TryStreamExt};
 use promql_parser::parser::VectorSelector;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
+use std::time::Instant;
+use tokio::sync::mpsc;
 
 // ---------------------------------------------------------------------------
 // Phase artifact types
@@ -66,9 +71,9 @@ pub(crate) struct QueryPlan {
 impl QueryPlan {
     /// Build sample work for a bucket, dispatching strict vs lenient by path kind.
     ///
-    /// Under the pipelined executor, all buckets load all matching series
-    /// uniformly (no cross-bucket pruning). Newest-wins semantics for instant
-    /// selectors are enforced in shaping, not here.
+    /// All matching series in the bucket become explicit sample work items.
+    /// Instant selectors rely on `shape_instant_results` to enforce newest-wins
+    /// across buckets, so planning does not prune cross-bucket duplicates here.
     fn build_work(&self, metadata: &BucketMetadata) -> EvalResult<BucketSampleWork> {
         let strict = !matches!(self.path_kind, QueryPathKind::SubqueryVectorSelector { .. });
         build_bucket_sample_work(self, metadata, strict)
@@ -524,24 +529,39 @@ struct BucketTaskTiming {
     sample_load_ms: f64,
 }
 
+/// One bucket waiting to be processed by the metadata worker pool.
+struct MetadataStageJob {
+    idx: usize,
+    bucket: TimeBucket,
+    queued_at: Instant,
+}
+
+/// One bucket whose metadata has been resolved and is ready for sample loading.
+struct SampleStageItem {
+    idx: usize,
+    work: Option<BucketSampleWork>,
+    timing: BucketTaskTiming,
+    queued_at: Instant,
+}
+
 /// Aggregate phase timings for the selector pipeline.
 ///
 /// Under parallel execution, wall time shows user-visible latency while sum
-/// time shows total work. Queue wait measures time spent waiting for a
-/// concurrency permit.
+/// time shows total work. Queue wait measures time spent waiting for a stage
+/// worker slot.
 #[derive(Debug, Default, Clone)]
 pub(crate) struct PipelineTimings {
     /// Wall-clock time for the entire metadata+sample pipeline phase.
     pub pipeline_wall_ms: f64,
     /// Sum of per-bucket metadata resolution durations.
     pub metadata_resolve_sum_ms: f64,
-    /// Sum of per-bucket metadata semaphore wait durations.
+    /// Sum of per-bucket metadata stage queue wait durations.
     pub metadata_queue_wait_sum_ms: f64,
     /// Number of buckets that had metadata resolved.
     pub metadata_bucket_count: usize,
     /// Sum of per-bucket sample load durations.
     pub sample_load_sum_ms: f64,
-    /// Sum of per-bucket sample semaphore wait durations.
+    /// Sum of per-bucket sample stage queue wait durations.
     pub sample_queue_wait_sum_ms: f64,
     /// Number of buckets that had samples loaded.
     pub sample_bucket_count: usize,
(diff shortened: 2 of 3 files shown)


apurvam:
Reviewer notes: 

1. Please check the changes to allow concurrent modification of `QueryReaderEvalCache` through `CachedQueryReader`. the `QueryEvalCacheReader` is shared across all concurrent tasks with this change.
2. The main logic change is in pipeline.rs:590 `execute_selector_pipeline`.
3. Out of the ~1000 additions, ~700 are tests, including new test infra like `CountingQueryReader` to ensure that we are reusing the cache across an entire query and also infra to introduce delays to test both the instrumentation and the permitting system.

timeseries/src/promql/evaluator.rs
```diff
@@ -45,53 +45,55 @@ impl From<crate::error::Error> for EvaluationError {
...
-        self.cache
-            .entry(*bucket)
-            .or_insert_with(QueryReaderBucketEvalCache::new)
+    ) -> dashmap::mapref::one::Ref<'_, TimeBucket, QueryReaderBucketEvalCache> {
+        if !self.cache.contains_key(bucket) {
+            self.cache
+                .entry(*bucket)
+                .or_insert_with(QueryReaderBucketEvalCache::new);
+        }
+        self.cache.get(bucket).expect("bucket just inserted")
```
agavra:
```suggestion
        self.cache
            .entry(*bucket)
            .or_insert_with(QueryReaderBucketEvalCache::new)
            .downgrade()
```
goes from 3 -> 1 lookup on the map (downgrade converts it to a read lock)

timeseries/src/promql/evaluator.rs
```diff
@@ -408,12 +427,14 @@ impl<'reader, R: QueryReader> CachedQueryReader<'reader, R> {
             .samples(bucket, series_id, i64::MIN, i64::MAX)
             .await?;
 
-        // Cache the full sample set
-        self.cache
-            .cache_samples(*bucket, series_id, samples.clone());
+        // Cache the full sample set, then filter the shared handle
+        self.cache.cache_samples(*bucket, series_id, samples);
```
agavra, line 431: why are we caching the full sample set before filtering? is it because `query_range` will need time samples that aren't covered by the start/end range or something like that? an inline comment would be useful

timeseries/src/promql/pipeline.rs
```diff
@@ -65,33 +65,13 @@ pub(crate) struct QueryPlan {
...
-        if matches!(self.path_kind, QueryPathKind::InstantVector { .. }) {
-            for series in &data.series_data {
-                if series.samples.last().is_some() {
-                    seen.insert(series.fingerprint);
-                }
-            }
-        }
+    ///
+    /// Under the pipelined executor, all buckets load all matching series
+    /// uniformly (no cross-bucket pruning). Newest-wins semantics for instant
```
agavra: this comment doesn't make sense on it's own without the context of what it was before

timeseries/src/promql/pipeline.rs
```diff
@@ -507,64 +477,223 @@ pub(crate) fn shape_subquery_results(
...
+/// preserve query semantics.
 pub(crate) async fn execute_selector_pipeline<R: QueryReader>(
-    reader: &mut CachedQueryReader<'_, R>,
+    reader: &R,
+    cache: &Arc<crate::promql::evaluator::QueryReaderEvalCache>,
     plan: &QueryPlan,
     selector: &VectorSelector,
+    concurrency: &PipelineConcurrency,
 ) -> EvalResult<ExprResult> {
+    use futures::TryStreamExt;
```
agavra: nit: probably worth telling your CLAUDE.md to put imports up top. i wish lint caught that automatically, but it's generally nicer for production code to have one place to look for imports

timeseries/src/promql/pipeline.rs
```diff
@@ -507,64 +477,223 @@ pub(crate) fn shape_subquery_results(
...
+        .map(|(idx, &bucket)| {
+            let meta_sem = metadata_sem.clone();
+            let samp_sem = sample_sem.clone();
+            let task_cache = cache.clone();
+            async move {
+                let mut timing = BucketTaskTiming::default();
+
+                // -- Metadata phase (bounded by metadata semaphore) --
+                let meta_queue_start = Instant::now();
+                let _meta_permit = meta_sem.acquire().await.map_err(|_| {
```
agavra: instead of manually droping it's nice to put this in a block (`{/}`) which makes it easier to see where the scope applies to

(you can ignore this comment if you take my other suggestion)

timeseries/src/promql/pipeline.rs
```diff
@@ -507,64 +477,223 @@ pub(crate) fn shape_subquery_results(
...
+                Ok((idx, Some(data), timing))
+            }
+        })
+        .collect();
 
-        plan.track_seen_fingerprints(&data, &mut seen_fingerprints);
-        all_bucket_data.push(data);
-    }
+    // Run all bucket tasks concurrently — semaphores bound actual I/O.
+    let results: Vec<(usize, Option<BucketSampleData>, BucketTaskTiming)> =
```
agavra: I don't love the pattern of spinning up a future per bucket and then having those futures block on semaphores as soon as they've been spun up since they hold onto their own state between the metadata and sample phases. For a 30-day window that's 720 tasks per query (x the number of queries in a dashboard), when the default is `8` that seems like overkill.

I think we'd be better off splitting this into two stream stages connected by a channel, where `buffer_unordered` is the concurrency limiter instead of semaphores:

```rust
let (meta_tx, meta_rx) = mpsc::channel(concurrency.samples);

// Stage 1: metadata, at most concurrency.metadata in flight
let metadata_stage = stream::iter(plan.buckets.iter().enumerate())
    .map(|(idx, &bucket)| async move {
        // resolve metadata + build work
    })
    .buffer_unordered(concurrency.metadata)
    .try_for_each(|item| { /* send to meta_tx */ });

drop(meta_tx); // so the channel closes when stage 1 finishes

// Stage 2: sample loading, at most concurrency.samples in flight
let sample_stage = ReceiverStream::new(meta_rx)
    .map(|(idx, work_opt)| async move {
        // load samples for this bucket
    })
    .buffer_unordered(concurrency.samples);

tokio::try_join!(metadata_stage, sample_stage.try_collect())?;
```

This way:
- No semaphores needed (which by the way using `buffer_unordered(N)` is exactly meant for this use case)
- The stream only creates the next future when a slot opens, so no idle tasks
- The channel lets stage 2 start consuming as stage 1 produces, so we still get pipeline overlap
- Channel capacity bounds how many resolved-but-not-yet-sampled buckets accumulate (so we can control memory pressure somewhat)

apurvam:
The current PR still has fairly coarse grained parallelism. I filed this to track followup work to make sample IO truly parallel: https://github.com/opendata-oss/opendata/issues/366

timeseries/src/promql/evaluator.rs
agavra, line 347: we should probably use `entry().or_insert_with_async(...)` to avoid racy checks that result in doing the work loading the index twice (same problem as the thundering herd we have in slate). Same for inverted idnex

timeseries/src/promql/pipeline.rs
```diff
@@ -968,4 +1335,505 @@ mod tests {
...
+        async fn inverted_index(
+            &self,
+            bucket: &TimeBucket,
+            terms: &[crate::model::Label],
+        ) -> crate::util::Result<Box<dyn crate::index::InvertedIndexLookup + Send + Sync + 'static>>
+        {
+            self.inverted_index_calls
+                .fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+            if let Some(&delay) = self.metadata_delays.get(bucket) {
+                tokio::time::sleep(delay).await;
```
agavra, line 1654: if we're adding a bunch of sleeps with delay we should make sure to use `#[tokio(test, start_paused = true)]` so taht the tests aren't flaky and using wall-clock time.

it looks like the tests written aren't depending on this to succeed, but we might get false positives in flaky environments. we should just make this deterministic.

timeseries/src/promql/pipeline.rs
apurvam: I like this approach. I'll need to think through how this will work with the followup #366 . THe current PR doesn't do sample IO in parallel within a bucket, which the followup is going to do. 

Still, creating a queue of work that fixed task pool can pick up is a better model than creating 100's of tasks and then blocking on a synchronization barrier.

apurvam: Also, the current implementation of spawning tasks and blocking on semaphores is an overhang from the benchmarking branch, which itself was arrived at incrementally. So it's probably more messy than it needs to be.

apurvam: This is a good suggestion. Updated. The code also has the plan to incoporate parallel sample IO on top of the current version.

timeseries/src/promql/evaluator.rs
apurvam: I don't think this is a real issue. Metadata loading happens serially per bucket. there shouldn't be a race with other loads of the same data, since buckets have disjoint metadata.

agavra: Ah I thought we get the forward index by fingerprint. That can be a future optimization. We can "fix" this in a follow up (would be good to do anyway)

timeseries/src/promql/pipeline.rs
```diff
@@ -507,64 +482,327 @@ pub(crate) fn shape_subquery_results(
...
+        .iter()
+        .enumerate()
+        .map(|(idx, &bucket)| MetadataStageJob {
+            idx,
+            bucket,
+            queued_at: pipeline_start,
+        })
+        .collect();
+
+    let (sample_tx, sample_rx) = mpsc::channel(sample_concurrency);
```
agavra, line 734: Claude seemed convinced we needed to manually `drop(sample_tx)` before we call try_join or else the `rx` stream will never end. It looks like you're using a different syntax (`stream::unfold`) than what Claude suggested though so it may not have the same problem (though I'm not sure how the stream knows it's done without all `tx` instances being dropped.

timeseries/src/promql/evaluator.rs
apurvam: Not sure I'm following the optimization you have in mind. The only further optimizations I see possible for metadata is to preserve the cache across queries. Otherwise we'd rely on the slatedb block cache to keep the loading fast and cheap across queries. Or am I missing something?

agavra: My line of thinking was that the forward index is going to be the same for every series with the same fingerprint, even across buckets. We don't need to load the forward index entry for a series if we've already loaded it for the same fingerprint in another bucket. Since the series churn is relatively slow, I expect we'd have lots of fingerprint cache hits.

agavra: Oh 🤦 sorry I goofed. You can't get fingerprint until you have the forward idx lol.

timeseries/src/promql/pipeline.rs
apurvam: Nice. I didn't know about start_paused.

timeseries/src/promql/pipeline.rs
apurvam: Here is codex's is explanation, which passes my sniff test: 

The queue is created at timeseries/src/promql/pipeline.rs:734. The metadata stage owns the original sample_tx by moving it into the async move block at  timeseries/src/promql/pipeline.rs:736. Each completed metadata job clones that sender just long enough to send(item) at timeseries/src/promql/pipeline.rs:747. 

When metadata production finishes, those temporary clones are gone, then the original sample_tx is dropped when metadata_stage returns. At that point sample_rx.recv().await in the stream::unfold(...) at timeseries/src/promql/pipeline.rs:763  starts returning None, so the sample stream ends naturally. No explicit drop(sample_tx) is needed because there is no live sender left  in outer scope.

The queueing mechanism is:

  1. Build one metadata job per bucket.
  2. Run metadata jobs with buffer_unordered(metadata_concurrency).
  3. As each metadata job finishes, push its SampleStageItem into the bounded mpsc channel.
  4. The sample side pulls from that channel, again with buffer_unordered(sample_concurrency).
  5. Because the channel capacity is sample_concurrency, if sample workers fall behind, send(...).await blocks and metadata gets backpressured instead of piling up unbounded work.
  6. Sample results come back out of order, then get sorted by original bucket index before shaping.

  So conceptually the queue is just the handoff buffer between stage 1 and stage 2, and channel closure is the signal that stage 1 is  done.

agavra: > The metadata stage owns the original sample_tx by moving it into the async move block at timeseries/src/promql/pipeline.rs:736. 

that was the part I missed. I didn't have `async move` in my original suggestion IIRC.

EDIT: looks like I did, either way that explains why you don't need a manual drop.

agavra (approved):

timeseries/src/promql/evaluator.rs
apurvam: Yes. assuming cross-bucket metadata will also complicate compaction, when we get to things like segment-based compaction.

PR #365, opened 2026-03-31
[timeseries] add write_timeout and related methods

### LINKED ISSUE #364: [timeseries] Add timeout-based write method
This is a subissue of #267 in relation to write method naming:

> TimeSeriesDb lacks a timeout variant while LogDb and VectorDb support timeout-based backpressure.

micllam:
## Summary

Adds `write_timeout` and related methods.

## Related Issues

Fixes #364.

## Test Plan

New unit tests added and existing updated with the extra `timeout: Option<Duration>` parameter.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/minitsdb.rs
+++ b/timeseries/src/minitsdb.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 use std::sync::Arc;
+use std::time::Duration;
 
 use async_trait::async_trait;
 use common::coordinator::{Durability, WriteCoordinator, WriteCoordinatorConfig, WriteError};
@@ -172,6 +173,9 @@ impl MiniTsdb {
     }
 
     /// Ingest a batch of series with samples in a single operation.
+    ///
+    /// If `timeout` is provided, waits up to the given duration for space in the
+    /// write queue. Otherwise, fails immediately when the queue is full.
     #[tracing::instrument(
         level = "debug",
         skip_all,
@@ -181,7 +185,11 @@ impl MiniTsdb {
             total_samples = series_list.iter().map(|s| s.samples.len()).sum::<usize>()
         )
     )]
-    pub(crate) async fn ingest_batch(&self, series_list: &[Series]) -> Result<()> {
+    pub(crate) async fn ingest_batch(
+        &self,
+        series_list: &[Series],
+        timeout: Option<Duration>,
+    ) -> Result<()> {
         let total_samples = series_list.iter().map(|s| s.samples.len()).sum::<usize>();
 
         tracing::debug!(
@@ -192,10 +200,16 @@ impl MiniTsdb {
         );
 
         let handle = self.write_coordinator.handle(WRITE_CHANNEL);
-        let mut write_handle = handle
-            .try_write(series_list.to_vec())
-            .await
-            .map_err(|e| map_write_error(e.discard_inner()))?;
+        let mut write_handle = match timeout {
+            Some(t) => handle
+                .write_timeout(series_list.to_vec(), t)
+                .await
+                .map_err(|e| map_write_error(e.discard_inner()))?,
+            None => handle
+                .try_write(series_list.to_vec())
+                .await
+                .map_err(|e| map_write_error(e.discard_inner()))?,
+        };
 
         write_handle
             .wait(Durability::Applied)
@@ -214,7 +228,7 @@ impl MiniTsdb {
 
     /// Ingest a single series with samples.
     pub(crate) async fn ingest(&self, series: &Series) -> Result<()> {
-        self.ingest_batch(std::slice::from_ref(series)).await
+        self.ingest_batch(std::slice::from_ref(series), None).await
     }
 
     /// Flush pending data to the storage memtable (not yet durable).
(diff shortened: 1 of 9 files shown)


timeseries/src/minitsdb.rs
```diff
@@ -190,10 +209,16 @@ impl MiniTsdb {
...
-            .try_write(series_list.to_vec())
-            .await
-            .map_err(|e| map_write_error(e.discard_inner()))?;
+        let mut write_handle = match timeout {
+            Some(t) => handle
+                .write_timeout(series_list.to_vec(), t)
+                .await
+                .map_err(|e| map_write_error(e.discard_inner()))?,
+            None => handle
+                .try_write(series_list.to_vec())
```
micllam, line 209: Should this be `write` instead of `try_write` ?

timeseries/src/minitsdb.rs
```diff
@@ -170,6 +171,20 @@ impl MiniTsdb {
     }
 
     /// Ingest a batch of series with samples in a single operation.
+    pub(crate) async fn ingest_batch(&self, series_list: &[Series]) -> Result<()> {
```
agavra: these are pub(crate) so we can probably only have a single method `ingest_batch` that includes an optional timeout (basically make your inner one the only one). Same for `ingest_samples`

timeseries/src/minitsdb.rs
agavra: `try_write` is appropriate with no timeout otherwise we risk blocking forever. this matches log/vector

timeseries/src/timeseries.rs
```diff
@@ -140,6 +140,12 @@ impl TimeSeriesDb {
         self.tsdb.ingest_samples(series).await
     }
 
+    /// Writes one or more time series, waiting up to `timeout` for space in
+    /// the write queue.
+    pub async fn write_timeout(&self, series: Vec<Series>, timeout: Duration) -> Result<()> {
```
agavra, line 172: we should have detailed rustdoc here since it's a public method. at a minimum we should link to `write` and explain the differences between them.

agavra, line 172: can we also add tests the confirm the behavior?

agavra (commented):
Thanks @micllam - some comments inline.

timeseries/src/minitsdb.rs
micllam: OK that makes sense thanks.

It looks like `VectorDb::write()` is currently using `write` instead of `try_write` IIUC.

timeseries/src/minitsdb.rs
micllam: Thanks for this suggestion. I've now updated accordingly, keeping only a single method with optional timeout.

There are some other methods where I didn't add an extra parameter as it didn't seem necessary for this current PR, e.g. `MiniTsdb::ingest()`, `Scraper::ingest_samples()`.

timeseries/src/timeseries.rs
micllam: I've updated the rustdoc. Happy to make further changes if required.

micllam: I've added a simple unit test that calls this method when the queue is not full.

In order to test what happens when the queue is full and the timeout elapses, would it be necessary to create a timeseries DB with a very small queue capacity? Otherwise, it might take a while to fill up with the default `10_000`?

micllam:
Thanks for your review @agavra.

I've addressed some of your comments and followed up with some questions inline.

I also rebased on main and edited the PR description slightly.

timeseries/src/timeseries.rs
agavra: the behavior isn't particularly interestign when the queue is not full 😆 yes, lets try and get it to fill up. I'm sure we can do that detemrinisitcally by using failpoints to prevent the queue from draining

timeseries/src/minitsdb.rs
agavra: Ah I checked log. Let's keep it as is for now and we can sort that out later.

timeseries/src/timeseries.rs
micllam: Indeed :smile: I'm still thinking about how to do that, but for now I've added a couple of tests for `MiniTsdb::ingest_batch()` where I control the queue capacity and draining. Do you think these would be sufficient?

micllam:
Added tests for `MiniTsdb::ingest_batch()` and rebased on main.

timeseries/src/timeseries.rs
```diff
@@ -291,4 +324,34 @@ mod tests {
             "expected series to survive close without explicit flush"
         );
     }
+
+    #[tokio::test]
+    async fn should_write_series_with_timeout() {
```
agavra: nit: let's remove this test, not particularly interesting

timeseries/src/minitsdb.rs
```diff
@@ -254,3 +268,156 @@ fn map_write_error(e: WriteError) -> Error {
...
+        let result = mini
+            .ingest_batch(&s1, Some(Duration::from_millis(50)))
+            .await;
+
+        // then
+        assert!(result.is_ok(), "expected success, got {:?}", result);
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn should_fail_ingest_batch_when_timeout_too_short_for_drain() {
```
agavra, line 366: This test is close, but I don't like depending on wall clock time for any test behavior as the GHA runners are pretty flaky. I think it's fine to just test the path where the coordinator is paused and making sure that the inserts timeout (we don't need to test the success case). 

If we really want to test the success case we can consider using some manual timing tricks (e.g. via `#[tokio::test(start_paused = true)]`) but that is a bit more complicated but makes it deterministic. If you want to try you hand at that then feel free to take a stab, but don't spend too much time on it.

timeseries/src/tsdb.rs
```diff
@@ -1989,6 +2010,41 @@ mod tests {
         assert!(matches!(result.unwrap_err(), QueryError::InvalidQuery(_)));
     }
 
+    #[storage_test(merge_operator = OpenTsdbMergeOperator)]
+    async fn should_ingest_samples_with_timeout(storage: Arc<dyn Storage>) {
```
agavra: ni: I also don't think we need this test, not especially interesting and potentially flaky if the GHA poops itself for >5s

agavra (commented):
Thanks @micllam for your patience with the back and forth here, I think we're almost there, but I'd prefer to avoid tests that depend on wall clock timing. I left a suggestion for how we could avoid it, but if it proves to be complicated lets just test the negative secnario and call it a day.

timeseries/src/tsdb.rs
micllam: Removed.

timeseries/src/timeseries.rs
micllam: Removed.

timeseries/src/minitsdb.rs
micllam: Thanks for this suggestion. I've now changed it to use `start_paused = true` and AFAICT this is sufficient to avoid depending on wall clock time (I played around with different `sleep` values and checked that the test execution time increases or not depending on the `start_paused` setting).

If this is still not quite right, then I'll remove the success case and maybe come back to it in a follow-up PR.

micllam:
> Thanks @micllam for your patience with the back and forth here,

That's no problem. Thanks also for your patience and helpful suggestions.

I've now modified the test as suggested (no other changes to this test were required AFAICT) and I removed the two other tests as indicated.

I also rebased on main.

timeseries/src/minitsdb.rs
agavra: ah yes confirmed locally! I guess using tokio::sleep will manually move the clock forward. nice 👍

agavra (approved):
Thanks @micllam

PR #394, opened 2026-04-09
[timeseries] handle counter resets and extrapolation for rate function

### LINKED ISSUE #89: [tsdb] rate fn should handle counter resets and extrapolation
## Summary

The promql rate function needs to handle counter resets, and should extrapolate to the bounds of the evaluation window. For counter resets, it should observe whether the value decreased between samples, and then treat this as a counter resetting due to a process restart. So, for example if the sequence of samples is [1, 2, 3, 4, 3, 4, 5], then the total of the count is 1 + (2-1) + (3-2) + (4-3) + 3 + (4-3) + (5-4) = 9.  

## Motivation

Full PromQL support.

## Additional Context

See https://github.com/prometheus/prometheus/blob/main/promql/functions.go for the rate implementation

micllam:
## Summary

This modifies the rate function to handle counter resets and extrapolation, following the Prometheus logic.

## Related Issues

Fixes #89.

## Test Plan

Unit & promqltest tests added/updated.

## Checklist

- [x] Tests added/updated


timeseries/src/promql/promqltest/testdata/functions.test
```diff
@@ -237,6 +237,26 @@ eval instant at 50m rate(testcounter_zero_cutoff_total[20m])
...
+	{} 0.11
+
+# When the range end is close to the last sample (30m-25m=300s < 330s),
+# duration_to_end is not clamped.
+# rate = 300 * (1500+0+300) / 1500 / 1800 = 0.2
+eval instant at 30m rate(testrate_right_clamp_total[30m])
+	{} 0.2
+
+clear
+
```
cadonna, line 277: You need to wrap this in a `resume` and an `ignore`, otherwise the tests do not run.

You can verify that by running 

```
cargo test -p opendata-timeseries promql::promqltest::tests::should_pass_functions
```

It does not fail. Then change `{} 0.11` to `{} 1234.34` and run the tests again. It still does not fail.

BTW, it is always a good practice to first let a test fail and then fix it to verify the correctness of the test.

micllam: Thanks for spotting this and for the suggestion. I'll make sure to do that in future.

micllam:
Thanks for the review. I've added the `resume` and `ignore` blocks as per the comment.

I also rebased on main.

timeseries/src/promql/promqltest/testdata/functions.test
```diff
@@ -237,6 +239,28 @@ eval instant at 50m rate(testcounter_zero_cutoff_total[20m])
...
 
+# Tests for right-side extrapolation threshold.
+# 6 samples at 5m intervals: last sample at t=25m, values 0..300.
+# avg_duration=300s, threshold=330s.
+load 5m
+	testrate_right_clamp_total	0+60x5
+
+# When the range end is far past the last sample (50m-25m=1500s >= 330s),
+# duration_to_end is clamped to avg_duration/2 (150s).
+# rate = 300 * (1500+0+150) / 1500 / 3000 = 0.11
```
cadonna: This explanations are quite hard to understand without knowledge about the code. Could you please try to describe the tests on a more abstract level?

timeseries/src/promql/functions.rs
```diff
@@ -763,6 +763,20 @@ impl RangeFunctionAdapter<'_> {
         self.inner
             .apply(PromQLArg::RangeVector(samples), eval_timestamp_ms)
     }
+
+    pub(crate) fn apply_with_range(
```
cadonna: Since we have this function now, would it make sense to get rid of `apply()` above?

timeseries/src/promql/functions.rs
```diff
@@ -989,42 +1003,91 @@ impl PromQLFunction for DateTimeFunction {
     }
 }
 
+/// Computes counter-reset-corrected increase for a sample series.
+/// Walks through samples and accumulates previous values at each reset point.
+fn counter_reset_increase(values: &[Sample]) -> f64 {
```
cadonna: Some more documentation with examples would make this function better understandable. It took me a bit to understand the details. The thing that I missed is that when a value decreases, it is assumed that the counter restarts at zero. Basically, first it computes the delta between the first and the last sample value:
```
A = x_(n-1) - x_0
``` 
If there is a reset it adds the value before the reset to `A`:
```
A = x_(n-1) - x_0 + x_i
``` 
 where `x_i` is the value before the reset. We can also change the formula to:
```
A = x_i - x_0 + x_(n-1)
```
So basically, the delta between the first sample and the value before the reset is computed and then the delta between 0 and the last sample value is added as the increase.

timeseries/src/promql/functions.rs
```diff
@@ -989,42 +1003,91 @@ impl PromQLFunction for DateTimeFunction {
...
+        if sample.value < prev_value {
+            result += prev_value;
+        }
+        prev_value = sample.value;
+    }
+    result
+}
+
+/// Computes extrapolated rate for a series as per Prometheus logic.
+fn extrapolated_rate(series: &EvalSamples) -> Option<f64> {
```
cadonna, line 1026: Also here some more docs with examples would be great. For example, that the extrapolation threshold is 110% of the average duration between samples and that after that threshold the samples are clamped to constant values.

cadonna (commented):
Thanks for the updates @micllam !

I reviewed the production code. The production code looks good!
I still need to look at the tests. In the meanwhile, I added some comments.

timeseries/src/promql/promqltest/testdata/functions.test
micllam: I've now updated the comments for these tests. Happy to make further changes if need be.

timeseries/src/promql/functions.rs
micllam: I decided to remove `apply_with_range` and instead added `create_eval_samples_with_range` that sets the range fields from the outset. This way, we can now use the `apply` function directly.

timeseries/src/promql/functions.rs
micllam: I've now added more details to the documentation.

I also modified the function slightly to return only the correction to the increase, rather than the increase itself, and kept the `last - first` calculation as part of `extrapolated_rate`, since that part will be common to other similar PromQL functions, e.g. `increase`, and it's only the correction part that is specific to counters.

timeseries/src/promql/functions.rs
micllam: I've updated this documentation and expanded the comment in the function regarding not extrapolating to negative values.

micllam:
Thanks for your review and comments, @cadonna. I've made some changes as described inline, and rebased on main.

timeseries/src/promql/functions.rs
```diff
@@ -2852,19 +2938,176 @@ mod tests {
...
+        assert!(
+            (result[0].value - expected).abs() < 1e-10,
+            "expected {}, got {}",
+            expected,
+            result[0].value
+        );
+    }
+
+    #[test]
+    fn should_extrapolate_rate_to_window_boundaries() {
```
cadonna: Here it would be great if we had tests where we need to:
- extrapolate to the start
- extrapolate to the end
- extrapolate to start and end
Maybe you can achieve that with a parametrized test?

timeseries/src/promql/functions.rs
```diff
@@ -2852,19 +2938,176 @@ mod tests {
...
+        assert!(
+            (result[0].value - expected).abs() < 1e-10,
+            "expected {}, got {}",
+            expected,
+            result[0].value
+        );
+    }
+
+    #[test]
+    fn should_limit_extrapolation_when_samples_far_from_boundary() {
```
cadonna: Also here it would be great to have tests where:
- start needs to be clamped
- end needs to be clamped
- both, start and end, need be to be clamped.

timeseries/src/promql/functions.rs
```diff
@@ -2852,19 +2938,176 @@ mod tests {
...
+        // result = 20.0 * 0.15 = 3.0
+        let expected = 3.0;
+        assert!(
+            (result[0].value - expected).abs() < 1e-10,
+            "expected {}, got {}",
+            expected,
+            result[0].value
+        );
+    }
+
```
cadonna, line 3121: Maybe also two mixed tests where:
- the start needs to clamped and end is extrapolated
- the end needs to clamped and start is extrapolated

timeseries/src/promql/functions.rs
```diff
@@ -3100,6 +3348,63 @@ mod tests {
         );
     }
 
+    #[test]
+    fn should_apply_rate_zero_point_counter_guard() {
```
cadonna: Also here:
- duration to zero applied to start
- duration to zero applied to end
- duration to zero applied to both, start and end

cadonna (commented):
Thanks for the updates, @micllam !

I just have some improvement proposals for the tests. Once that is done, we can merge.

timeseries/src/promql/functions.rs
micllam: I've now parametrized this test to cover these cases.

timeseries/src/promql/functions.rs
micllam: I've parametrized this test to cover these cases and the two mentioned below.

timeseries/src/promql/functions.rs
micllam: I've now included these two cases.

timeseries/src/promql/functions.rs
micllam: The duration to zero logic applies only to the start, but I've now parametrized this test to cover the two cases where the duration to zero guard fires:
- when start is extrapolated to boundary
- when start is extrapolated to avg duration between samples / 2

micllam:
Thanks @cadonna. I've now made the suggested changes to the tests. Happy to make further changes if needed.

cadonna (approved):
@micllam Thanks for the updates!

LGTM!

Thank you for your contribution!

PR #397, opened 2026-04-10
update vector RFCs to reflect switch to ANN tree+LIRE

rodesai:
update vector RFCs to reflect switch to ANN tree+LIRE


vector/rfcs/0001-storage.md
```diff
@@ -62,33 +63,37 @@ index structures are stored as key-value pairs in the LSM tree.
...
 │   │                                                     │   │
-│   │   Centroids:     [chunk 0] [chunk 1] ... [chunk N]  │   │
-│   │                  (loaded into HNSW for navigation)  │   │
-│   │                                                     │   │
-│   │   Posting Lists: centroid_id →                      │   │
+│   │   Centroids:     centroid tree depth                │   │
+│   │   Posting Lists: vector_id →                        │   │
 │   │                  [(vector_id, f32[dimensions])]     │   │
-│   │                                                     │   │
+│   │   Centroid Info: vector_id →                        │   │
```
agavra, line 80: nit: I assume that `vector_id` here is the centroid id? and `parent vector_ids` are the parent centroids? 

EDIT: I see the section below that specifies this. I'm curious why you decided to model vectors/centroids using the same data model?

> This allows for clearer logging and invariant checking throughout the indexer. It also allows for having a single ID for vectors thats usable in bitmaps.

Not sure what this means?

vector/rfcs/0001-storage.md
```diff
@@ -114,64 +118,117 @@ index structures are stored as key-value pairs in the LSM tree.
...
+   for driving high i/o parallelism when indexing requires reads, and amortizes some of the
+   index maintenance work over many vectors.
+1. Doing logical read-modify-writes of large values like centroid postings and inverted indexes as
+   SlateDB merges. Further, as another benefit of the aforementioned batching, vector applies
+   many such writes together in a single SlateDB `WriteBatch` to reduce merge overhead within
+   SlateDB and instead do the merges directly within the batch. Generally, merges done by SlateDB
+   are more expensive as they require deserializing to the Vector in-memory model, applying the
+   merge, and then reserializing.
 
 **Delete Operation:**
```
agavra, line 225: Can you add here a full explanation of how the tombstones get cleared? It's a bit unclear from the document (I asked Claude to index all the RFCs and answer it but it tripped up as well). Are we still relying on Compaction Filters or just merge operator application in compaction?

vector/rfcs/0005-ann-indexing.md
```diff
@@ -0,0 +1,663 @@
...
+  the index to improve performance.
+- Filtered search: to achieve strong recall for filtered search the ANN index should be able to
+  prune clusters based on filter expressions. This is an out-of-scope extension to this design.
+- Physically decoupling indexing from write. The RFC proposes implementing indexing batches of
+  writes. We won't cover how we can physically decouple these processes using stateless ingest
+  to a WAL, though this should be a straightforward subsequent improvement.
+
+## High Level Description of LIRE
+
+Let's assume a flat centroid structure for nwo. LIRE aims to do the following:
```
agavra:
```suggestion
Let's assume a flat centroid structure for now. LIRE aims to do the following:
```

vector/rfcs/0005-ann-indexing.md
```diff
@@ -0,0 +1,663 @@
...
+
+A Merge for a clusters with centroid `C` does the following:
+1. Remove `C` from the centroid index
+2. For every vector `V` that was assigned to `C`, assign `V` to its current nearest centroid.
+
+## Extension to a Tree of Centroids
+
+We can extend our flat set of centroids into a tree by splitting the centroids themselves into 
+clusters with a new layer of centroids when too many accumulate. Then these new centroids can 
+also be maintained using LIRE. Lets take a look at how this structure would evolve.
```
agavra, line 119: That's very elegant!

vector/rfcs/0005-ann-indexing.md
```diff
@@ -0,0 +1,663 @@
...
+
+Lets say that I want to limit clusters to < N vectors. Lets start with an initial db with a root
+that has 1 initial centroid C_0 with all its dimensions set to 0. After N vectors are inserted
+C_0 will be split so the root now has 2 centroids. Inserts keep coming, and eventually the root 
+grows to N centroids. Now the root is too large, so we need to split it. We can do this by running
+kmeans over the centroids to compute new centroids for a new level of the tree. Let's call the new
```
agavra, line 125: I wonder if reassignment in the centroid index should be exact. I'm worried about what it would mean for using the same heuristic as we do for data vector reassignment if it means centroids are also approximate

vector/rfcs/0005-ann-indexing.md
```diff
@@ -0,0 +1,663 @@
...
+level level 2 and the old level level 1. So now we have 2 layers of centroids - the old ones at
+level 1, and 2 (or some K) new centroids at level 2. Each level 1 centroid is assigned to a level 2
+centroid's cluster. When level 1 centroids are split/merged, they are added to/removed from clusters
+in level 2. As the size of level 1 grows, the postings in level 2 grow, so we run LIRE in level 2
+to balance the size of the level 2 clusters. And so on until there are 10 level 2 centroids, at
```
agavra:
```suggestion
to balance the size of the level 2 clusters. And so on until there are `N` level 2 centroids, at
```
I assume this is what you meant?

vector/rfcs/0005-ann-indexing.md
```diff
@@ -0,0 +1,663 @@
...
+To a large degree, recall is a tunable tradeoff with performance. For example, you could 
+configure queries to search more clusters (the beam width described above) to increase the 
+likelihood of finding close results at the cost of more query work. Similarly, you can tune the 
+size of the neighbourhood that’s searched when evaluating reassignments after a split.
+
+On the other hand, implementation or synchronization bugs can cause the index to become 
+permanently degraded. We want Vector to ensure that:
+1. Vectors are never dropped from the index (so every vector should be assigned to a posting)
+2. Preserve NPA - so each vector should be assigned to its nearest neighbour (as defined by the 
+   centroid index).
```
agavra: I thought we just mentioned this was best effort? Or is it only considered preserved within a "neighborhood"?

vector/rfcs/0005-ann-indexing.md
```diff
@@ -0,0 +1,663 @@
...
+3. Reads always observe the index in a state that preserves the above 2 properties.
+
+We will achieve this by doing indexing in batches decoupled from ingest and query. Ingest appends
+vectors to a queue. For now this will be an in-memory queue using the Write Coordinator, so writes
+don't become durable until the indexer processes them. In the future, we can write to a durable
+WAL first for lower-latency durability. The indexer picks up batches of writes and indexes them.
+The output of indexing is an atomic `WriteBatch` that always advances the db between snapshots
+that preserve properties (1) and (2) from above. Reads always read from one of these snapshots to 
+ensure property (3). To support low-latency read-your-writes we can allow reads to observe vectors
```
agavra: why wouldn't we default to reads of snapshot + in-memory? I assume the in-memory scan would be <10k vectors which you benchmarked to be ~3ms?

vector/rfcs/0005-ann-indexing.md
```diff
@@ -0,0 +1,663 @@
...
+
+## Configuration
+
+- `root_threshold_vectors`: The max number of root vectors before which a root split is triggered.
+- `split_search_neighbourhood`: The number of neighbouring centroids whose postings will be 
+  considered for reassignment after a split.
+- `split_threshold_vectors`: The minimum number of vectors assigned to a centroid before the 
+  Write Coordinator triggers a spit.
+- `merge_threshold_vectors`: The maximum number of vectors assigned to a centroid before the 
+  Write Coordinator triggers a merge.
```
agavra, line 213: do these apply to the intermeidate stage LIREs as well (on the centroids)?

vector/rfcs/0005-ann-indexing.md
```diff
@@ -0,0 +1,663 @@
...
+
+**Inputs**
+- `writes`: The set of writes to be applied to the db. 
+
+**Computation**
+- First compact the batch so the last write for each `external_id` wins. This prevents multiple
+  writes to the same logical vector in a single indexing round from doing redundant work.
+- For all writes, compute the target centroid assignment by running ANN search against the current
+  centroid tree. 
+- For upserts, read the old `VectorIndexData`. This is the source of truth for:
```
agavra, line 384: an optimization for another time, would be interesting to start prefetching these while we're building up the batch

agavra (approved):
@rodesai I'm super impressed with this, very elegant design and written with a clarity that makes me feel smart because I can follow it so easily 😆 bravo! Mostly nits and clarification questions inline.

vector/rfcs/0001-storage.md
rodesai: They don't have the same data model. I'm just proposing using a common type for the ID that encodes both the level and a unique identifier. The model for the details about a centroid vs data vector is still different. With a single id type we always use the same type when referencing vectors (both centroids and data vectors) in the various storage keys, and in posting lists and inverted index bitmaps.

> Not sure what this means?

Invariant checking is easier because you can always observe the level from `VectorId`. So you can for example always assert that l_x vectors are going into an l_x+1 posting. Its also nice to have a standard format for the vector ids in log msgs (`<level>:<id number>`)

vector/rfcs/0001-storage.md
rodesai: Ah yeah I'll add a note. We need a compaction filter that clears the tombstones when compacting into the last level.

vector/rfcs/0005-ann-indexing.md
rodesai: The heuristics are not what makes the resulting index approximate. They are sufficiently strict to determine whether or not its possible that a vector might now be closer to another centroid. The structure is approximate because it intentionally probes a subset of the clusters to find the nearest neighbors. The only way to make it exact is to do an exhaustive search. That said, I do think it's worth playing with different nprobes at different levels of the tree or enabling replication of boundary vectors for some levels of the tree. I'll add that to the future improvements section.

vector/rfcs/0005-ann-indexing.md
rodesai: It's best effort. I'll clarify

vector/rfcs/0005-ann-indexing.md
rodesai: Yeah that's what I'm saying here. The reads query the ANN index in the snapshot and then can factor in the in-memory delta that's not yet applied. We haven't implemented that yet - but it makes sense for that to be the default.

vector/rfcs/0005-ann-indexing.md
rodesai: yeah

vector/rfcs/0005-ann-indexing.md
rodesai: yeah I think there's lots of opportunity to pipeline work in the indexer. Right now all the "heavy" work is parallel but synchronous - so indexing blocks until everything in the batch finishes a given task (e.g. assigning all the vectors in the batch to centroids, reading metadata fro storage, etc).

PR #398, opened 2026-04-10
common: add status method in Storage trait for health checks, #165

### LINKED ISSUE #165: Use SlateDB's DB::status() for storage health checks
## Summary

The Log and Timeseries have `/healthy` and `/ready` end points which check if the underlying Slate is healthy as part of their response. Today, these methods are either no-ops or use contrived queries to check the underlying slate. See https://github.com/slatedb/slatedb/pull/1234

PR454D:
## Summary

What does this PR do?
Added status method in [Storage](https://github.com/opendata-oss/opendata/blob/30d8bad7385e147aff50ae9fda2f535fdce6f817/common/src/storage/mod.rs#L250).

Implement status for Inmemory & SlateDB storage backend, update the `/health` endpoint in `opendata-log` and `opendata-timeseries` to delegate to the storage status.

## Related Issues

Fixes: #165 

## Test Plan

How was this tested? Existing Unit tests.

## Checklist

- [X] Tests added/updated

### DIFF
--- a/common/src/storage/mod.rs
+++ b/common/src/storage/mod.rs
@@ -368,4 +368,7 @@ pub trait Storage: StorageRead {
     /// `DbStatus::durable_seq`. For in-memory storage, writes are immediately
     /// "durable" so the watermark matches the latest written seqnum.
     fn subscribe_durable(&self) -> tokio::sync::watch::Receiver<u64>;
+
+    /// Returns the status of the storage engine.
+    async fn status(&self) -> StorageResult<()>;
 }
--- a/common/src/storage/slate.rs
+++ b/common/src/storage/slate.rs
@@ -343,6 +343,15 @@ impl Storage for SlateDbStorage {
         self.db.flush().await.map_err(StorageError::from_storage)?;
         Ok(())
     }
+
+    async fn status(&self) -> StorageResult<()> {
+        // fetch status from slatedb
+        let mut status = self.db.as_ref().subscribe();
+        status
+            .changed()
+            .await
+            .map_err(|e| StorageError::Storage(e.to_string()))
+    }
 }
 
 impl From<Ttl> for slatedb::config::Ttl {
--- a/log/src/server/handlers.rs
+++ b/log/src/server/handlers.rs
@@ -231,8 +231,13 @@ pub async fn handle_metrics(State(state): State<AppState>) -> String {
 /// Handle GET /-/healthy
 ///
 /// Returns 200 OK if the service is running.
-pub async fn handle_healthy() -> (axum::http::StatusCode, &'static str) {
-    (axum::http::StatusCode::OK, "OK")
+pub async fn handle_healthy(
+    State(state): State<AppState>,
+) -> (axum::http::StatusCode, &'static str) {
+    match state.log.as_ref().status().await {
+        Ok(_) => (axum::http::StatusCode::OK, "OK"),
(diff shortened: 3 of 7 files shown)


PR454D:
The **timeseries** implementation for storage health check is done in `/-/ready` endpoint.. whereas it is done in `/-/health` for **Logdb**. Usually this should be done in a same endpoint, but adding this in the /health endpoint introduces test case changes, as the tests use a [router without state](https://github.com/opendata-oss/opendata/blob/30d8bad7385e147aff50ae9fda2f535fdce6f817/timeseries/src/server/http.rs#L636).

common/src/storage/slate.rs
```diff
@@ -343,6 +343,15 @@ impl Storage for SlateDbStorage {
         self.db.flush().await.map_err(StorageError::from_storage)?;
         Ok(())
     }
+
+    async fn status(&self) -> StorageResult<()> {
+        // fetch status from slatedb
+        let mut status = self.db.as_ref().subscribe();
+        status
+            .changed()
+            .await
```
agavra, line 352: instead of changed we should just use `borrow()` to get the current one

common/src/storage/mod.rs
```diff
@@ -368,4 +368,7 @@ pub trait Storage: StorageRead {
     /// `DbStatus::durable_seq`. For in-memory storage, writes are immediately
     /// "durable" so the watermark matches the latest written seqnum.
     fn subscribe_durable(&self) -> tokio::sync::watch::Receiver<u64>;
+
+    /// Returns the status of the storage engine.
+    async fn status(&self) -> StorageResult<()>;
```
agavra, line 373: I'm not sure I understand this API? it seems to return `()`, how does that tell us the status? note that slatedb can return a CloseReason that is an error, which would be an unhealthy status as well

log/src/server/handlers.rs
```diff
@@ -265,10 +270,14 @@ mod tests {
...
     async fn should_return_ok_for_healthy() {
+        let log = Arc::new(LogDb::open(test_config()).await.unwrap());
+        let metrics = Arc::new(Metrics::new());
+        let state = AppState { log, metrics };
         // given/when
-        let (status, body) = handle_healthy().await;
+        let (status, body) = handle_healthy(State(state)).await;
 
         // then
+        // should return OK in case of InMemory Log backend
```
agavra, line 280: we should test a slatedb instance, even an in-memory backed one, in order to make sure this works with `DbStatus` which is the interesting part. We should also test some error scenarios to make sure it works when not healthy

agavra (commented):
Thanks @PR454D - I left some comments

common/src/storage/mod.rs
PR454D: Makes sense to use some kind of information attached with status. Is it okay to just use DbStatus from slateDB, so like this `StorageResult<slatedb::DbStatus>`.

PR454D: But in case of InMemory storage this creates a dependency on slatedb.. and we cannot create No-Op for in-memory storage..

agavra:
@PR454D thanks for your work on this PR, I think it's best to just close it out for now. I have a PR staged to remove the `Storage` abstraction altogether so we don't need to constant wire in new methods to our Storage traits (see discussion on discord): https://discord.com/channels/1450884405219561628/1497295148906971237

PR454D:
Sure, I read that discussion. and it makes sense to integrate directly with slatedb.. Closing this.

PR #402, opened 2026-04-14
timeseries: skip MiniTsdb::load for query readers

agavra:
## Summary

`/labels` and `/label/{name}/values` queries were taking 15+ seconds even with a warm block cache. Tracing showed ~13s idle time in `find_labels`, all spent in `query_reader` sequentially calling `get_bucket` → `MiniTsdb::load` for each bucket.

`MiniTsdb::load` scans the series dictionary and spins up a write coordinator per bucket. The read path never uses either — `MiniQueryReader` only needs a bucket and a storage snapshot. The query cache (LRU of 50 `MiniTsdb` instances) existed to amortize this load cost, but the load was unnecessary in the first place. The read-only `TimeSeriesDbReader` already does the right thing: its cache stores `MiniQueryReader` directly, and "loading" one is just wrapping a bucket + snapshot reference.

This PR:
- Constructs `MiniQueryReader` directly from the storage snapshot on cache miss
- Checks the ingest cache first so unflushed data stays visible for active buckets
- Removes the `query_cache` field from `Tsdb` since it no longer serves a purpose

Evaluator microbenchmarks show no regression (within noise on all three benchmarks).

## Test plan
- [x] `cargo test -p opendata-timeseries` — 711 passed, 0 failed
- [x] `cargo bench --bench evaluator_micorbench` — no regression vs main
- [ ] Deploy and verify `/labels` latency drops from ~15s to <2s


apurvam:
Does the impact show up in the existing criterion benchmarks? Also does this make it harder for queries to read unflushed data from the write coordinator?

agavra:
> Does the impact show up in the existing criterion benchmarks? 

I haven't run them but I don't imagine it does. This prevents fetching data we don't need (e.g. the series dictionary) for every bucket.

> Also does this make it harder for queries to read unflushed data from the write coordinator?

It shouldn't, the write coordinator communicates independently of the MiniTsdb. Though we don't have the wired in at all at the moment and I probably wouldn't recommend we wire it in (the delay of a few seconds of buffering in the write coordinator is OK for timeseries IMO and we don't need read-your-writes)

agavra:
Digging into the test failure, I thought I ran things to make sure this is safe but it might not be. Please hold 🙏

agavra:
@apurvam ran the benchmark, no regression

```
  ┌────────────────────────────┬───────────┬───────────┐
  │         Benchmark          │   main    │  branch   │
  ├────────────────────────────┼───────────┼───────────┤
  │ subquery_label_cache       │ 103.74 ms │ 102.42 ms │
  ├────────────────────────────┼───────────┼───────────┤
  │ warm_range_vector_selector │ 286.39 ms │ 275.83 ms │
  ├────────────────────────────┼───────────┼───────────┤
  │ warm_range_aggregation     │ 277.16 ms │ 260.42 ms │
  └────────────────────────────┴───────────┴───────────┘
```

I also fixed the write coordinator problem by reading from the writer cache if it exists, and removing the query_cache altogether, which was actually not helpful since it wasn't caching any materialized indexes. We can consider caching materialized indexes if the block cache isn't good enough for it.

timeseries/src/tsdb.rs
```diff
@@ -566,68 +556,53 @@ impl Tsdb {
...
-            return Ok(mini);
-        }
-
-        // 2. Check query cache
-        if let Some(mini) = self.query_cache.get(&bucket).await {
-            return Ok(mini);
-        }
-
-        // 3. Load from storage into query cache (NOT ingest cache)
-        let mini = Arc::new(MiniTsdb::load(bucket, self.storage.clone()).await?);
```
agavra, line 582: This was the expensive part that wasn't necessary

timeseries/src/tsdb.rs
```diff
@@ -566,68 +556,53 @@ impl Tsdb {
...
+    async fn build_readers(
+        &self,
+        snapshot: &Arc<dyn common::storage::StorageSnapshot>,
+        buckets: Vec<TimeBucket>,
+    ) -> Vec<(TimeBucket, MiniQueryReader)> {
+        let mut readers = Vec::with_capacity(buckets.len());
         for bucket in buckets {
-            let mini = self.get_bucket(bucket).await?;
-            let reader = mini.query_reader();
+            let reader = if let Some(mini) = self.ingest_cache.get(&bucket).await {
```
apurvam, line 598: This is the cache populated by the write coordinator?

agavra: We have a write coordinator per `MiniTsdb`, whatever does ingest uses `Tsdb::get_or_create_for_ingest` to access that cache.

apurvam (approved):
This change makes sense to me. It seems like the query_cache was a workaround for a structurally needless operation (loading buckets + series dictionaries for old buckets which were never being written to anyway).

PR #411, opened 2026-04-16
[common] configure foyer hybrid cache

### LINKED ISSUE #405: Tune foyer hybrid cache: WriteOnInsertion policy and Large engine options
> **Note:** This issue tracks work already implemented locally on a prototype branch. The implementation section below is detailed context so that a future agent (or contributor) can cleanly reimplement this as a standalone change on a fresh checkout.

## Summary

The foyer `HybridCacheBuilder` in `create_block_cache_from_config` uses default engine settings (`Engine::large()`) and the default admission policy. We should switch to `WriteOnInsertion`, configure the `LargeEngineOptions` explicitly, and expose these knobs in our config layer so operators can tune them per deployment.

## Motivation

- **`WriteOnInsertion`**: The default policy may skip writing blocks to the disk cache if the admission filter rejects them. For our use case we want every block that enters the memory cache to also be persisted to the disk tier — especially metadata blocks that are expensive to re-read from object storage.
- **Explicit `LargeEngineOptions`**: The defaults for flushers (1), buffer pool size, and submit queue threshold are conservative. Bumping flushers to 4 with a 256 MB buffer pool and 1 GB submit queue threshold better matches our I/O patterns and avoids backpressure during bulk cache warming or compaction.
- **Configurability**: These values shouldn't be hardcoded — different deployments (reader replicas vs. ingest nodes, local SSD vs. network disk) will want different settings.

### 1. Update `HybridCacheBuilder` in `create_block_cache_from_config`

In `common/src/storage/factory.rs`, change the builder chain from:

```rust
.storage(Engine::large())
```

to:

```rust
.with_policy(HybridCachePolicy::WriteOnInsertion)
.storage(Engine::Large(
    foyer::LargeEngineOptions::default()
        .with_flushers(4)
        .with_buffer_pool_size(256 * 1024 * 1024)
        .with_submit_queue_size_threshold(1024 * 1024 * 1024),
))
```

This requires adding `HybridCachePolicy` to the existing `use foyer::{...}` import.

### 2. Make settings configurable

Extend `FoyerHybridCacheConfig` (in `common/src/storage/config.rs`) to include fields for the new knobs (flushers, buffer pool size, submit queue threshold, write policy) with sensible defaults matching the values above. Wire those through to the builder so they can be set via the YAML config.

Note: this pairs well with the foyer metrics bridge (#403) — once metrics are visible, we can validate whether these tuning values are right for production.

agavra:
Expose write policy, flushers, buffer pool size, and submit queue threshold as configurable fields on FoyerHybridCacheConfig. Defaults to WriteOnInsertion with 4 flushers and buffer pool derived from memory_capacity / 32.

fixes #405

### DIFF
--- a/common/src/storage/config.rs
+++ b/common/src/storage/config.rs
@@ -62,6 +62,20 @@ pub enum BlockCacheConfig {
     FoyerHybrid(FoyerHybridCacheConfig),
 }
 
+/// Write policy for foyer's hybrid cache.
+///
+/// Controls when entries are written to the disk tier.
+#[derive(Default, Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
+pub enum FoyerWritePolicy {
+    /// Write to disk when an entry is inserted into the memory cache.
+    /// Ensures every cached block is also persisted to the disk tier.
+    #[default]
+    WriteOnInsertion,
+    /// Write to disk only when an entry is evicted from the memory cache.
+    /// This is foyer's default policy.
+    WriteOnEviction,
+}
+
 /// Configuration for foyer's hybrid (memory + disk) block cache.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct FoyerHybridCacheConfig {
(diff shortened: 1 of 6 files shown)


apurvam (approved):
LGTM. We should figure out how to expose this as user configuration, along with the concurrency settings: https://github.com/opendata-oss/opendata/blob/main/timeseries/src/promql/pipeline.rs#L718

agavra:
> We should figure out how to expose this as user configuration

I expose it in this PR, I will add it to the docs later today. See for example:
```
block_cache:
  type: FoyerHybrid
  memory_capacity: 4294967296
  disk_capacity: 10737418240
  disk_path: /mnt/nvme/cache
  write_policy: WriteOnEviction
  flushers: 2
  buffer_pool_size: 134217728
  submit_queue_size_threshold: 536870912
```

PR #412, opened 2026-04-16
add startup cache warmer to pre-populate block cache

agavra:
Temporary workaround until SlateDB's CacheManager is available (see https://github.com/slatedb/slatedb/pull/1520). Spawns a one-off task at startup that scans recent time bucket key ranges through StorageRead, populating the block cache as a side effect.

### DIFF
--- a/timeseries/src/reader.rs
+++ b/timeseries/src/reader.rs
@@ -144,7 +144,7 @@ impl QueryReader for ReaderQueryReader {
 /// let result = reader.query("rate(http_requests_total[5m])", None).await?;
 /// ```
 pub struct TimeSeriesDbReader {
-    pub(crate) storage: Arc<dyn StorageRead>,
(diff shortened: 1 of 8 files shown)


timeseries/src/reader.rs
```diff
@@ -144,7 +144,7 @@ impl QueryReader for ReaderQueryReader {
 /// let result = reader.query("rate(http_requests_total[5m])", None).await?;
 /// ```
 pub struct TimeSeriesDbReader {
-    storage: Arc<dyn StorageRead>,
+    pub(crate) storage: Arc<dyn StorageRead>,
```
rodesai: I think this doesn't need to be `pub(crate)` since the storage is accessed via `storage_read`

rodesai (approved):
LGTM!

PR #409, opened 2026-04-16
[timeseries] Implement PromQL `increase` function

micllam:
## Summary

This builds on the changes from #394 to implement the PromQL `increase` function with counter reset handling and extrapolation.

## Related Issues

Relates to #89.

## Test Plan

Unit tests added.
Existing PromQL tests un-ignored.

## Checklist

- [x] Tests added/updated
- [x] `cargo fmt` and `cargo clippy` pass
- [ ] Documentation updated (if applicable)

PR #416, opened 2026-04-17
[timeseries] Implement PromQL `delta` function

micllam:
## Summary

This builds on the work from #394 and #409 to implement the PromQL `delta` function, which is similar to `increase` with the following differences:
- no counter reset handling (`delta` should be applied to gauges)
- extrapolated values can be negative

## Related Issues

No specific issue but this is a natural follow-up to the PRs mentioned above.

## Test Plan

Unit tests added.
Those existing PromQL tests for `delta` that are non-histogram have been un-ignored.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/timeseries/src/promql/promqltest/testdata/functions.test
+++ b/timeseries/src/promql/promqltest/testdata/functions.test
@@ -337,18 +337,27 @@ eval instant at 20m irate(http_requests_histogram{path="/g"}[20m])
 
 clear
 
-# Tests for delta().
+resume
+
+# Tests for delta() (non-histogram).
 load 5m
 	http_requests{path="/foo"}	0 50 100 150 200
 	http_requests{path="/bar"}	200 150 100 50 0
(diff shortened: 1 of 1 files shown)


agavra:
@micllam sorry for the jitter, but we just merged a full new query engine so it's going to be a bit of work to rebase this on the new stuff. LMK if you have the bandwidth to do that

micllam:
Hi @agavra, that's no problem. I noticed there was some work going on with that and figured I may need to rebase this PR. I'll aim to do that by end of next week; should be a good chance to get some familiarity with the changes :)

micllam:
Hi @agavra, it looks like the `delta` function is now implemented in main, so the only change I've kept from this PR is the small adjustment to promqltest to include the non-histogram delta function tests.

agavra (approved):
Ah looks like Claude implemented the new function without me even noticing 😅 - thanks for the added test coverage @micllam!

PR #419, opened 2026-04-20
[timeseries] RFC 0007: PromQL execution engine

agavra:
Extracts RFC 0007 for standalone review. Proposes rewriting the PromQL execution pipeline as a columnar, step-major, pull-based perator tree with explicit logical and physical planning stages.

The implementation lives on the [`columnar_execution`](https://github.com/opendata-oss/opendata/tree/columnar_execution) branch and will land separately once this RFC is approved.

### DIFF
--- a/timeseries/rfcs/0007-promql-execution.md
+++ b/timeseries/rfcs/0007-promql-execution.md
@@ -1,0 +1,296 @@
+# RFC 007: PromQL Execution Engine
+
+**Status**: Draft
+**Authors**: [Almog Gavra](https://github.com/agavra)
+
+## Summary
+
+This RFC proposes a rewrite of the core query execution pipeline for timeseries to a columnar,
+timestamp-major, pull-based operator pipeline with proper logical and physical planning steps.
+
+## Motivation
+
+There are the following problems in the original implementation:
+
+1. **Per-Step Repeated Work:** the implementation loops on every step during the query’s effective
+   range and re-runs the evaluator evaluation. This pipeline does not properly amortize repeated
+   work like label hashing, metadata lookups and per-series iteration. We papered over some of this
+   using intra-query caches, but this fix was imperfect.
+2. **Row-Oriented Compu**t**ation:** the data model stored samples keyed by label. This meant that
+   each step evaluation would binary search for the closest matching sample (in the step’s
+   neighborhood according to lookback window), and perform an indexed array lookup. This not only
+   repeats work that is unnecessary for every step of evaluation it also is extremely inefficient
+   for memory cache lines.
+3. **Concurrency Is Implicit:** there is parallelism encoded within the data fetching using the
+   semaphores, but the actual computation is serial and there is no easy way to speed that up.
+4. **No Physical Planning Stage:** this not only makes it difficult to hook in optimizations like
+   projection pushdown or range coalescing into SlateDB if/when we support that but it also makes it
+   harder in the future for us to add things like a fan-out, multi-node query engine.
+
+The new proposed pipeline brings `timeseries` closer to the SOTA implementation for query
+evaluation.
+
+## Goals
+
+- One execution model for instant and range queries
+- Columnar, step-major data flow from storage to result
+- Explicit physical plan with a small and testable operator set
+- Pipelined execution with async operators so we can stream a single query and saturate both CPU and
+  I/O
+
+## Non-Goals
+
+- Distributed execution
+- Cost-based optimizations
+- PromQL language extensions
+- Native Histograms
+
+## Architecture
+
+```
+HTTP / embedded API
+        │
+        ▼
+┌───────────────────┐     parse PromQL → parser::Expr
+│    Parse          │
+└──────┬────────────┘
+       ▼
+┌───────────────────┐     AST rewrites: constant fold, label pushdown,
+│    Logical Plan   │     CSE
+│    + Optimizer    │
+└──────┬────────────┘
+       ▼
+┌───────────────────┐     LogicalPlan → tree of PhysicalOp
+│    Physical Plan  │     Bind storage, resolve step grid, compute
+│                   │     series schema per op
+└──────┬────────────┘
+       ▼
+┌───────────────────┐     Pull StepBatch streams from the root
+│    Executor       │     Each op: next() → Poll<Option<StepBatch>>
+└──────┬────────────┘
+       ▼
+┌───────────────────┐     StepBatch → QueryValue (instant or range)
+│    Result Shape   │
+└───────────────────┘
+```
+
+## Planning Stage
+
+Planning is split into two stages with a hard boundary between them: `Expr → LogicalPlan` is the *
+*rewritable** stage, `LogicalPlan → PhysicalPlan` is the compiling stage. Everything that wants to
+inspect, modify, or explain a query touches `LogicalPlan`; once we cross into physical, the tree is
+opaque and only the executor talks to it.
+
+We introduce a dedicated logical plan instead of reusing the AST module we depend on because it
+allows us to attach information such as the query’s timestamps and reinterpret concepts in more
+structure form. The conversion is pretty minor, though it also enables us to power `EXPLAIN`
+queries.
+
+The physical plan is a tree of `Operators` that are each individually testable and have the same
+input/output types (defined below).
+
+## Physical Plan Execution
+
+### Core Data Model
+
+The universal on-the-wire shape between operators is a `StepBatch`: a contiguous range of output
+steps for a contiguous chunk of series, with data stored in columnar representation. This data model
+gives us two major benefits:
+
+1. It is columnar with a step-major, series-minor format which allows us to quickly compute
+   aggregations within a single time-step.
+2. It is chunked, allowing us to pipeline and stripe execution across both dimensions (time and
+   labels).
+
+```
+StepBatch  (step_range = 2..6, series_range = 1..4)
+──────────────────────────────────────────────────────────────────
+
+step_timestamps: Arc<[i64]>      series: Arc<SeriesSchema>
+┌──────┐                         ┌────────────────┐
+│ t₀   │                         │ 0: {job=api}   │
+│ t₁   │                         ├────────────────┤
+├──────┤ ◄─ step_range.start=2   │ 1: {job=web}   │ ◄─ series_range.start=1
+│ t₂   │                         │ 2: {job=db}    │
+│ t₃   │                         │ 3: {job=cache} │
+│ t₄   │                         ├────────────────┤
+│ t₅   │                         │ 4: {job=edge}  │ ◄─ series_range.end=4
+├──────┤ ◄─ step_range.end=6     │ 5: {job=lb}    │
+│ t₆   │                         └────────────────┘
+│ t₇   │                         (fixed at plan time)
+└──────┘
+
+           web         db          cache
+         ┌──────────┬──────────┬──────────┐
+    t₂   │ 12.3  ✓  │  —    ✗  │  7.1  ✓  │
+         ├──────────┼──────────┼──────────┤
+    t₃   │ 12.9  ✓  │ 44.0  ✓  │  7.4  ✓  │
+         ├──────────┼──────────┼──────────┤
+    t₄   │ 13.1  ✓  │ 45.2  ✓  │  —    ✗  │
+         ├──────────┼──────────┼──────────┤
+    t₅   │ 13.4  ✓  │ 46.0  ✓  │  7.8  ✓  │
+         └──────────┴──────────┴──────────┘
+```
+
+Each batch is a rectangle of `K series × N steps`. Default `N ≈ 64`, `K ≈ 512` giving a ~256 KB
+working set that stays in L2. Both step-wise ops (`sum by` across series at one step) and time-wise
+ops (`rate` across steps of one series) have cache-friendly traversal. Operators that need the other
+axis insert a `Rechunk` breaker.
+
+Here is the actual struct. Note that we us validity bitsets to distinguish between 0/null values.
+
+```rust
+struct StepBatch {
+    // Time axis — shared by all series in the batch. One allocation, reused
+    // across many batches within a query.
+    step_timestamps: Arc<[i64]>,          // length = step_count
+    step_range: Range<usize>,             // slice into step_timestamps
+
+    // Series axis — schema is fixed at plan time.
+    series: Arc<SeriesSchema>,            // labels, fingerprint, stable index
+    series_range: Range<usize>,           // slice into series schema
+
+    // Value columns — SoA, float-only in v1.
+    values: Vec<f64>,                     // step_count × series_count, row-major by step
+    validity: BitSet,                     // same shape; absent = no sample in lookback window
+}
+```
+
+### Data Source / Storage API
+
+The interaction between the query execution pipeline and storage happens over the `SeriesSource`
+trait with two methods:
+
+- `resolve(selector, time_range)` to handle metadata lookups and resolve series info
+- `samples(SamplesRequest)` to retrieve the raw samples
+
+The storage is unaware of PromQL concepts such as lookback deltas, offsets, step alignment, etc… all
+of that is handled by the leaf operators (`VectorSelector` and `MatrixSelector`).
+
+Storage is accessed both during planning and execution. In the planning phase, `build_physical_plan`
+walks the logical tree and, for each selector, calls `resolve(..)` and drains the stream into the
+plan-time series roster (`Arc<SeriesSchema>`). This is the *only* place series identity is resolved;
+after this, operators index series by `u32` slot. `resolve` runs eagerly and synchronously with
+regards to physical-plan construction which means the planning time includes metadata latency.
+
+The execution `VectorSelector` and `MatrixSelector` operators call `samples(..)` with the
+pre-resolved series list and the absolute time window they need. Batches come back streamed and the
+leaf reshapes them into `StepBatch`es that the rest of the tree polls. No other operator
+communicates with the storage layer.
+
+### Concurrency & Parallelism
+
+There are a few areas where we can introduce parallelism and concurrency:
+
+**Inside the storage fetching.** Selectors typically span multiple storage buckets and since buckets
+are disjoint keyspaces, fan-out is safe). The implementation uses a buffered stream with max
+concurrency 32 for both metadata and sample streams. These constants mirror v1's
+`METADATA_STAGE_READAHEAD` / `SAMPLE_STAGE_READAHEAD` so observed concurrency is comparable.
+
+**Pipelining operator execution.** A selector leaf produces I/O-bound batches but downstream ops are
+CPU-bound. `ConcurrentOp` decouples them: the child runs on a spawned tokio task and pushes into a
+bounded mpsc channel (default bound = 4), giving implicit back-pressure since if the consumer
+stalls, the child blocks on `send`. The physical planner inserts one wrapper per selector leaf whose
+resolved series count is ≥ 64.
+
+**Vertical sharding.** We defer using the `Coalesce` operator in the initial implementation but the
+wiring is stubbed in `parallelism` and `coalesce_max_shards` defaults to `0`. Needs end-to-end
+correctness work (per-series independence above the leaf isn't free) before it's turned on.
+
+**No global permit layer.** v1 had a separate `QueryReaderEvalCache` metadata/sample semaphore
+throttling real I/O independent of scheduler readahead. v2 collapses that: the cross-bucket
+constants above are both scheduler and I/O ceiling. If we later find storage backends that need hard
+global throttling, it goes inside the `SeriesSource` implementation, not in the engine.
+
+### Operators
+
+Each operator is a `trait Operator` that pulls from children when they are ready for their next
+batch of work (pulling is a blocking operation). Some operators are breaking, which means that any
+child operators must complete fully before they complete. Most queries will not have any breaking
+operators.
+
+| **Operator**     | **Role**                                                                  | **Breaking** |
+|------------------|---------------------------------------------------------------------------|--------------|
+| `VectorSelector` | Leaf. Opens storage cursor; one sample per series per step via lookback.  | No           |
+| `MatrixSelector` | Leaf. Sliding lookback window per step for rollup functions.              | No           |
+| `Rollup`         | Unified range-function driver (rate, increase, `*_over_time`, ...).       | No           |
+| `InstantFn`      | Pointwise scalar functions (abs, ln, clamp, histogram_quantile, ...).     | No           |
+| `Binary`         | Vector/vector or vector/scalar binop. Pre-computed series matching.       | No           |
+| `Aggregate`      | sum/avg/min/max/count/stddev/topk/bottomk/quantile by labels.             | partial      |
+| `Subquery`       | Re-grids child onto inner step; feeds outer MatrixSelector semantics.     | Yes          |
+| `Rechunk`        | Transposes series-major ↔ step-major when ops need the other axis.        | Yes          |
+| `CountValues`    | Data-dependent schema. Drains child, emits with runtime-derived labelset. | Yes          |
+| `Concurrent`     | Producer/consumer decoupling with a bounded mpsc channel.                 | No           |
+| `Coalesce`       | Fan-in: merges parallel child streams that share schema.                  | No           |
+
+`Aggregate` with `topk`/`bottomk`/`quantile` buffers a whole step before emitting; `sum`/`avg`/etc.
+are streaming. Group maps and series matching are computed at plan time and reused for every batch
+since they are invariant across steps.
+
+### Caching & Per-Query State
+
+All caching in v2 is intra-query. Each query builds its own state during Plan, consults it during
+Execution, and drops it on completion. This keeps the memory story trivial (everything rolls up to
+the query's `MemoryReservation`) and sidesteps the cross-query contention problem.
+
+There are two categories of per-query state with different concurrency contracts:
+
+**Frozen State:** The`build_physical_plan`, handed to operators as `Arc<…>`, never mutated during
+execution. This has the series roster, which is the resolved `SeriesSchema` produced by draining
+`SeriesSource::resolve(..)` streams. Indexed by `series_idx: u32`; labels, fingerprint, and bucket
+membership baked in. Operators use the index for dense-array state (group maps, binary-match tables)
+compiled once from the roster at plan time.
+
+**Index Cache:** This is a concurrent cache that lives for the duration of the query. it
+deduplicates index lookups within the query because `resolve` fan-out (cross-bucket × per-key, up to
+1024 in-flight gets) would otherwise issue the same `(bucket, term)` or `(bucket, series_id)` fetch
+multiple times from parallel tasks. It caches inverted and forward index fetches.
+
+### Introspection: EXPLAIN and Trace
+
+Every query endpoint accepts two opt-in flags that surface the planner and executor internals without
+changing the result shape.
+
+**`?explain=true`** short-circuits evaluation and returns a three-stage dry run: unoptimised logical,
+optimised logical, and a pure description of the physical tree the planner would build. No operators
+are instantiated and no storage I/O is issued. Add `?pretty=true` to get a DataFusion-style indented
+text tree instead of the JSON `ExplainResult`.
+
+**`?trace=true`** runs the query normally and attaches a structured trace to the response. The
+physical planner inserts a transparent `TracingOperator` around each concrete operator; per-op poll
+time is recorded without the operator knowing. Storage I/O is bucketed separately (`SamplesFetch`,
+`Deserialize`, `ForwardIndexFetch`, inverted-index fetch) and attributed to the currently-polling
+operator via a thread-local node id. The same spans are emitted to `tracing` whenever `RUST_LOG`
+allows, so out-of-band log capture works without the flag.
+
+```
+GET /api/v1/query?query=sum(rate(http_requests_total[5m]))&explain=true&pretty=true
+GET /api/v1/query_range?query=...&start=...&end=...&step=15s&trace=true
+```
+
+## Testing Strategy
+
+Tests are layered to match the architecture: each operator is verified in isolation, the planner is
+pinned with snapshots, and full-pipeline behaviour is exercised by scenarios and HTTP e2e tests.
+
+| Layer            | Location                          | What it covers                                                         |
+|------------------|-----------------------------------|------------------------------------------------------------------------|
+| Operator unit    | `operators/*.rs` `#[cfg(test)]`   | One operator vs. mock child / source; state edges, validity, memory.   |
+| Planner golden   | `plan/explain.rs`                 | Snapshots logical → optimised → physical trees for canonical queries.  |
+| Stress scenario  | `testing/columnar_stress.rs`      | Multi-bucket synthetic load (~1k–18k series) vs. precomputed `Oracle`. |
+| HTTP end-to-end  | `tests/http_server.rs`            | Drives `/api/v1/query{,_range}` through the full stack.                |
+| Microbenchmarks  | `benches/v2_engine_micorbench.rs` | Criterion benches on a `WarmRangeQueryHarness`.                        |
+| Manual Testing   | `N/A`                             | Ran queries with ?explain=true and ?trace=true                         |
+
+**Known gap.** The `rate()` portion of the range-query stress scenario has a ~1.5× inflation at
+bucket boundaries, pointing at cross-bucket counter handling. Instant probes and the count/sum
+portions of the same scenario pass. Tracked for a dedicated fix pass rather than blocking this RFC.
+
+## Alternatives Considered
+
+1. Use `Arrow` for the on-the-wire protocol. This was tempting since it would implement some SIMD
+   vectorized computations for the aggregations but the dependency is significant and adds real
+   compile-time dependency scope for a relatively simple usage of it.
+2. Use `DataFusion` for the query planning. There are too many PromQL-specific concepts (lookback
+   deltas, offsets, step-aligned rollups) that don’t clearly map to Data Fusion, so the integration
+   cost and dependency weight make it less tempting.
--- a/timeseries/rfcs/0007-promql-execution.md
+++ b/timeseries/rfcs/0007-promql-execution.md
@@ -26,7 +26,7 @@ There are the following problems in the original implementation:
    range and re-runs the evaluator evaluation. This pipeline does not properly amortize repeated
    work like label hashing, metadata lookups and per-series iteration. We papered over some of this
    using intra-query caches, but this fix was imperfect.
-2. **Row-Oriented Compu**t**ation:** the data model stored samples keyed by label. This meant that
+2. **Row-Oriented Computation:** the data model stored samples keyed by label. This meant that
    each step evaluation would binary search for the closest matching sample (in the step’s
    neighborhood according to lookback window), and perform an indexed array lookup. This not only
    repeats work that is unnecessary for every step of evaluation it also is extremely inefficient
@@ -88,10 +88,17 @@ The EXPLAIN examples below walk a concrete query through the planning half of th
 
 ## Planning Stage
 
-Planning is split into two stages with a hard boundary between them: `Expr → LogicalPlan` is the *
-*rewritable** stage, `LogicalPlan → PhysicalPlan` is the compiling stage. Everything that wants to
-inspect, modify, or explain a query touches `LogicalPlan`; once we cross into physical, the tree is
-opaque and only the executor talks to it.
+Planning is split into three passes with a hard boundary between logical and physical:
+
+1. `Expr → LogicalPlan` — binding: lowers the parser AST into a typed logical tree, attaching
+   query-level context (step grid, lookback, storage handles).
+2. `LogicalPlan → LogicalPlan` — optimization: a rewritable pass for constant folding, CSE,
+   label pushdown, etc. The current optimizer is intentionally minimal.
+3. `LogicalPlan → PhysicalPlan` — compilation: resolves storage, binds the series roster, and
+   emits the operator tree the executor runs.
+
+Everything that wants to inspect, modify, or explain a query touches `LogicalPlan`; once we
+cross into physical, the tree is opaque and only the executor talks to it.
 
 We introduce a dedicated logical plan instead of reusing the AST module we depend on because it
 allows us to attach information such as the query’s timestamps and reinterpret concepts in more
@@ -102,6 +109,30 @@ The physical plan is a tree of `Operators` that are each individually testable a
 input/output types (defined below).
 
 ## Physical Plan Execution
+
+### Operator Trait
+
+Every node in the physical tree implements `Operator`. It is pull-based and object-safe so the
+planner can build `Box<dyn Operator>` trees without monomorphising over every combination.
+
+```rust
+pub trait Operator: Send {
+    fn schema(&self) -> &OperatorSchema;
+    fn next(&mut self, cx: &mut Context<'_>) -> Poll<Option<Result<StepBatch, QueryError>>>;
+}
+```
+
+- `schema()` is callable before any `next()` so downstream operators can size buffers and
+  precompute state up front.
+- `next()` returning `Ready(None)` is terminal — callers must not poll again.
+- Each operator owns its own iteration cursor, which is how e.g. `VectorSelectorOp` knows which
+  slice of its resolved series/step tile to emit on each call.
+
+The executor drives the query by polling the root operator in a loop until it yields `None`.
+
+Range-vector producers (`MatrixSelectorOp`, `SubqueryOp`) use a parallel trait, `WindowStream`,
+whose `poll_windows` returns a `MatrixWindowBatch` instead of a `StepBatch`. `RollupOp` is the
+only consumer. See [Range-Vector Batches](#range-vector-batches).
 
 ### Core Data Model
 
@@ -149,7 +180,8 @@ working set that stays in L2. Both step-wise ops (`sum by` across series at one 
 ops (`rate` across steps of one series) have cache-friendly traversal. Operators that need the other
 axis insert a `Rechunk` breaker.
 
-Here is the actual struct. Note that we us validity bitsets to distinguish between 0/null values.
+Here is the actual struct. Note that we use validity bitsets to distinguish between 0/null
+values.
 
 ```rust
 struct StepBatch {
@@ -170,7 +202,8 @@ struct StepBatch {
 
 The `SeriesSchema` is constant for the lifetime of a single query. It is built once and contains
 the full roster of series that the query touches, allowing operators to reference them by index
-instead of re-hashing labels:
+instead of re-hashing labels. Series are indexed globally by fingerprint across all buckets the
+query touches — identical labels in two buckets collapse to one entry.
 
 ```rust
 struct SeriesSchema {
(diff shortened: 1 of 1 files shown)


timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+compiled once from the roster at plan time.
+
+**Index Cache:** This is a concurrent cache that lives for the duration of the query. it
+deduplicates index lookups within the query because `resolve` fan-out (cross-bucket × per-key, up to
+1024 in-flight gets) would otherwise issue the same `(bucket, term)` or `(bucket, series_id)` fetch
+multiple times from parallel tasks. It caches inverted and forward index fetches.
+
+### Introspection: EXPLAIN and Trace
+
+Every query endpoint accepts two opt-in flags that surface the planner and executor internals without
```
rodesai, line 439: I think I'm mostly following the above (great work btw!), but having a little trouble concretizing how the plans would look. It would be helpful if you could include the output of `explain=true` for some queries (maybe a simple instant query, a range query, one that uses a range-vector selector, and one with subqueries)

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+There are two categories of per-query state with different concurrency contracts:
+
+**Frozen State:** The`build_physical_plan`, handed to operators as `Arc<…>`, never mutated during
+execution. This has the series roster, which is the resolved `SeriesSchema` produced by draining
+`SeriesSource::resolve(..)` streams. Indexed by `series_idx: u32`; labels, fingerprint, and bucket
+membership baked in. Operators use the index for dense-array state (group maps, binary-match tables)
+compiled once from the roster at plan time.
+
+**Index Cache:** This is a concurrent cache that lives for the duration of the query. it
+deduplicates index lookups within the query because `resolve` fan-out (cross-bucket × per-key, up to
```
apurvam: Very nice!

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
+# RFC 007: PromQL Execution Engine
+
+**Status**: Draft
+**Authors**: [Almog Gavra](https://github.com/agavra)
+
+## Summary
+
+This RFC proposes a rewrite of the core query execution pipeline for timeseries to a columnar,
+timestamp-major, pull-based operator pipeline with proper logical and physical planning steps.
```
apurvam: what's "timestamp-major" ? I think I know what 'pull-based is', though I think the old executor is also reasonably considered pull based right?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+   projection pushdown or range coalescing into SlateDB if/when we support that but it also makes it
+   harder in the future for us to add things like a fan-out, multi-node query engine.
+
+The new proposed pipeline brings `timeseries` closer to the SOTA implementation for query
+evaluation.
+
+## Goals
+
+- One execution model for instant and range queries
+- Columnar, step-major data flow from storage to result
```
apurvam, line 46: What is 'step-major'. Is it the same as 'timestamp-major' above?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+
+```
+HTTP / embedded API
+        │
+        ▼
+┌───────────────────┐     parse PromQL → parser::Expr
+│    Parse          │
+└──────┬────────────┘
+       ▼
+┌───────────────────┐     AST rewrites: constant fold, label pushdown,
```
apurvam, line 68: I wonder if this is a good place to add an example on the right hand side to show how a given query string looks through the pipeline. It may be too much, for one diagram, but I think that detail will be very helpful in constructing a clean mental model for a human.

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+queries.
+
+The physical plan is a tree of `Operators` that are each individually testable and have the same
+input/output types (defined below).
+
+## Physical Plan Execution
+
+### Core Data Model
+
+The universal on-the-wire shape between operators is a `StepBatch`: a contiguous range of output
```
apurvam: By 'on-the-wire' do you mean in memory? IIUC nothing about `StepBatch` is transmitted across the network. It's just an inmemory representation of query state during execution. I think this sentence is essentially: operators interface via step-batch, either loading data into a particular 'cell' (my term, to mean the timestamp x selector resolution to a 'sample list cell'), or reading from it to evaluate an expression.

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+steps for a contiguous chunk of series, with data stored in columnar representation. This data model
+gives us two major benefits:
+
+1. It is columnar with a step-major, series-minor format which allows us to quickly compute
```
apurvam, line 143: I think defining what 'major' and 'minor' mean up front will be good. I can't get a confident sense of the meaning of these words in the doc.

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+   aggregations within a single time-step.
+2. It is chunked, allowing us to pipeline and stripe execution across both dimensions (time and
```
apurvam, line 145: Intuitively, i think different parts of the 2d spaces are completely independent of each other, allowing contention free parallel processing. Is this what you mean?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+
+### Data Source / Storage API
+
+The interaction between the query execution pipeline and storage happens over the `SeriesSource`
+trait with two methods:
+
+- `resolve(selector, time_range)` to handle metadata lookups and resolve series info
+- `samples(SamplesRequest)` to retrieve the raw samples
+
+The storage is unaware of PromQL concepts such as lookback deltas, offsets, step alignment, etc… all
```
apurvam: So you mean the leaf operators encode this information and use it and make the correct calls to storage covering the data they need?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+of that is handled by the leaf operators (`VectorSelector` and `MatrixSelector`).
+
+Storage is accessed both during planning and execution. In the planning phase, `build_physical_plan`
+walks the logical tree and, for each selector, calls `resolve(..)` and drains the stream into the
```
apurvam: First introduction of the word 'stream'. What do you mean?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+plan-time series roster (`Arc<SeriesSchema>`). This is the *only* place series identity is resolved;
```
apurvam: Would be worth spelling out the `SeriesSchema` above, along side the `StepBatch` struct.

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+after this, operators index series by `u32` slot. `resolve` runs eagerly and synchronously with
+regards to physical-plan construction which means the planning time includes metadata latency.
```
apurvam: does this mean that metadata is loaded serially?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+
+There are a few areas where we can introduce parallelism and concurrency:
+
+**Inside the storage fetching.** Selectors typically span multiple storage buckets and since buckets
+are disjoint keyspaces, fan-out is safe). The implementation uses a buffered stream with max
+concurrency 32 for both metadata and sample streams. These constants mirror v1's
+`METADATA_STAGE_READAHEAD` / `SAMPLE_STAGE_READAHEAD` so observed concurrency is comparable.
+
+**Pipelining operator execution.** A selector leaf produces I/O-bound batches but downstream ops are
+CPU-bound. `ConcurrentOp` decouples them: the child runs on a spawned tokio task and pushes into a
```
apurvam, line 316: I think I know what you mean, but defining what a 'child' and 'consumer' is would be clear. Basically, a selector is loading IO using tokio tasks, and any higher level operator that needs the data will be block. With a channel size of 4, the 5th higher level operator that needs that particular datum will stall the pipeline.

is this what you mean?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+bounded mpsc channel (default bound = 4), giving implicit back-pressure since if the consumer
+stalls, the child blocks on `send`. The physical planner inserts one wrapper per selector leaf whose
+resolved series count is ≥ 64.
+
+**Vertical sharding.** We defer using the `Coalesce` operator in the initial implementation but the
```
apurvam, line 321: This may be a know rust concept, but I don't know what you mean by `Coalesce` here.

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+
+We introduce a dedicated logical plan instead of reusing the AST module we depend on because it
+allows us to attach information such as the query’s timestamps and reinterpret concepts in more
+structure form. The conversion is pretty minor, though it also enables us to power `EXPLAIN`
+queries.
+
+The physical plan is a tree of `Operators` that are each individually testable and have the same
+input/output types (defined below).
+
+## Physical Plan Execution
```
apurvam, line 111: Would be worth talking about how sub-queries are handled. Are they one opaque leaf operator for the rest of the query that depends on their outputs?

apurvam (commented):
Fantastic work! I think this architecture is a major improvement. I specifically like:

1. Only the leaf operators access storage
2. the separation of logical and physical plans (in theory, not sure how they will look in code)
3. The unified parallelism 
4. The unified caching architecture.
5. The data layout of time x selector => sample
6. the `?explain=` and `?trace=` parameters to enable printing query plans and tracing. 

These are all material improvements we should add. 

My main feedback is on the document itself at this point. I know this work leaned heavily on AI agents, and some of that is bleeding up into the doc. I've marked places where the terms there is assumed context. My bar for the RFC is a reader with minimal context should be able to understand a) the architecture, b) concurrency model,, c) code structure well enough to internalize it and get an solid intuition for how it will work. My questions inline are where the lack of context left me guessing and didn't give me confidence in my intuition. 

Once those are filled in, I can re-read it. IF the mental model is clear (to me) and still attractive (for reasons laid out above), I'd be +1 on this RFC.

timeseries/rfcs/0007-promql-execution.md
agavra: "Major" is the primary dimension of a matrix, so basically in a timestamp major data matrix you'll have each row be associated with a single timestamp (or more accurately a "step"). The old one was kind-of pull base, it was more a "load-then-execute" model. By "pull based" I mean that the parents pull data incrementally from the child nodes, and that mechanism applies backpressure (or rather forward pressure?)

timeseries/rfcs/0007-promql-execution.md
agavra: Yes, let me standardize on "step major" - a step is a timestamp, but aligned to the query not to the data

timeseries/rfcs/0007-promql-execution.md
agavra: Yes, "on the wire" just means in between steps, which for now is just in-memory (there are various "in memory" representations of the data, the "on the wire" intends to make it clear that it's the format used in the channels that are communicated between the operators). The "wire" implementation can be made pluggable for if/when we do multi-node query execution. 

Your understanding is correct though.

timeseries/rfcs/0007-promql-execution.md
agavra: yup exactly!

timeseries/rfcs/0007-promql-execution.md
agavra: Yup exactly, specifically the Vector/Matrix selectors will convert their requests into the appropriate time windows. This is how it works today, so it's not a departure. Just here for completeness.

timeseries/rfcs/0007-promql-execution.md
agavra: Literally the rust stdlib concept of a `Stream`, the API for `resolve` (for example) returns:
```rs
impl Stream<Item = Result<ResolvedSeriesChunk, QueryError>> + Send;
```
I dropped the detailed API here which I guess made this section less verbose but also a bit confusing. I'll add the full API back in.

timeseries/rfcs/0007-promql-execution.md
agavra: no, but it's blocking for the physical plan construction. Metadata loading works basically the same way it does today, so not pipelined but we kick off multiple concurrent requests. I believe we do pipeline inverted -> forward index lookups but I need to double check that.

timeseries/rfcs/0007-promql-execution.md
agavra: > With a channel size of 4, the 5th higher level operator that needs that particular datum will stall the pipeline.

Not quite. Consider this plan (output of EXPLAIN by the way):
```
AggregateOp [grouping={\"axis\":\"by\",\"labels\":[\"job\"]}, kind=Sum]
  ConcurrentOp [channelBound=4, concurrentGate=series_count >= 64]
    VectorSelectorOp [lookbackMs=300000, matcher=foo]
```

What this will do is basically create a channel between the VectorSelectorOp and the AggregateOp. The VectorSelectorOp will write into the channel and the AggregateOp pulls from the channel. The benefit of this is that the VectorSelectorOp can produce a batch into the channel and immediately start fetching the next batch before the aggregate op finishes processing that batch it produced.

Because the pipeline is "pull" driven, if we didn't have this the `VectorSelectorOp::next` otherwise wouldn't be called by the `AggregateOp` until it was finished.

timeseries/rfcs/0007-promql-execution.md
agavra: Not a rust concept, more a database query concept (see Spark's `Coalesce` for example: https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-hints.html#partitioning-hints). I admit I only heard about it here for the first time so it's probably worth making it explicit.

timeseries/rfcs/0007-promql-execution.md
agavra: Yeah, I admit I glossed over the subquery part and the current implementation is very naive. It's worth flushing it out in more details.

timeseries/rfcs/0007-promql-execution.md
agavra: I'm adding a section (as per Rohan's suggestion) that shows some concrete EXPLAIN plan outputs. Here's an example:

```
Aggregation with grouping — sum by (job)(foo):

=== logical (unoptimized) ===
Aggregate [grouping={"axis":"by","labels":["job"]}, kind=Sum]
  VectorSelector [lookbackMs=300000, matcher=foo]
=== logical (optimized) ===
Aggregate [grouping={"axis":"by","labels":["job"]}, kind=Sum]
  VectorSelector [lookbackMs=300000, matcher=foo]
=== physical ===
AggregateOp [grouping={"axis":"by","labels":["job"]}, kind=Sum]
  ConcurrentOp [channelBound=4, concurrentGate=series_count >= 64]
    VectorSelectorOp [lookbackMs=300000, matcher=foo]
```

Hopefully that helps tie it all together.

timeseries/rfcs/0007-promql-execution.md
agavra: Note it is also defined in the table above:
```
| `Coalesce`       | Fan-in: merges parallel child streams that share schema.                  | No           |
```

timeseries/rfcs/0007-promql-execution.md
agavra: Updated with definition at the top:

```
- **Step-major**: query intermediate state is a 2D grid indexed by (step, series), stored row by row
  so all series at one step are contiguous in memory. Aggregations and binary ops walk that row.
  Contrast this with the existing series-major representation, where the data is indexed by 
  (series, step).
```

timeseries/rfcs/0007-promql-execution.md
agavra: Added a section with some explain=true queries. Note that none of the examples actually leverage the logical plan optimizations (they're really stupid like deleting repeat selectors (`{job="a", job="a"} -> {job="a"}`)

timeseries/rfcs/0007-promql-execution.md
agavra: Remove the "on the wire" terminology in latest patch to clarify.

timeseries/rfcs/0007-promql-execution.md
agavra: I added the API back in for clarification.

timeseries/rfcs/0007-promql-execution.md
agavra: I added a dedicated section on Coalesce/Rechunk

timeseries/rfcs/0007-promql-execution.md
agavra: Added a dedicated section (pretty short and sweet since subqueries are relatively rare) about how subqueries work.

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,453 @@
...
+```
+
+The `SeriesSchema` is constant for the lifetime of a single query. It is built once and contains
+the full roster of series that the query touches, allowing operators to reference them by index
+instead of re-hashing labels:
+
+```rust
+struct SeriesSchema {
+    // Parallel arrays indexed by `series_idx: u32`. Length = total series in the query.
+    labels: Arc<[Labels]>,                // one Labels per series
```
apurvam, line 211: curious why we have the labels and the fingerpint. Isn't the fingeprint a hash of the labels?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,453 @@
...
+}
+```
+
+### Data Source / Storage API
+
+Storage talks to the engine through the `SeriesSource` trait:
+
+```rust
+pub trait SeriesSource: Send + Sync {
+    fn resolve(
```
apurvam, line 255: What's a `ResolvedSeriesChunk`? from the following text, it seems like it includes the `SeriesSchema` which means resolve extracts the labels and computes the fingerprint for each series. What else is included?

apurvam: does this mean that this is effectively a inverted index lookup step to fetch the series matching the selector + a forward index lookup to fetch all the labels and compute the fingerprint?

timeseries/rfcs/0007-promql-execution.md
apurvam: Makes sense. How does the `VectorSelectorOp` know what the next batch to fetch is if it has only one parent in the tree and it's a pull based model? Does `AggregateOp` also issue a batch of reads on `VectorSelectorOp`?

May be worth spelling this out little more precisely since it is both nuanced and critical.

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,453 @@
...
+global throttling, it goes inside the `SeriesSource` implementation, not in the engine.
+
+## Operators
+
+Each operator is a `trait Operator` that pulls from children when they are ready for their next
+batch of work (pulling is a blocking operation). Some operators are breaking, which means that any
+child operators must complete fully before they complete. Most queries will not have any breaking
+operators.
+
+| **Operator**     | **Role**                                                                  | **Breaking** |
```
apurvam, line 339: Worth including `SubQueryOp` here, and referencing the following section for details.

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,453 @@
...
+steps complete. 
+
+### Coalesce and Rechunk
+
+Both operators live in the `LogicalPlan` enum but are not yet inserted by the planner
+(`coalesce_max_shards` defaults to `0`; a `Rechunk` in a lowered plan errors today). They're
+reserved for future work:
+
+- **Vertical sharding via `Coalesce`.** A selector splits its resolved series list across
+  shards that run in parallel through the shared tree, and `Coalesce` merges at the first common
```
apurvam: what's a shard? a set of series or a set of time buckets?

apurvam (approved):
I left some more clarification questions. But overall this looks good to me. The approach and the structure and the decisions are all substantial improvements. I think we should make this document as clear as possible for human readers, but everything that needs clarification is relatively minor and we can add to it in a living fashion.

timeseries/rfcs/0007-promql-execution.md
agavra: It maintains its own state. You can think of it as an iterator across the grid of step x series, if that grid is 100x100, then you can think of a chunk as the first `data[0:25][0:25]` chunk. It knows it has returned that on the last call to `next()` so the next time it calls `next` it will return `data[0:25][26:50]` or whatever.

I think this is best explained by the Operator interface. I can just hoist that into the RFC but here it is directly for now:

```rs
/// The one trait every query-plan node implements. Consumers call
/// [`Self::next`] to pull the next [`StepBatch`], or [`Self::schema`] to
/// learn the output shape before polling.
///
/// Object-safe on purpose — the planner builds heterogeneous operator
/// trees as `Box<dyn Operator>` without monomorphising over every
/// combination. `Send` is required so
/// [`ConcurrentOp`](super::operators::concurrent::ConcurrentOp) can move a
/// subplan onto a spawned task; `Sync` is deliberately not required,
/// because operators hold per-poll mutable state.
pub trait Operator: Send {
    fn schema(&self) -> &OperatorSchema;

    /// - `Ready(Some(Ok))` — one batch.
    /// - `Ready(Some(Err))` / `Ready(None)` — terminal; must not poll again.
    /// - `Pending` — back-pressure or I/O wait; runtime wakes via `cx`.
    fn next(&mut self, cx: &mut Context<'_>) -> Poll<Option<Result<StepBatch, QueryError>>>;
}
```

timeseries/rfcs/0007-promql-execution.md
agavra: that's a good question, it's a reference to the global list of labels so it isn't expensive. looking at the code it looks like it's only ever really used for debug printing.

timeseries/rfcs/0007-promql-execution.md
agavra: > does this mean that this is effectively a inverted index lookup step to fetch the series matching the selector + a forward index lookup to fetch all the labels and compute the fingerprint?

yeah that's exactly right. it resolves buckets and then returns a stream of `ResolvedSeriesChunk` it gets from reading the inverted + forward indexes.

As far as what it "is", it's probably best just looked at in the code. It's basically just a stupid reference because we need some kind of struct type to pass along. It doesn't even compute the fingerprints at this point, that happens when we create the `SeriesSchema` which we do in the leaf node (or aggregate node if we're computing a new `SeriesSchema` from an aggregate group by) from the `ResolvedSeriesChunk`.

```rs
/// One contiguous slice of resolved-series metadata streamed from
/// [`SeriesSource::resolve`]. A selector typically spans multiple storage
/// buckets, and the source returns one chunk per bucket — the caller
/// concatenates them into the plan-time series roster.
///
/// Each chunk is self-describing: `labels` and `series` are parallel arrays
/// indexed together, and every handle in `series` points into this chunk's
/// `bucket_id`.
#[derive(Debug, Clone)]
pub struct ResolvedSeriesChunk {
    pub bucket_id: u64,
    pub labels: Arc<[Labels]>,
    pub series: Arc<[ResolvedSeriesRef]>,
}

/// Opaque handle minted by [`SeriesSource::resolve`] and threaded back into
/// [`SamplesRequest::series`]. Callers treat the contents as opaque.
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
pub struct ResolvedSeriesRef {
    pub bucket_id: u64,
    /// Bucket-scoped series identifier (matches [`crate::model::SeriesId`]).
    pub series_id: u32,
}
```

timeseries/rfcs/0007-promql-execution.md
agavra: it is there already:
```
| `Subquery`       | Re-grids child onto inner step; feeds outer MatrixSelector semantics.     | Yes          |
```
I can add a reference, though AFAICT there's no "link" capability in basic markdown.

timeseries/rfcs/0007-promql-execution.md
agavra: tbd, I haven't implemented any sharding mechanism but you could shard on any dimension you want (series or steps/time buckets)

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+**Authors**: [Almog Gavra](https://github.com/agavra)
+
+## Summary
+
+This RFC proposes a rewrite of the core query execution pipeline for timeseries to a columnar,
+timestamp-major, pull-based operator pipeline with proper logical and physical planning steps.
+
+## Motivation
+
+There are the following problems in the original implementation:
```
rodesai, line 23: just curious - do you have a sense of the magnitude of the various improvements? I think the biggest improvements are:
- resolving series up front in a single pass to avoid repeated work per step eval
- pipelining compute with i/o to reduce latency
- computing the results in a streaming form to (1) avoid searching the samples for the current step on each eval and (2) get better cache behavior (less pointer chasing, plays better with the cpu's readahead, etc)

Is that a fair assessment?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+## Architecture
+
+```
+HTTP / embedded API
+        │
+        ▼
+┌───────────────────┐     parse PromQL → parser::Expr
+│    Parse          │
+└──────┬────────────┘
+       ▼
```
rodesai, line 67: does the implementation have separate rewrite and logical plan optimization phases?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+         ├──────────┼──────────┼──────────┤
+    t₃   │ 12.9  ✓  │ 44.0  ✓  │  7.4  ✓  │
+         ├──────────┼──────────┼──────────┤
+    t₄   │ 13.1  ✓  │ 45.2  ✓  │  —    ✗  │
+         ├──────────┼──────────┼──────────┤
+    t₅   │ 13.4  ✓  │ 46.0  ✓  │  7.8  ✓  │
+         └──────────┴──────────┴──────────┘
+```
+
+Each batch is a rectangle of `K series × N steps`. Default `N ≈ 64`, `K ≈ 512` giving a ~256 KB
```
rodesai, line 178: what is stored in each "cell" of the matrix - is it a single data point? Or does it depend on whether the data represents an instant vector or range vector? I'm not really grokking how we store range vectors in this model.

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+child operators must complete fully before they complete. Most queries will not have any breaking
+operators.
+
+| **Operator**     | **Role**                                                                  | **Breaking** |
+|------------------|---------------------------------------------------------------------------|--------------|
+| `VectorSelector` | Leaf. Opens storage cursor; one sample per series per step via lookback.  | No           |
+| `MatrixSelector` | Leaf. Sliding lookback window per step for rollup functions.              | No           |
+| `Rollup`         | Unified range-function driver (rate, increase, `*_over_time`, ...).       | No           |
+| `InstantFn`      | Pointwise scalar functions (abs, ln, clamp, histogram_quantile, ...).     | No           |
+| `Binary`         | Vector/vector or vector/scalar binop. Pre-computed series matching.       | No           |
```
rodesai, line 345: how does series matching work? I'm guessing it happens during the planning phase, so the plan encodes the series that match with each other. But then we need to ensure that all grouped series wind up in the same `StepBatch` right? Is that also handled by the plan?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+1. It is columnar with a step-major, series-minor format which allows us to quickly compute
+   aggregations within a single time-step.
+2. It is chunked, allowing us to pipeline and stripe execution across both dimensions (time and
+   labels).
+
+```
+StepBatch  (step_range = 2..6, series_range = 1..4)
+──────────────────────────────────────────────────────────────────
+
+step_timestamps: Arc<[i64]>      series: Arc<SeriesSchema>
```
rodesai, line 152: Are the series inside `SeriesSchema` global or local to a bucket? E.g. if I have the same labels in 2 different buckets - do those show up as the same entry in `SeriesSchema` or as separate entries?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+The execution `VectorSelector` and `MatrixSelector` operators call `samples(..)` with the
+pre-resolved series list and the absolute time window they need. Batches come back streamed and the
+leaf reshapes them into `StepBatch`es that the rest of the tree polls. No other operator
+communicates with the storage layer.
+
+### Concurrency & Parallelism
+
+There are a few areas where we can introduce parallelism and concurrency:
+
+**Inside the storage fetching.** Selectors typically span multiple storage buckets and since buckets
```
rodesai: Not quite following this. Isn't it safe to fan-out within a single bucket as well (e.g. concurrently fetch multiple series)?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+are disjoint keyspaces, fan-out is safe). The implementation uses a buffered stream with max
+concurrency 32 for both metadata and sample streams. These constants mirror v1's
+`METADATA_STAGE_READAHEAD` / `SAMPLE_STAGE_READAHEAD` so observed concurrency is comparable.
+
+**Pipelining operator execution.** A selector leaf produces I/O-bound batches but downstream ops are
```
rodesai, line 315: what drives consumption of the final query results? I guess the server polls some output operator?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,296 @@
...
+resolved series count is ≥ 64.
+
+**Vertical sharding.** We defer using the `Coalesce` operator in the initial implementation but the
+wiring is stubbed in `parallelism` and `coalesce_max_shards` defaults to `0`. Needs end-to-end
+correctness work (per-series independence above the leaf isn't free) before it's turned on.
+
+**No global permit layer.** v1 had a separate `QueryReaderEvalCache` metadata/sample semaphore
+throttling real I/O independent of scheduler readahead. v2 collapses that: the cross-bucket
+constants above are both scheduler and I/O ceiling. If we later find storage backends that need hard
+global throttling, it goes inside the `SeriesSource` implementation, not in the engine.
```
rodesai: We could also do this at the storage layer itself initially. `ObjectStore` supports this out-of-the-box using the `LimitStore` wrapper

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,453 @@
...
+streaming lets the engine fan out across buckets and pipeline per-bucket results without waiting
+for the slowest one.
+
+The trait is deliberately PromQL-unaware: lookback deltas, offsets, step alignment, and rollup
+windows live in the leaf operators (`VectorSelector`, `MatrixSelector`), not in storage.
+
+Storage is accessed both during planning and execution. In the planning phase,
+`build_physical_plan` walks the logical tree and, for each selector, calls `resolve(..)` and
+drains the resulting stream into the plan-time series roster (`Arc<SeriesSchema>`). This is the
+*only* place series identity is resolved; after this, operators index series by `u32` slot.
```
rodesai, line 295: I think there's a routing problem and I'm not quite following how its addressed. There are some expressions that operate on multiple series, so those series need to be in the same batch for that expression to be evaluated. It would be helpful to explain how that happens. Is that one place where we might have a `Coalesce` operator?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,453 @@
...
+
+There are two categories of per-query state with different concurrency contracts:
+
+**Frozen State:** The`build_physical_plan`, handed to operators as `Arc<…>`, never mutated during
+execution. This has the series roster, which is the resolved `SeriesSchema` produced by draining
+`SeriesSource::resolve(..)` streams. Indexed by `series_idx: u32`; labels, fingerprint, and bucket
+membership baked in. Operators use the index for dense-array state (group maps, binary-match tables)
+compiled once from the roster at plan time.
+
+**Index Cache:** This is a concurrent cache that lives for the duration of the query. it
```
rodesai: this is only used during planning right? When executing there's no reason to consult the index?

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,453 @@
...
+    Rollup [kind=Rate]
+      MatrixSelector [matcher=foo, rangeMs=60000]
+=== logical (optimized) ===
+Rollup [kind=AvgOverTime]
+  Subquery [rangeMs=300000, stepMs=60000]
+    Rollup [kind=Rate]
+      MatrixSelector [matcher=foo, rangeMs=60000]
+=== physical ===
+RollupOp [kind=AvgOverTime]
+  SubqueryOp [rangeMs=300000, stepMs=60000]
```
rodesai, line 524: this will currently execute each step serially right?

timeseries/rfcs/0007-promql-execution.md
rodesai: What I took it to mean - the data stored in a single batch is a 2-d vector of steps and series laid out in a flat array in memory. We could either choose to order the data by step first (so all the series at a single time are contiguous), or series first (so all the points for a single series are contiguous). What we use should be optimal for the operation we're doing (e.g. to sum all points at a given step its better to lay it out step-major).

rodesai (commented):
Took a full pass through and left some questions. Overall I'm totally on-board with the direction - just need some help understanding how a couple things fit together. It might help if you could throw the code up somewhere.

timeseries/rfcs/0007-promql-execution.md
agavra: Resolving the series up front in a single pass was by and large the biggest win (2-3x). The second biggest was actually moving it to columnar so we don't have to binary search on each step (another 20-30% improvement after optimization).

At this point, before I added pipelining, gorilla encode/decode was the biggest contributor at ~45% of query latency. I don't remember exactly where the rest of it went, it's probably worth getting a good trace of that back in the system.

I honestly don't have a good sense of how much pipelining compute with i/o and the cache-friendly behavior is. I can probably devise some tests to figure that out (run it with no pipelining at least should be easy) but I figured those are probably good design.

timeseries/rfcs/0007-promql-execution.md
agavra: not sure what you mean by "rewrite", the current implementation parses AST into a Logical Plan and then has a separate optimization pass before it creates the Physical Plan. The optimization path is pretty dumb at the moment.

timeseries/rfcs/0007-promql-execution.md
agavra: I'm glad you called this out because I admit I glossed over the implementation for matrix selectors which return ranges. Right now the `MatrixSelectorOp` is always paired with a `RollupOp` which bypass the normal `next() -> StepBatch` data model. 

I think this is a reasonable approach (since you can't return matrix values directly anyway in promql, it has to be coupled with a Rollup) but LMK if you think of anything better.

Let me update the RFC with more details on this mechanism, but here's the exact data type used for communication between Rollup and Matrix operators:

```rs
/// The window-batch counterpart to [`StepBatch`]: a rectangle of *sample
/// lists* (one list of `(ts, val)` pairs per `(step, series)` cell)
/// instead of single values. This is what flows between range-vector
/// producers ([`MatrixSelectorOp`], [`SubqueryOp`]) and
/// [`RollupOp`](super::rollup::RollupOp).
///
/// Covers a contiguous `(series_range × step_range)` tile. Samples for
/// each cell are packed into the flat `timestamps` / `values` columns and
/// indexed per cell via [`Self::cells`] (row-major by step, matching
/// [`StepBatch`]'s layout).
///
/// Consumers read per-cell slices via [`Self::cell_samples`]. `STALE_NAN`
/// values are already filtered out by the producer — consumers see only
/// valid numeric samples in ascending timestamp order.
///
/// [`SubqueryOp`]: super::subquery::SubqueryOp
#[derive(Debug, Clone)]
pub struct MatrixWindowBatch {
    /// Absolute step timestamps (ms), shared with the rest of the query
    /// via `Arc` just like [`StepBatch::step_timestamps`].
    pub step_timestamps: Arc<[i64]>,
    /// Slice of [`Self::step_timestamps`] covered by this batch.
    pub step_range: std::ops::Range<usize>,

    /// Series roster the planner built.
    pub series: SchemaRef,
    /// Slice of the series roster covered by this batch.
    pub series_range: std::ops::Range<usize>,

    /// Flat per-cell sample timestamps (ms), packed in row-major step ×
    /// series order. A cell's slice lives at [`CellIndex::range`].
    pub timestamps: Vec<i64>,
    /// Flat per-cell sample values, parallel to [`Self::timestamps`].
    pub values: Vec<f64>,
    /// Per-cell index. Length is `step_count * series_count`, row-major
    /// by step (cell `(t_off, s_off)` lives at `t_off * series_count +
    /// s_off`).
    pub cells: Vec<CellIndex>,

    /// Optional per-step **effective** timestamps — the window-end each
    /// step's samples actually live under after folding `@` / `offset`.
    /// `None` means the effective time equals the step timestamp (the
    /// common no-modifier case). When present, shares layout with
    /// [`Self::step_timestamps`] (absolute, indexed by global step idx)
    /// so consumers slice via `step_range` the same way.
    ///
    /// [`RollupOp`] prefers this over `step_timestamps` when computing
    /// `(window_start, window_end)` for rate-family extrapolation; without
    /// it, `rate(metric[100s] @ 100)` at outer step `t=25s` would compute
    /// rate over the window `(-75s, 25s]` while the packed samples
    /// actually cover `(0, 100s]`, producing a negative rate disjoint
    /// from the data.
    ///
    /// [`RollupOp`]: super::rollup::RollupOp
    pub effective_times: Option<Arc<[i64]>>,
}
```

And a pretty useful visual from claude:


## StepBatch — one float per cell

```
  series →   s0    s1    s2
  step 0  [ 1.0 | 2.0 | 3.0 ]    values:   Vec<f64>  (row-major: step * S + series)
  step 1  [ 4.0 | 5.0 | 6.0 ]    validity: BitSet    (same shape, 1 bit per cell)
  step 2  [ 7.0 | NaN | 9.0 ]    source_timestamps: Arc<[i64]>  (per-cell sample ts)
```

Fixed size: `step_count × series_count` floats. Absence = validity bit clear.

## MatrixWindowBatch — a *list* of samples per cell

```
    ── cells grid (fixed: 2 steps × 3 series = 6 entries) ─────────────────

    series →    s0            s1            s2
    step 0   {off=0, len=1}  {off=1, len=0}  {off=1, len=2}
    step 1   {off=3, len=1}  {off=4, len=1}  {off=5, len=0}

    ── flat sample pool (length = sum of all cell lens = 5) ──────────────

              cell(0,s0)  cell(0,s2)   cell(1,s0)  cell(1,s1)
               ┌──┐       ┌───┬───┐    ┌──┐        ┌──┐
    timestamps:[ 5 ,       7 , 9 ,      15 ,        18 ]
    values:    [ 5 ,       7 , 9 ,      15 ,        18 ]
     indices:    0         1   2        3           4
```

**Key shape difference:** `StepBatch` is a dense rectangle of floats; `MatrixWindowBatch` is a dense rectangle of *slice descriptors* pointing into a ragged flat sample pool.

timeseries/rfcs/0007-promql-execution.md
agavra: Yes, series matching happens when we build the physical plan. And yes, there's a concept called the `GroupMap` which the planner also builds to understand the grouping requirements. 

Though your question about ensuring grouped series wind up together is another good one. I looked back at the code and it looks like aggregate operators just don't emit until it's collected all the data from the children so that it can ensure it has all the data it needs.  

I think the RFC is wrong here to call in "non-breaking". That's a bit of a bummer, I just trusted Claude on this one 😞 but I think it's fine to just mark it breaking for now. 

I think an improvement would be to hash partition the groups and then have children that are responsible for group scopes instead of the fully series scopes, allowing us to identify when we've filled a full group and emit the result at that point.

timeseries/rfcs/0007-promql-execution.md
agavra: Global, they're indexed by `[Fingerprint]`.

timeseries/rfcs/0007-promql-execution.md
agavra: yes that's correct, sorry this is some AI slop. It had originally thought fan-out wasn't safe and I told it it was, it made a modification and added this back-reference to something that doesn't make sense anymore. I'll clean it up.

timeseries/rfcs/0007-promql-execution.md
agavra: Yup, the physical plan just has a root operator that it polls:
```rs
        loop {
            match poll_fn(|cx| plan.root.next(cx)).await {
                Some(Ok(batch)) => batches.push(batch),
                Some(Err(e)) => return Err(execution_error_to_query_error(e)),
                None => break,
            }
        }
```

timeseries/rfcs/0007-promql-execution.md
agavra: I actually like that a lot better than the semaphores we had because it gets straight to the point of what we wanted to do anyway (limit concurrent calls to object storage).

timeseries/rfcs/0007-promql-execution.md
agavra: I think the comment on `AggregateOp` answers that for you, let me know if that's right.

timeseries/rfcs/0007-promql-execution.md
agavra: yeah that's a good point, we can then drop the cache earlier to free up some memory 👍 the code actually used it during execution to fetch the metric name (because to fetch the samples you need the metric name since we now have the slate key include that) but I think we should jsut resolve that upfront and drop the cache.

rodesai: Sounds good. Not a big deal right now, but I think it would be nice to have distinct boundaries between what's accessed at planning vs execution time to make our lives easier if/when we want to allow distributed execution.

agavra: Makes sense to me, I tried to quickly implement it and it seems possible. Subqueries add a bit of complexity since the current implementation replans inside execution (yuck) and it acknowledges this is a hack (haha):
```
        // Use a blocking task-local poll for the factory's sync-Future
        // surface: the factory's return type is sync, but the planner
        // walk is async. Real wiring will replace this with a planner-
        // cached precomputed tree per unique (range, step); for v1 the
        // factory just re-invokes the planner synchronously.
```

timeseries/rfcs/0007-promql-execution.md
agavra: Yes, a subquery fully executes one step at a time at the moment. I think a partition + coalesce would be the parallelism strategy going forward.

agavra:
@rodesai the code is on https://github.com/opendata-oss/opendata/tree/columnar_execution

timeseries/rfcs/0007-promql-execution.md
rodesai: This makes sense to me. The only limitation I see currently is that you can't do a query that directly returns a range vector. I think we'd have to generalize Operator somehow to allow it to return `StepBatch` or `MatrixWindowBatch` and allow that for the root expression, and then have some abstract way of reshaping that to `QueryValue`.

timeseries/rfcs/0007-promql-execution.md
```diff
@@ -0,0 +1,453 @@
...
+```rust
+struct StepBatch {
+    // Time axis — shared by all series in the batch. One allocation, reused
+    // across many batches within a query.
+    step_timestamps: Arc<[i64]>,          // length = step_count
+    step_range: Range<usize>,             // slice into step_timestamps
+
+    // Series axis — schema is fixed at plan time.
+    series: Arc<SeriesSchema>,            // labels, fingerprint, stable index
+    series_range: Range<usize>,           // slice into series schema
```
rodesai, line 195: I'm finding it hard to understand how this can be a range rather than a set. How can the series be ordered in a way that the planner can slice up batches vertically in a useful way? E.g. for one are all the series ordered by bucket? And then even then it might be important for certain series to land in the same batch based on what series should feed into a given computation.

timeseries/rfcs/0007-promql-execution.md
rodesai: Yep they were the same question more or less

timeseries/rfcs/0007-promql-execution.md
agavra: Discussed offline. Instead of partitioning within the full series range the alternative is to partition the sources themslves. This means the physical plan will be built with the subset grouping schema at the vector selectors. e.g. if you `sum by (container)` then you could (at most) partition the plan so that all series with the same container name end up on a dedicated partition. With high cardinality groupings we probably want to make this less aggressive and partition with some limits on cardinality.

rodesai (approved):

PR #422, opened 2026-04-23
Jemalloc build batch

agavra:
CPU profiling showed ~29% of time in glibc malloc/free and 11.87% self time in `VectorSelectorOp::build_batch`. Two changes address both:

- Install `tikv-jemallocator` as the global allocator in the timeseries binary (gated to non-msvc targets, wired to the `http-server` feature).
- Rewrite the per-cell lookback scan as a forward two-pointer per series. `effective_times` is non-decreasing across steps (monotonic grid, or constant under `@`), and each series' timestamps are ascending, so a cursor advancing past samples with `ts <= window_hi` is correct across all steps. Complexity drops from O(step_count × series × ts.len()) to O(step_count × series + Σ ts.len()). STALE_NAN semantics preserved.

### Results

Note that this also includes the slatedb `main` upgrades to foyer cache, DBstats cloning and a few other changes.

loadtest_compare

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2715,6 +2715,7 @@ dependencies = [
  "snap",
  "tempfile",
  "thiserror 2.0.17",
+ "tikv-jemallocator",
  "tokio",
  "tokio-util",
  "tower",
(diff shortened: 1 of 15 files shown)


apurvam (approved):
LGTM!

Issue #338, opened 2026-03-25
[timeseries] Significant read amplification when loading data from S3


apurvam:
I'm seeing what looks like very large cold-read amplification from timeseries. Basically, the bytes loaded as reported by SlateDB `get` and `scan` operations inside timeseries are 1/100th of the bytes reported as downloaded from S3. In other words, we are downloading 100x the data we need from S3 while evaluating queries. 

For context, I've been benchmarking and optimizing timeseries in the last couple of weeks. Initially I focused on optimizing warm queries by removing repeated work within the evaluator, and then switched to optimizing cold queries. For cold query optimizations, the two major improvements were from adding parallelism to the IOs and by adding `metric name` as a prefix to the series id in the `Timeseries` data structure. This groups series by metric names and improved IO locality. My benchmarking added _a lot_ of instrumentation as well, that informed various experiments. [perf/query-io-stats](https://github.com/opendata-oss/opendata/tree/perf/query-io-stats) has the latest code.

With that said, here are more detailed numbers highlighting the read amplification.

## Context

Environment:

  - Branch: https://github.com/opendata-oss/opendata/tree/perf/query-io-stats
  - node type: r5d.xlarge
  - bucket region: us-west-2

  Dataset / ingest:

  - index age: about 3 days
  - object-store footprint: about 71 GB in S3
  - ingest rate: about 8k samples/sec

  Workload shape:

  - Prometheus-style range queries
  - mostly aggregations over a single metric with few or no label filters
  - benchmark shape:
      - 18 isolated range queries
      - 6 hour range
      - 15s step
      - cold process / cold in-memory caches

  One heavy query dominates bytes and latency, but the same pattern shows up across the suite.

  ## Instrumentation

  I added per-query instrumentation in the Storage layer of opendata to measure bytes returned by SlateDB.

  Around our storage abstraction’s get() and scan() paths, we sum: `record.key.len() + record.value.len()`. 

  and attribute those bytes by logical read kind:

* bucket list
* inverted index
* forward index
* sample point gets
* sample range scans

I also emit a per-query total `io_physical_bytes_total`
 
  Separately, I pulled CloudWatch BytesDownloaded for the S3 bucket over the same benchmark window.

  ## Application-side bytes

  Across the 18 isolated queries, the summed storage-layer bytes received were 17,002,362 bytes, or about 17.0 MB.

[...]

PR #427, opened 2026-04-27
[timeseries] Stamp flushed records with bucket-aligned Ttl::ExpireAt

### LINKED ISSUE #342: [timeseries] set record level TTL
## Summary

We need to set record level TTL using SlateDB's new (not yet released) [ExpireAt](https://github.com/slatedb/slatedb/pull/1450) option.

We need to keep the expire time for all samples of a particular series in a particular bucket constant, because otherwise samples for a series in a bucket can't be merged because every one of them will have a separate expiry time by default (because today, expiry time is based on the time of the write).

This causes massive read amplification and high cold latencies, as documented here: https://github.com/opendata-oss/opendata/issues/338#issuecomment-4145276222

Filing the ticket to kick off the design discussion. Once we agree on an approach we can update the body with the details.

agavra:
Every record produced by a flush now carries the same `Ttl::ExpireAt(bucket_start_ms + retention_ms)`, so SlateDB compaction can collapse merge operands for a given (series, bucket). Previously the default per-write `ExpireAfter` gave each operand a slightly different expiration, blocking the merge and driving read amplification on hot keys (see #338).

Fixes #342

### DIFF
--- a/common/src/storage/in_memory.rs
+++ b/common/src/storage/in_memory.rs
@@ -48,12 +48,12 @@ impl StoredValue {
 
 /// Computes the absolute expiration timestamp from TTL options.
 fn compute_expire_ts(now: i64, ttl: Ttl, default_ttl: Option<u64>) -> Option<i64> {
-    let duration = match ttl {
-        Ttl::Default => default_ttl,
+    match ttl {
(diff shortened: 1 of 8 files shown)


apurvam (approved):
LGTM, thanks!

PR #433, opened 2026-05-05
[vector] Add delete vector functionality

cadonna:
## Summary

Adds delete vector functionality to vectorDB.

## Test Plan

- Unit tests in each component
- Integration tests in db.rs

## Checklist

- [x] Tests added/updated

### DIFF
--- a/vector/src/storage/record.rs
+++ b/vector/src/storage/record.rs
@@ -30,7 +30,6 @@ pub fn put_id_dictionary(external_id: &str, internal_id: VectorId) -> RecordOp {
 }
 
 /// Create a RecordOp to delete an IdDictionary mapping.
-#[allow(dead_code)]
 pub fn delete_id_dictionary(external_id: &str) -> RecordOp {
     let key = IdDictionaryKey::new(external_id).encode();
     RecordOp::Delete(key)
--- a/vector/src/write/delta.rs
+++ b/vector/src/write/delta.rs
@@ -214,4 +214,86 @@ mod tests {
             .unwrap();
         assert_eq!(delta.estimate_size(), 3);
     }
+
+    #[test]
+    fn delta_should_record_deletes_in_frozen_view() {
+        let mut delta = VectorDbWriteDelta::init(());
+
+        delta
+            .apply(VectorDbWrite::Delete(vec![
+                "v1".to_string(),
+                "v2".to_string(),
+            ]))
+            .unwrap();
+        delta
+            .apply(VectorDbWrite::Delete(vec!["v3".to_string()]))
+            .unwrap();
+
+        let (_frozen, frozen_view, _ctx) = delta.freeze();
+        assert_eq!(frozen_view.writes.len(), 0);
+        assert_eq!(frozen_view.deletes, vec!["v1", "v2", "v3"]);
+    }
+
+    #[test]
+    fn delta_should_record_writes_and_deletes_mixed() {
+        let mut delta = VectorDbWriteDelta::init(());
+
+        // Interleave writes and deletes; the frozen view should preserve each
+        // collection's insertion order, regardless of how the calls are mixed.
+        delta
+            .apply(VectorDbWrite::Write(vec![
+                make_write("v1", vec![1.0, 0.0, 0.0]),
+                make_write("v2", vec![0.0, 1.0, 0.0]),
+            ]))
+            .unwrap();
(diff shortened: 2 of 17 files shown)


vector/src/storage/record.rs
```diff
@@ -30,7 +30,6 @@ pub fn put_id_dictionary(external_id: &str, internal_id: VectorId) -> RecordOp {
 }
 
 /// Create a RecordOp to delete an IdDictionary mapping.
-#[allow(dead_code)]
```
apurvam, line 33: curious why this annotation is dropped? IF the code is dead, we should delete it right?

vector/src/write/delta.rs
```diff
@@ -128,67 +124,179 @@ mod tests {
         }
     }
 
+    /// Flatten the view's per-call ops into a list of every individual
+    /// `VectorWrite`, preserving order. Test-only helper.
+    fn flatten_writes(view: &VectorDbDeltaView) -> Vec<&VectorWrite> {
```
apurvam, line 129: Curious why we need this with the deletion patch?

apurvam (approved):
This LGTM. I didn't go deep into it since I don't have context on Vector as much. So it may be worth @rodesai 's review as well. But on the surface it looks reasonable. 

Also: is there an HTTP API for vector already? Does it need to be updated as well to include deletes? I didn't see that in this patch.

vector/src/write/indexer/tree/mod.rs
```diff
@@ -135,35 +136,55 @@ impl Indexer {
...
-            parent: &update_span,
-            op = "write_vectors",
-            elapsed_ms = elapsed_ms(write_start),
-            inserts,
-            updates,
-            "completed"
-        );
-        stats.inserts = inserts;
-        stats.updates = updates;
+        for op in ops {
```
rodesai: We should switch back to applying writes/deletes in a batch within `WriteVectors`. Indexing throughput depends on being able to batch expensive operations like computing centroid assignment and loading indexing metadata for updates/deletes. Doing them serially here serializes all those operations. We could either apply all writes in 1 pass and then all deletes, or just extend `WriteVectors` to apply deletes as well as writes. I think the latter would be simpler and potentially improve batching.

vector/src/storage/record.rs
rodesai: It was dead before but is used with this change.

vector/src/write/indexer/tree/state.rs
```diff
@@ -162,9 +165,18 @@ impl ForwardIndexDelta {
         self.vector_index_updates.remove(&vector_id);
     }
 
+    pub(crate) fn delete_external_id(&mut self, external_id: String, vector_id: VectorId) {
```
rodesai: can you clear dictionary_deletes when adding a vector to the dictionary? We should never actually do a delete and a write in the current impl, but better to ensure this type keeps things consistent.

rodesai (commented):
Thanks @cadonna! Left a couple things inline - the bit about preserving batching during index maintenance is significant. We should update the HTTP API as well, as @apurvam said.

vector/src/write/indexer/tree/state.rs
cadonna: I am not completely sure what you mean.

On line 144, in `add_vector()`, there is:
```rust
self.dictionary_deletes.remove(external_id);
```

Is this not enough?

vector/src/write/delta.rs
cadonna: That is just a helper for tests.

Before the deletion patch, the delta only contained writes. Now it contains writes and deletes interleaved. Some tests, verify writes and deletes separately. This helper is used to separate deletes and writes for that purpose.

vector/src/write/indexer/tree/mod.rs
cadonna: Done in [6aa3486](https://github.com/opendata-oss/opendata/pull/433/commits/6aa3486fa307bf26d783b61b97504793f63eb03a)

Please check if everything is as you envisioned.

cadonna:
Added the HTTP endpoint in [633bd16](https://github.com/opendata-oss/opendata/pull/433/commits/633bd16191b05d246d764d6d7ca2a2c8553d5344)

rodesai (approved):
LGTM!

PR #435, opened 2026-05-06
vector: add optional buffer consumer to ingest from buffer queue

cadonna:
## Summary

Adds a `buffer` cargo feature that wires an `opendata-buffer` Consumer into the vector server. When configured, the server runs a background task that polls the queue, decodes each entry as a protobuf `WriteRequest` (with a 2-byte version/encoding prefix), and writes through `VectorDb`. The consumer starts before the HTTP listener and is shut down before the server exits.

## Test Plan

- unit tests
- integration test

## Checklist

- [x] Tests added/updated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2744,6 +2744,7 @@ dependencies = [
  "figment",
  "futures",
  "log",
+ "opendata-buffer",
  "opendata-common",
  "opendata-macros",
  "prost",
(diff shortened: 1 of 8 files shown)


apurvam (approved):
Simple and straightforward. LGTM, thanks!

rodesai (approved):
LGTM!

PR #437, opened 2026-05-07
buffer-inspect: read-only inspector for ODB manifests

apurvam:
## Summary

Adds two pieces, originally developed during the ClickHouse-ingestor work but generic to any Buffer signal:

- **`buffer::reader::BufferReader`** — a typed read-only entry point on the `buffer` crate that fetches a manifest object, parses it via the existing `parse_manifest`, and walks referenced batches via `model::decode_batch`. It deliberately does **not** call `Consumer::new` (which bumps the manifest epoch and fences any active consumer), never writes the manifest, and never starts the garbage collector. Tests pin the manifest object's ETag and epoch across multiple read passes.
- **`buffer-inspect` workspace crate** (binary + library) — CLI that uses `BufferReader` to enumerate recent batches, decode per-entry metadata envelopes, and summarize OTLP logs/metrics payloads. Unknown signals/encodings print metadata bytes verbatim instead of crashing. Validated against the production logs and metrics manifests in the alpha; the inspector is signal-agnostic at the envelope layer and dispatched on each entry.

The ClickHouse-ingestor crate (which originally lived alongside this code on `ch-integration/odb-clickhouse-ingestor`) has been extracted into the new [`opendata-oss/opendata-contrib`](https://github.com/opendata-oss/opendata-contrib) repository and released as `clickhouse-ingestor v0.1.1`. This PR brings the generic Buffer tooling — the only artifact from that branch that belongs in the core repo — onto `main`.

## Test plan

- [x] `cargo test -p opendata-buffer -p buffer-inspect --lib` (96 passing locally on this branch)
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p opendata-buffer -p buffer-inspect --all-targets -- -D warnings`
- [ ] CI green

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -422,6 +422,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "buffer-inspect"
+version = "0.2.0"
+dependencies = [
+ "anyhow",
+ "base64 0.22.1",
+ "bytes",
+ "clap",
+ "opendata-buffer",
+ "opendata-common",
+ "opentelemetry-proto",
+ "prost",
+ "serde",
+ "serde_json",
+ "serde_yaml",
+ "slatedb",
+ "tempfile",
+ "tokio",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


apurvam:
PR unintentionally created. Will add payload-inspection features onto the existing buffer-cli

PR #439, opened 2026-05-08
vector rfc-0006: add full-text search using BM25 scoring

rodesai:
## Summary

RFC for adding FTS to vector

### DIFF
--- a/vector/rfcs/0006-text-search.md
+++ b/vector/rfcs/0006-text-search.md
@@ -174,6 +174,7 @@ subsequent block. It is typical (e.g. in Lucene) to further improve on this by a
 blocks for larger ranges (e.g. every 32 blocks in Lucene).
 
 **Impacts**
+
 To support cheaply computing max scores for a window, you can store max score statistics for the 
 documents covered by the Skip block. In particular, you can store the "dominating" term 
 frequencies (F), length (L) pairs, called Impacts, of the contained documents. Impacts have a 
@@ -252,8 +253,7 @@ then written 4 more times in SR0 before it's compacted away) to the write amplif
 for the FTS data, so it's roughly doubled. This is not ideal, but feels manageable.
 
 To address 3 and 5 we'll simply load the bitmap from storage after every flush and put it in the 
-`LastDbSnapshot` published by the Flusher. For 100M documents with a 20% deletions threshold, 
-the bitmap should be at most ~4MB so this should be a quick read, typically out of cache.
+`LastDbSnapshot` published by the Flusher.
 
 ## Configurations
 
--- a/vector/rfcs/0006-text-search.md
+++ b/vector/rfcs/0006-text-search.md
@@ -1,0 +1,741 @@
+# RFC-0006: Full Text Search
+
+**Status**: Draft
+
+**Authors**:
+- [Rohan Desai](https://github.com/rodesai)
+
+## Summary
+
+This RFC details the design for supporting full text search (FTS) in OpenData Vector.
+
+## Motivation
+
+Modern search use cases typically require a combination of semantic search and full text search.
+Semantic search can easily miss documents if the user's query is interested in specific keywords 
+(proper nouns, emails, etc). To mitigate this, applications can combine semantic and full text 
+search results to improve overall search quality.
+
+## Overview
+
+Like ANN, FTS returns the top K documents that are closest to some query. Search requires 
+computing a set of candidate documents, assigning the documents some relevance score, and then 
+returning the documents with the highest score. Unlike ANN, FTS scores documents based on the 
+presence of query terms rather than distance to an embedding.
+
+### BM25
+
+These days the de-facto ranking function for scoring documents is BM25. BM25 scores documents 
+using a "bag of words" model. It factors in only the frequency of each query term in a document 
+(along with some statistics about each term and the corpus), so it's disjunctive and ignores 
+position. More specifically, it combines the inverse-document frequency (IDF) for each term with 
+the frequency of each term normalized by the document length and saturated using static 
+parameters. IDF represents how infrequent a given term is. It increases as the number of 
+documents with a given term decreases.
+
+For a query `q` and document `D`, BM25 is commonly written as:
+
+$$
+\mathrm{score}(D, q) =
+\sum_{t \in q}
+\mathrm{IDF}(t)
+\cdot
+\frac{
+f(t, D)(k_1 + 1)
+}{
+f(t, D) + k_1 \left(1 - b + b \cdot \frac{|D|}{\mathrm{avgdl}}\right)
+}
+$$
+
+Where:
+
+- \(t\): a query term
+- \(f(t, D)\): frequency of term \(t\) in document \(D\)
+- \(|D|\): length of document \(D\)
+- \(avgdl\): average document length in the corpus
+- \(k_1\): term-frequency saturation parameter, often around `1.2` to `2.0`
+- \(b\): document-length normalization parameter, often around `0.75`
+
+A common IDF variant is:
+
+$$
+\mathrm{IDF}(t) =
+\log \left(
+\frac{N - n(t) + 0.5}{n(t) + 0.5} + 1
+\right)
+$$
+
+Where:
+
+- \(N\): total number of documents in the corpus
+- \(n(t)\): number of documents containing term \(t\)
+
+To efficiently score documents, search databases do the following:
+
+First, when ingesting documents, each document is tokenized to break it up into discrete terms. 
+
+The database stores per-term posting lists. Each posting list stores the set of documents that 
+contain the term, along with the term's frequency in the document (`f(t, D)`) and (often a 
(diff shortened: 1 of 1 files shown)


vector/rfcs/0006-text-search.md
```diff
@@ -0,0 +1,741 @@
...
+      2. Compute `I`'s max score contribution for the window using `I.max_score()` and `Min(TOPK)`
+   3. Compute which iterators are essential/non-essential for the window.
+   4. Score all documents for the essential iterators.
+   5. Update `TOPK` with the scored documents.
+
+### Field Norms
+
+We'll store field norms rather than document lengths in postings. This reduces posting size and
+overhead for tracking block-level scoring statistics. The norm is the field length quantized to
+1-bit. We'll use the same quantization formula as lucene (
```
cadonna: Should this be 1 bytes instead of 1 bit?

cadonna:
Thanks for the RFC, @rodesai !

I am wondering if we should separate the full-text search completely from the vector search to follow the guideline "Do One Thing And Do It Well" from UNIX. As far as I can see, the storage design already separates data for the full-text search from the data for the vector search. Also the queries seem to be separate for full-text and vector. Would it be possible to have different crates, so that the full-text search create can be used to build a database that only offers full-text search and the vector search crate can be used to build a database that only offers vector search and combining the two crates allows to build a database that offers both searches.

For example, a log search database would only need the full-text search crate.

vector/rfcs/0006-text-search.md
rodesai: oops yeah

rodesai:
> Thanks for the RFC, @rodesai !
I am wondering if we should separate the full-text search completely from the vector search to follow the guideline "Do One Thing And Do It Well" from UNIX. As far as I can see, the storage design already separates data for the full-text search from the data for the vector search. Also the queries seem to be separate for full-text and vector. Would it be possible to have different crates, so that the full-text search create can be used to build a database that only offers full-text search and the vector search crate can be used to build a database that only offers vector search and combining the two crates allows to build a database that offers both searches.
For example, a log search database would only need the full-text search crate.

Yeah, I started thinking along the same lines but got stuck on how to model non-text attributes properly. Both FTS and vector support simple attributes that can be filtered on, and require storing an inverted index to filter efficiently. The problem is for vector search those need to be integrated very closely with the ANN index so that tree traversal factors in those filters for every tree expansion, which required the index to store maintain inverted indexes for centroids as well. I think what we _can_ do is have common storage formats and indexing logic for FTS (and to some extent inverted indexes) and reuse that for text search in other contexts (e.g. logs).

vector/rfcs/0006-text-search.md
```diff
@@ -0,0 +1,741 @@
...
+contain the term, along with the term's frequency in the document (`f(t, D)`) and (often a 
+quantized representation of) the document's size (`|D|`) in number of tokens. The database also 
+stores statistics about each term and the corpus. In particular, for each term it stores the 
+number of documents containing the term (`n(t)`) and for the corpus the total number of 
+documents (`N`) and total document length (used to compute `avgdl`).
+
+At query time, the database tokenizes the query, loads the posting lists for each term, and 
+scores the contained documents using the BM25 formula. It then returns the top documents.
+
+### MaxScore
```
agavra, line 87: I found this pretty confusing at first pass so I had a good discussion with the AIs others might find interesting:

----

## Blocks vs. windows

There are two kinds of "chunk" here. They're different.

**Blocks** — *static, set at indexing time.* Each posting list is stored on disk as a sequence of fixed-size blocks (RFC targets 256 docs per block). Immediately before each Posting block sits a **Skip block** that summarizes it: the last doc id in the block, the block's length on disk, and its **Impact stats** (the dominating `(freq, norm)` pairs from the docs inside).

The Impact stats let you compute "what is the highest BM25 score *any* doc in this block could contribute for this term?" — without reading the actual postings. That number is fixed at write time. Compaction can rewrite blocks, but during a query they're frozen.

Different terms have *different* blocks: `fox` only appears in 3 docs in our toy example, so its entire posting probably fits in one block. `quick` is in every doc, so it has many blocks.

**Windows** — *dynamic, chosen at query time.* A window is a doc-id range `[window_start, window_end]` that the query loop processes together. The query coordinator picks `window_end` each iteration of the outer loop.

## How `window_end` is picked

Not statically fixed — it's a heuristic, and the conventional choice is:

> Take the **minimum** of "the last doc id of the current block" across all the term iterators.

In other words: snap the window end to whichever term's current block runs out next. That way, each iteration of the outer loop "consumes" exactly one full block on at least one iterator. The other iterators may consume part of a block, or nothing.

This is the natural choice because each iterator's `max_score()` is cheap exactly when the window is aligned to that iterator's block boundary — you just read the Impact stats from the Skip block. Span a window across multiple blocks of one term and you have to take the max of multiple Impacts; smaller windows give tighter bounds but more loop overhead.

## What per-window `max_score()` actually is

`I.max_score()` for a window = the highest BM25 contribution any single doc *in this window* could get from term `I`, computed from the Impact stats of whichever block(s) of `I` overlap the window.

So per-window max scores aren't arbitrary numbers I made up — they're a property of which docs happen to live in that doc-id range, baked into the index at write time.

## Concrete walkthrough

Suppose blocks look like this (made up, ascending order for sanity):

```
quick blocks:  [docs 1-3, max=0.4]  [docs 4-6, max=0.6]  [docs 7-10, max=0.2]
brown blocks:  [docs 2,4,6, max=2.0]                      [docs 8,10, max=0.6]
fox blocks:    [docs 3,6,9, max=4.0]
```

Loop iteration 1:
- All three iterators start at the lowest doc in their first block. Last doc ids in current blocks: quick→3, brown→6, fox→9.
- `window_end = min(3, 6, 9) = 3`. **Window is `[1, 3]`.**
- `max_score` per term in `[1, 3]`: quick=0.4 (its block ends at 3), brown=2.0 (its block ends at 6, so use that block's Impact), fox=4.0 (same).
- Do the essential/non-essential math with `Min(TOPK)`, score whichever candidates, update TOPK.

Loop iteration 2:
- quick advances to its next block. Last-doc-ids now: quick→6, brown→6, fox→9.
- `window_end = 6`. **Window is `[4, 6]`.**
- quick's max for this window = 0.6 (new block). brown's max = 2.0 (still the same block). fox's max = 4.0.

Loop iteration 3:
- quick and brown both advanced. Last-doc-ids: quick→10, brown→10, fox→9.
- `window_end = 9`. **Window is `[7, 9]`.**
- quick=0.2, brown=0.6, fox=4.0. Etc.

Loop iteration 4:
- `window_end = 10`. **Window is `[10, 10]`.** fox iterator is exhausted, skip.

So the per-window maxes I wrote in the last message weren't statically fixed numbers — they're whatever the Impact stats say for the blocks overlapping each window. The numbers vary across windows precisely because each window sees a different combination of blocks, and each block summarizes a different set of docs.

## Why this matters

The RFC stores Impact stats on every Skip block (RFC line 500-505) specifically so this `max_score()` call is O(1) per iterator per window. Without the Impact stats you'd be stuck using the *global* per-term max — which is the original MaxScore algorithm and gives much weaker bounds (my first example).

Want me to walk through one full window now end-to-end, showing how the iterator API in the RFC (`advance_end`, `max_score`) ties to the Skip-block reads?

vector/rfcs/0006-text-search.md
```diff
@@ -0,0 +1,741 @@
...
+Actual cleanup of postings is done by a compaction filter. If the compaction doesn't target the 
+FTS segment, it does nothing, Otherwise, it:
+- Loads the deletions bitmap into memory from the `Merge` entries present in a compaction.
+- Filters out the deletion bitmap entries if compacting to SR 0.
+- Compacts posting lists by removing deleted vectors and re-aligning blocks to 
+  `target_posting_block_size`. As it does this it tracks the delta to the number of vectors with 
+  the posting's term
+- Applies the delta to the term statistics.
+- Removes `VectorFieldStatistics` entries for deleted vectors and accumulates field stats deltas.
+- Applies the delta to the field stats.
```
agavra: I was discussing this with Claude since I was wondering how we can guarantee we'll always have access to all of the fields. I guess the same applies to deletion vector, but it doesn't matter if it's not there since we just won't do anything.

**One gap worth flagging in the RFC**: step 5 only works if the `FieldStats` key for a field actually appears in the same compaction as the dependent `VectorFieldStats` entries. If a compaction touches upper SRs containing deletes but the FieldStats key only lives in a lower SR that isn't included, `self.field_deltas` is lost. In practice the indexer writes a FieldStats delta on every flush, so they stay co-resident — but the RFC leans on that implicitly without saying so. Could be worth a sentence pinning that down.

agavra (approved):
This looks really solid to me @rodesai - most of my questions/concerns were clarified by the LLMs :) I think the full-segment compaction is a totally acceptable simple way to maintain the index

vector/rfcs/0006-text-search.md
rodesai: Ah yeah that's true. I'll make a note in the RFC

PR #442, opened 2026-05-10
promql: add limitk aggregation operator (#88)

### LINKED ISSUE #88: [tsdb] implement the full set of aggregation operators
## Summary

PromQL's query language defines a number of different aggregation operators over instant vectors. We should extend the aggregation support in the evaluator to cover this set: https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators

## Motivation

Full PromQL support.

petrpan26:
## Summary

Adds evaluator support for `limitk` with Prometheus-compatible grouping and filter-shape output.

<hr>

- add limitk aggregation reusing the topk/bottomk per-group heap as a push-only buffer
- walk inputs in fingerprint-sorted order for stable selection
- extend lowering and physical planning to recognize limitk
- add limitk-focused operator tests
- add limitk parity fixtures to `aggregators.test`

## Related Issues

Fixes #88

## Test Plan

- `cargo test -p opendata-timeseries --lib`
- `cargo clippy -p opendata-timeseries --lib --all-targets -- -D warnings`

## Checklist

- [x] Tests added/updated


petrpan26:
Added a draft PR here for review purposes + testing blocking on https://github.com/GreptimeTeam/promql-parser/issues/142 for a new release so we use an official version.

timeseries/src/promql/operators/aggregate.rs
```diff
@@ -633,8 +640,11 @@ impl<C: Operator> AggregateOp<C> {
         let schema = OperatorSchema::new(SchemaRef::Static(output_schema.clone()), grid);
         debug_assert!(
             param_child.is_none()
-                || matches!(kind, AggregateKind::Topk(_) | AggregateKind::Bottomk(_)),
-            "dynamic aggregate params are only supported for topk/bottomk",
+                || matches!(
+                    kind,
+                    AggregateKind::Topk(_) | AggregateKind::Bottomk(_) | AggregateKind::Limitk(_),
+                ),
+            "dynamic aggregate params are only supported for topk/bottomk/limitk",
```
cadonna: There is also a comment to update on line 666. I would prefer to remove the specific aggregations `topk`, `bottomk`, and `quantile` from that comment since it is only important that the aggregation function is a breaker.

timeseries/src/promql/operators/aggregate.rs
```diff
@@ -949,13 +961,15 @@ impl<C: Operator> AggregateOp<C> {
     }
 
     /// Finalise breaker output from the buffered full grid. Routes to
-    /// the kind-specific `finalise_topk_or_bottomk` / `finalise_quantile`
-    /// helpers, each of which runs its per-step selection / quantile
-    /// against the complete `(step_count × input_series_count)` grid.
+    /// the kind-specific `finalise_topk_or_bottomk` / `finalise_limitk` /
+    /// `finalise_quantile` helpers, each of which runs its per-step
+    /// selection / quantile against the complete `(step_count ×
+    /// input_series_count)` grid.
```
cadonna: Could you please remove the actual names of the helpers? 
Such comments are hard to maintain.
The names are already present in the code.

timeseries/src/promql/operators/aggregate.rs
```diff
@@ -1037,6 +1051,85 @@ impl<C: Operator> AggregateOp<C> {
...
+                let group = match self.group_map.input_to_group[series_us] {
+                    Some(g) => g as usize,
+                    None => continue,
+                };
+                let heap = &mut self.heaps[group];
+                if heap.len() >= k_usize {
+                    continue;
+                }
+                // KOrder is unused — limitk never evicts.
+                heap.push(KHeapEntry {
```
cadonna: Why do we need to first push the series samples to a `heap` and then from the `heap` to `out`? Why can we not set the samples directly in `out`?

timeseries/src/promql/plan/lowering.rs
```diff
@@ -679,7 +679,7 @@ fn lower_aggregate(
         T_STDDEV => Ok(aggregate_streaming(AggregateKind::Stddev, child, grouping)),
         T_STDVAR => Ok(aggregate_streaming(AggregateKind::Stdvar, child, grouping)),
         T_GROUP => Ok(aggregate_streaming(AggregateKind::Group, child, grouping)),
-        T_TOPK | T_BOTTOMK => {
+        T_TOPK | T_BOTTOMK | T_LIMITK => {
```
cadonna: IMO, it would be better to have a branch for each of the three constants and extract the code in lines 683-708 to a function that is called in each of the three branches. Nesting matching `op_id` does not seem clean to me.

timeseries/src/promql/plan/lowering.rs
```diff
@@ -770,7 +771,7 @@ fn aggregate_streaming(
     }
 }
 
-fn aggregate_op_name(op_id: u8) -> &'static str {
+fn aggregate_op_name(op_id: u16) -> &'static str {
```
cadonna: Why not `TokenId`?

timeseries/src/promql/plan/lowering.rs
```diff
@@ -832,7 +834,7 @@ fn lower_binary(bin: &parser::BinaryExpr, ctx: &LoweringContext) -> Result<Logic
     })
 }
 
-fn binary_op_kind(op_id: u8, return_bool: bool) -> Option<BinaryOpKind> {
+fn binary_op_kind(op_id: u16, return_bool: bool) -> Option<BinaryOpKind> {
```
cadonna: Same here?

cadonna (commented):
Thanks for the PR, @petrpan26 !

I reviewed the production code and had a couple of comments.

I am aware the some comments are not directly related to your code.

timeseries/src/promql/operators/aggregate.rs
petrpan26: Resolved

timeseries/src/promql/operators/aggregate.rs
petrpan26: Resolved

timeseries/src/promql/plan/lowering.rs
petrpan26: That's true, extracted out a helper function.

timeseries/src/promql/operators/aggregate.rs
petrpan26: Hi I think I got a very bad case of dumb ai coding in this PR. Upstreams code were reusing the same allocated heap for topk and bottomk and limitk. And i think i blindly follow it. I think our code structure is completely different.

petrpan26: Leave this here so you can actually review the new logic

petrpan26:
@cadonna Thanks for the review I think I was too enthusiastic when upstream deps was merged and forgot to do an actual code sweep.

apurvam:
Hi @petrpan26 . This looks good. I think the only thing that's missing is to add a memory reservation, like the `topk` functions do. See `should_respect_memory_reservation_for_topk_heap` for a test that throws a `MemoryLimit` error when the topk function exceeds it's memory reservation. 

TBH, I'm not exactly sure how this memory limits work, but it's better to be consistent here IMO.

timeseries/src/promql/operators/aggregate.rs
```diff
@@ -1037,6 +1053,72 @@ impl<C: Operator> AggregateOp<C> {
...
+        let mut counts = vec![0u32; group_count];
+
+        for global_step in 0..step_count {
+            let k_usize = self.k_for_step(global_step, in_series_count, k);
+            counts.fill(0);
+            if k_usize == 0 {
+                continue;
+            }
+            let grid_base = global_step * in_series_count;
+            let out_base = global_step * in_series_count;
```
cadonna, line 1082: `grid_base` and `out_base` are always equal, right?
Same question about `out_idx` and `cell`.
If yes, you can just use `grid_base` and `cell`.

cadonna:
> Hi @petrpan26 . This looks good. I think the only thing that's missing is to add a memory reservation, like the `topk` functions do. See `should_respect_memory_reservation_for_topk_heap` for a test that throws a `MemoryLimit` error when the topk function exceeds it's memory reservation.
> 
> TBH, I'm not exactly sure how this memory limits work, but it's better to be consistent here IMO.

I agree that a test is needed for the memory reservation. However, as limitk does not use a heap, it should only test for the reservation of the grid (line 670 in `aggregates.rs`). Additionally, as far as I can see test `should_respect_memory_reservation_for_topk_heap()` does not test what is suggests it tests. The verified error is caused by the grid reservation, not by the heap reservation (line 698 in `aggregates.rs`). Ideally, we have the following tests:
- for each of topk, bottomk, limitk, and quantile one test where the reservation of the grid exceeds the cap
- for each of topk, bottomk one test where the heap exceeds the cap
- for quantile one test where the sort buffer exceeds the cap
- 
@petrpan26 if you have cycles, you are welcome to fix that in a separate PR.

timeseries/src/promql/promqltest/testdata/aggregators.test
```diff
@@ -398,6 +398,47 @@ eval instant at 50m topk(scalar(foo), http_requests)
 	http_requests{group="canary", instance="0", job="app-server"} 700
 	http_requests{group="production", instance="1", job="app-server"} 600
 
+# Tests for limitk.
```
cadonna, line 401: For these tests did you also consider the tests here: https://github.com/prometheus/prometheus/blob/main/promql/promqltest/testdata/limit.test ?

cadonna (commented):
Thanks for the updates @petrpan26 !

The algorithm looks much better now.

I have one minor comment for the algorithm and two comments for tests.

cadonna:
@petrpan26 do you plan to update this PR?
If not, we can merge as is and I can include my feedback on a separate PR.
Just let us know. Either way is fine. We appreciate your contribution.

petrpan26:
@cadonna im afk this week. Do you mind if i take this on this weekend. If this block anything pls feel free to merge it

cadonna:
@petrpan26 It is not blocking anything. I just wanted to understand if the PR is still on your radar. Great that you plan to work on the PR!

petrpan26:
Looking into this tomorrow !

PR #447, opened 2026-05-14
[vector] Replace snapshot Mutex with ArcSwap for lock-free reads — 1.5x QPS, up to 2.2x lower p99

Venkat2811:
## Summary

`VectorDb::query_engine()` takes a `Mutex<LastAppliedSnapshot>` on every read at `vector/src/db.rs:840`, serializing every query through one lock. The snapshot is read-mostly and replaced wholesale on flush — clean fit for `arc_swap::ArcSwap`, which gives lock-free reads with atomic publication.

Identified while experimenting with "puffer" architecture vector databases and SlateDB-backed snapshot patterns. Change is type-substitution only; ~30 LOC core diff, no public API surface affected. Skipped RFC per CONTRIBUTING ("for bug fixes and minor features, skip the RFC and go straight to a PR").

## Related Issues

n/a

## Test Plan

**Two new lib tests** in `vector/src/db.rs` (multi-thread tokio runtime, given/when/then pattern, existing `create_test_config` fixture):

- `should_serve_concurrent_readers_against_snapshot` — 16 reader tasks x 25 queries each on a 50-vector index. Asserts every query returns hits and no task panics. Surfaces deadlocks, mutex poisoning, or torn-snapshot reads.
- `should_not_block_readers_during_concurrent_flush` — 8 hot-loop readers across 4 interleaved write+flush batches. Asserts each reader completes at least one search and every result set is non-empty.

331 lib tests pass (329 prior + 2 new). `cargo fmt --all` clean. `cargo clippy -p opendata-vector --all-targets --all-features -- -D warnings` clean.

### HTTP wire bench

cohere1M dataset (1M f32 768-dim cosine), HTTP POST `/api/v1/vector/write` ingest at batch=100 vectors/request (axum default 2 MB body cap), 1000 unique queries from `queries.f32`, k=10, nprobe=100, 100-query prewarm, c in {1, 16, 32, 64}. Two independent post-change runs to characterize variance.

**Environment:** Linux 6.8 x86_64, Ryzen 7 5800X (8c/16t), 62 GiB RAM, NVMe SSD; local MinIO single-node; rustc 1.97.0-nightly. Server: `SlateDb on Aws`, `dimensions=768`, `distance_metric=Cosine`, `flush_interval=5s`, `split_threshold_vectors=200`, `merge_threshold_vectors=50`, `block_cache_bytes=1 GiB`. Query client: Python `urllib` + `ThreadPoolExecutor` (no HTTP keep-alive; same TCP-handshake tax applied to pre and post).

**Results**

| c  | QPS pre | QPS post (run A) | QPS post (run B) | p99 pre (ms) | p99 post run A (ms) | p99 post run B (ms) | recall@10 pre -> post |
|---:|--------:|-----------------:|-----------------:|-------------:|--------------------:|--------------------:|:---------------------|
|  1 |   10.17 |            15.33 |            14.98 |          118 |                  78 |                  82 | 0.9238 -> 0.9256     |
| 16 |    9.82 |            15.95 |            15.92 |        3,308 |               1,509 |               1,507 | 0.9238 -> 0.9256     |
| 32 |    9.72 |            15.76 |            15.43 |        8,988 |               4,803 |               3,288 | 0.9238 -> 0.9256     |
| 64 |    9.66 |            15.63 |            13.76 |       26,201 |              16,262 |               8,139 | 0.9238 -> 0.9256     |

Across both post-change runs: QPS gain 1.47x–1.62x at every cell; p99 reduction 1.44x–2.19x at low concurrency, 1.61x–3.22x at high concurrency. Recall@10 unchanged within noise. Errors zero across all post-change cells (pre-change c=64 had 3 errors out of 1000 queries; post-change zero).

The QPS curve remains essentially flat across c={1, 16, 32, 64} after this change — removing the snapshot lock is a measurable win but other serializers remain on the read path. Identifying them is out of scope for this PR.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (if applicable) — PR description carries the rationale; no doc files touched

**Note:** ideation by me; implementation, tests, and verification done with `claude-code`. I take full responsibility and ownership of changes in this PR; happy to iterate on feedback and get this merged.

### DIFF
--- a/vector/src/db.rs
+++ b/vector/src/db.rs
@@ -48,7 +48,9 @@ use common::storage::{Storage, StorageRead, StorageSnapshot};
 use common::{StorageBuilder, StorageSemantics};
 use rayon::prelude::*;
 use std::collections::{HashMap, HashSet};
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
+
+use arc_swap::ArcSwap;
 use std::time::Duration;
 
 pub(crate) const WRITE_CHANNEL: &str = "write";
@@ -135,8 +137,17 @@ pub struct VectorDb {
     /// The WriteCoordinator itself (stored to keep it alive).
     write_coordinator: WriteCoordinator<VectorDbOpDelta, VectorDbFlusher>,
 
-    /// snapshot state for queries
-    last_applied_snapshot: Arc<Mutex<LastAppliedSnapshot>>,
+    /// snapshot state for queries.
+    ///
+    /// `ArcSwap<LastAppliedSnapshot>` lets concurrent query threads
+    /// observe the current snapshot without taking a lock — reads are
+    /// a single atomic load plus an Arc-clone. The flusher publishes a
+    /// new snapshot via `store(Arc::new(new))` which replaces the
+    /// pointer atomically. In-flight queries holding a Guard against
+    /// the previous snapshot observe the complete prior view through
+    /// to completion (ArcSwap defers drop until the last Guard
+    /// releases).
+    last_applied_snapshot: Arc<ArcSwap<LastAppliedSnapshot>>,
 }
 
 impl VectorDb {
@@ -240,7 +251,7 @@ impl VectorDb {
             config.distance_metric,
         )
         .await?;
-        let last_applied_snapshot = Arc::new(Mutex::new(LastAppliedSnapshot {
+        let last_applied_snapshot = Arc::new(ArcSwap::from_pointee(LastAppliedSnapshot {
             snapshot: snapshot.clone(),
             centroid_cache: Arc::new(loaded_tree.centroid_cache.cache()),
             centroid_index: loaded_tree.query_centroid_index,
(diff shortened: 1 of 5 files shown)


apurvam:
thanks @Venkat2811 .. @rodesai would be great for you to have a look!

vector/src/db.rs
```diff
@@ -814,20 +825,13 @@ impl VectorDb {
...
+        self.last_applied_snapshot.load().centroid_count
     }
 
     pub async fn validate_cache(&self) {
-        let las = self
-            .last_applied_snapshot
-            .lock()
-            .expect("lock_poisoned")
-            .clone();
+        let guard = self.last_applied_snapshot.load();
```
hachikuji: The old code cloned the LAS. I am not that familiar with arc-swap, but is there any downside to holding the guard?

hachikuji (commented):
Nice find! LGTM overall. I had just one minor question.

vector/src/db.rs
Venkat2811: thanks ! good catch. Guard across `.await` defers writer reclamation under concurrent `store`. fixed

Venkat2811:
my apologies, there was an issue in my benchmarks setup yesterday. closing the pr @apurvam @agavra @hachikuji @rodesai 

minio mount state was reused n re ingested. data was not wiped clean, there was separate stale bench process running one my machine too

was able to produce 15QPS w/o my changes as well

using arcswap or not is separate discussion, as it doesn't produce any perf improvement as claimed, i'm closing this.

sry again, got excited by perf improvement numbers, i made sure to verify it, but missed the bucket state.

apurvam:
Thanks for the diligence @Venkat2811 . Much appreciated!

PR #448, opened 2026-05-15
Upgrade to slatedb 0.13

hachikuji:
This is a claude one-shot. Looks right to me. The only interesting part to me is the use of PsyncIoEngine. We could consider enabling UringIoEngine when the target is linux.


agavra (approved):
LGTM, will require blowing out the Foyer cache on any deployed instance with it configured (just an FYI)

PR #449, opened 2026-05-19
Implement LogDb count api

hachikuji:
Implements the count API for LogDb. I've removed the approximate count option, which I thought was kind of useless. For lag computation, it might be better to have smarter caching of intermediate results. But for a first pass, we just rely on the  block cache. Each count query is divided into one portion which uses slatedb BlockStats on the written SSTs, and one portion which scans keys in memtables and does a direct count.

### DIFF
--- a/common/src/storage/sst_blocks.rs
+++ b/common/src/storage/sst_blocks.rs
@@ -1,0 +1,690 @@
+//! Precise key-range record counts over a SlateDB manifest.
+//!
+//! Built on SlateDB RFC 0020 primitives — [`slatedb::Db::manifest`],
+//! [`slatedb::SstReader`], [`slatedb::SstFile::stats`],
+//! [`slatedb::SstFile::index`] — plus [`slatedb::SstFile::read_block`] for
+//! per-row precision on boundary blocks.
+//!
+//! The primary entry point is [`count_in_range`], which walks the manifest
+//! and returns the exact number of physical write operations (puts,
+//! deletes, merges) recorded within a query range.
+//!
+//! ## Walk structure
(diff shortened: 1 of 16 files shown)


common/src/storage/sst_blocks.rs
```diff
@@ -0,0 +1,690 @@
...
+    let last_entry = sst_file.info().last_entry.clone();
+
+    let n = index.len();
+    let overlapping: Vec<usize> = (0..n)
+        .filter(|i| ranges_overlap(query, &block_key_range(&index, *i, last_entry.as_ref())))
+        .collect();
+
+    // Back-scan: walk overlapping blocks high-to-low, reading each, until we
+    // find one with actual in-query rows. SlateDB's index returns truncated
+    // separators (`sep[i] <= first_key(block i)`), so a block's separator can
```
cadonna: What are truncated separators?
Worth explaining it here or linking to a place where it is explained.

cadonna (approved):
Thanks for the PR, @hachikuji !

I have just one nit.

cadonna:
`cargo fmt` is not happy 🙁

common/src/storage/sst_blocks.rs
hachikuji: Claude reworded the comment. Basically the keys stored in the index are not actual block start keys, but prefixes: i.e. the shortest prefix which provides an upper bound of the prior block. This leads to an ambiguity where the prefix key from the index falls within the desired range, but the actual start key is outside. In this case, we have to go back to the previous block to find the actual end of the count query.

PR #450, opened 2026-05-19
Log: integration with slatedb segments

hachikuji:
Since segments in Slatedb are prefix-based, this patch required some rejiggering of Logdb paths. The prior structure didn't work because record type came before segment_id in the key schema. This means that listing records and log records would be divided into separate segments. To fix it, I had to move segment_id ahead of the record type. Since `SeqBlock` keys do not fall within a segment, I decided to use a sentinel system segment (0x0). This gives all keys in LogDb, a common segment prefix. Within the compaction scheduler (deferred for subsequent work), the system segment will not be eligible for draining.

These changes put LogDb at odds with the common key structure, which specifies a record tag in the shared prefix. My feeling is that the common prefix was going a little too far by getting into the record type structure. We take some of the flexibility out of systems without getting much in return. In any case, timeseries is going to face the same problem with the key structure, so some change there is inevitable. If the change makes sense, a follow-up can revise the common key codec library.

### DIFF
--- a/log/rfcs/0002-logical-segmentation.md
+++ b/log/rfcs/0002-logical-segmentation.md
@@ -115,7 +115,9 @@ Living in the system segment decouples metadata from the per-segment
 SlateDB lifecycle: when a user segment is drained and pruned from the
 SlateDB manifest, the `SegmentMeta` record for it is not automatically
 removed and must be deleted by the writer as it observes the manifest
-update. See [Retention](#segment-retention-future) below.
+update. The exact cleanup mechanism is out of scope for this RFC and is
+called out under [Segment-Based Deletion](#segment-based-deletion) as
+future work.
 
 #### Metadata Lifecycle
 
--- a/log/rfcs/0002-logical-segmentation.md
+++ b/log/rfcs/0002-logical-segmentation.md
@@ -46,11 +46,22 @@ Key properties of segments:
 
 ### Key Encoding with Segments
 
-The log entry key format is extended to include the segment ID:
+Every log storage key — log entries, listings, segment metadata, and the
+global sequence block — shares a 7-byte segmented prefix:
+
+```
+| subsystem (u8) | version (u8) | segment_id (u32 BE) | record_type (u8) | ... record-specific fields ... |
+```
+
+Placing `segment_id` before `record_type` keeps every record for a given
+segment contiguous in storage regardless of record type. This is the
+property a SlateDB segment extractor relies on: with a fixed 6-byte prefix
+`[subsystem, version, segment_id]`, all records for a segment land in the
+same SlateDB segment.
 
 ```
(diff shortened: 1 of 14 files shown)


log/rfcs/0002-logical-segmentation.md
```diff
@@ -64,13 +75,33 @@ This encoding ensures:
 - Within a key, entries are ordered by sequence number
 - Prefix scans can be constrained to specific segments
 
+### The system segment
```
agavra, line 78: this is actually really cool, opens the door to some really efficient metadata-only operations (e.g. a CLI or soemthign that just pulls the metadata segment)

log/src/segment_extractor.rs
```diff
@@ -0,0 +1,462 @@
...
+
+    fn prefix_len(&self, target: &PrefixTarget) -> Option<usize> {
+        // The `Point` and `Prefix` variants share the same answer here: the
+        // extracted bytes are determined entirely by positions 0..6, which
+        // both variants observe directly.
+        let bytes: &[u8] = match target {
+            PrefixTarget::Point(b) | PrefixTarget::Prefix(b) => b.as_ref(),
+        };
+        if bytes.len() < ROUTING_PREFIX_LEN {
+            return None;
```
agavra: interesting, I wonder what we're going to do when we remove `Option`. should we just panic here instead of returning none?

agavra (approved):
Design LGTM, skimmed the code and it makes sense. Focused on the prefix extractor

log/rfcs/0002-logical-segmentation.md
```diff
@@ -80,6 +111,12 @@ The metadata tracks:
 
 End boundaries (end sequence, end time) are derived from the next segment's start values, or from the current log state for the active segment.
 
+Living in the system segment decouples metadata from the per-segment
+SlateDB lifecycle: when a user segment is drained and pruned from the
+SlateDB manifest, the `SegmentMeta` record for it is not automatically
+removed and must be deleted by the writer as it observes the manifest
+update. See [Retention](#segment-retention-future) below.
```
apurvam: Did I miss this section? I tried searching for 'Retention' but couldn't find a section related to it.

apurvam (approved):
Left a question. But this LGTM. I think it's right that we don't prescribe the record type being at byte 2 globally.

log/src/segment_extractor.rs
hachikuji: Yeah, panic is better.

log/rfcs/0002-logical-segmentation.md
hachikuji: Optimistic addition by claude I guess. I removed the broken link. I was planning to do retention semantics in the next follow-up PR.

PR #451, opened 2026-05-20
Update common serde for changes in prefix scheme

hachikuji:
In #450, we updated the common prefix to only specify subsystem and version. This patch updates the serde library. I've also added subsystem.rs as a common place to keep subsystem ids.


apurvam (approved):
This seems pretty mechanical. LGTM!

PR #452, opened 2026-05-21
RFC on log compaction and retention

hachikuji:
This proposal focuses on segment-aware compaction and retention for LogDb. The main points of interests are 1) using the system segment as the source of truth to trigger draining, and 2) the compaction strategies for active and inactive segments.

### DIFF
--- a/log/rfcs/0005-compaction-and-retention.md
+++ b/log/rfcs/0005-compaction-and-retention.md
@@ -1,0 +1,551 @@
+# RFC 0005: Compaction and Retention
+
+**Status**: Draft
+
+**Authors**:
+- Jason Gustafson <12502538+hachikuji@users.noreply.github.com>
+
+## Summary
+
+This RFC defines OpenData-Log's compaction strategy on top of SlateDB, and
+the retention semantics that ride on top of it. Retention physically
+removes data by issuing drain compactions; the same custom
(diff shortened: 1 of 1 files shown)


apurvam (approved):
I read through this a couple of times. It is pretty elegant and LGTM.  

One of the key assumptions I think is that you can do atomic writes across segments: when a new segment gets the first write, it is critical to write the `SegmentMeta` to the system segment atomically with the bytes in the actual segment. I'm assuming that that is trivial in Slate?

hachikuji:
> One of the key assumptions I think is that you can do atomic writes across segments: when a new segment gets the first write, it is critical to write the SegmentMeta to the system segment atomically with the bytes in the actual segment. I'm assuming that that is trivial in Slate?

I am not sure they need to be atomic. Just the SegmentMeta must get written first. The way LogDb works is that any write failure is fatal and causes all subsequent writes to fail. So if SegmentMeta fails to be written, then no writes on the segment will succeed.

hachikuji:
I guess there is a bit more nuance. We take advantage of the fact that writes in SlateDb are serialized, regardless of segment destination. If the WAL were disabled (probably a common setting for Log), then it is possible for the segment log data to be uploaded to object storage before the segment metadata. This is still OK because visibility is controlled by the manifest, which ensures that all tables added to the manifest reflect the sequential ordering.

PR #454, opened 2026-05-22
[common] add list_segments

cadonna:
## Summary

Adds accessors to the storage exposing the SlateDB manifest's segment list to subsystems that need bucket-style discovery (e.g. timeseries buckets) without re-scanning record keys. On SlateDB-backed storage the list of segments is read from the manifest subscription on each call. Snapshots capture the list at snapshot creation, returning a frozen view — SlateDB does not let us pin a DbSnapshot to a specific sequence number, so the snapshot's segment list is documented as an upper bound on segments containing flushed visible data.

InMemoryStorage takes an optional slatedb::PrefixExtractor via with_prefix_extractor(...), runs it over every written key, and reads the resulting prefixes from a BTreeSet.

## Test Plan

- unit tests

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/common/src/storage/in_memory.rs
+++ b/common/src/storage/in_memory.rs
@@ -134,24 +134,23 @@ impl InMemoryStorage {
         self
     }
 
-    /// Records `key`'s routing prefix in the segment set, if a prefix
+    /// Records each key's routing prefix in the segment set, if a prefix
     /// extractor is configured and accepts the key.
-    fn update_segments(&self, key: &Bytes) {
+    fn update_segments(&self, keys: &[Bytes]) {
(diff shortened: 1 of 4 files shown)


apurvam:
thanks @cadonna . the changes look fine to me. 

Semantically, however, how are users supposed to reason about the contract? It looks like `list_segments` against a snapshot can contain segments not in that snapshot (because segments are read straight from the manifest). And the snapshot can also include in-memory writes that don't have a segment yet. 

Is my understanding correct?

How would the timeseriesdb reconcile this when trying to list buckets?

cadonna:
This PR is superseded by https://github.com/slatedb/slatedb/pull/1731

PR #458, opened 2026-05-22
Log: use prefix bloom filters

hachikuji:
LogDb keys are roughly structured as `segment + log_key + sequence`. The default bloom filter from SlateDb uses the full key, but this just wastes space. Instead, we can use Slatedb's fancy new prefix bloom filters and index just `segment + log_key`. The bloom filter tells us whether a log_key may be present in the table, and the index tells us exactly which blocks may have entries.


hachikuji:
Closing this for now. I think the scan logic needs a bit more thought.

PR #459, opened 2026-05-24
fix(timeseries): sort points before Gorilla encode to avoid overflow panic

### LINKED ISSUE #432: Out-of-order samples within a series cause subtract-with-overflow panic in tsz encoder
## Summary

The timeseries server panics with `attempt to subtract with overflow` from `tsz-0.1.4` whenever a series receives samples whose timestamps are not strictly monotonically increasing. The panic kills a tokio worker; if it happens on the buffer consumer's flush path, the consumer stalls and pending samples are lost.

## Repro

`TimeSeriesValue::encode()` in `timeseries/src/serde/timeseries.rs` feeds `self.points` straight into `StdEncoder`:

https://github.com/opendata-oss/opendata/blob/main/timeseries/src/serde/timeseries.rs#L177-L195

`StdEncoder::write_next_timestamp` does an unsigned subtraction:

```rust
fn write_next_timestamp(&mut self, time: u64) {
    let delta = time - self.time;
    ...
}
```

So a `Vec<Sample>` that isn't already time-ordered for the same series panics on encode. Equivalent to:

```rust
let v = TimeSeriesValue {
    points: vec![
        Sample { timestamp_ms: 200, value: 1.0 },
        Sample { timestamp_ms: 100, value: 2.0 }, // backwards -> panic
    ],
};
v.encode().unwrap();
```

Stack from a real run:

```
thread 'tokio-runtime-worker' (...) panicked at .../tsz-0.1.4/src/encode/std_encoder.rs:78:21:
attempt to subtract with overflow
   ...
   2: core::panicking::panic_const::panic_const_sub_overflow
```

## Suggested fix

Non-exclusive options, in rough order of robustness:

- Sort by timestamp inside `TimeSeriesValue::encode()` (or assert ordering and surface an error).
- Sort/merge per-series samples in `Tsdb::ingest_samples` before they reach tsz.
- Drop or quarantine out-of-order samples with a counter (e.g. `tsdb_out_of_order_samples_total`).
- `catch_unwind` at the encode boundary so a single bad batch does not kill the worker and stall the consumer.

Minimum acceptable behavior IMO: out-of-order sample = recorded metric and possibly dropped, not crashed worker.

## Environment

- opendata main branch as of early May 2026
- `tsz = "0.1.4"` per `Cargo.lock`

SAY-5:
Fixes #432.

`TimeSeriesValue::encode()` fed `self.points` straight into the Gorilla `StdEncoder`, whose `write_next_timestamp` does an unsigned `time - self.time` subtraction. Any series whose samples are not in non-decreasing timestamp order triggered `attempt to subtract with overflow` and killed the tokio worker.

This sorts the points by timestamp inside `encode()` before encoding, which also makes the `start_time = points[0]` assumption correct for unordered input. The merge path already produces sorted output, so this only changes behavior for out-of-order batches.

Test plan:
- Added `should_encode_when_points_are_out_of_order`, which panicked with the exact `subtract with overflow` before the fix and passes after.
- `cargo test -p opendata-timeseries`, `cargo clippy -p opendata-timeseries --all-targets -- -D warnings`, and `cargo fmt --all` are clean.

### DIFF
--- a/timeseries/src/serde/timeseries.rs
+++ b/timeseries/src/serde/timeseries.rs
@@ -180,12 +180,17 @@ impl TimeSeriesValue {
             return Ok(Bytes::new());
         }
 
+        // Gorilla delta encoding requires monotonically non-decreasing timestamps,
+        // so sort defensively to avoid a subtract-with-overflow panic on out-of-order input.
+        let mut points: Vec<&Sample> = self.points.iter().collect();
+        points.sort_by_key(|p| p.timestamp_ms);
+
         // Use Gorilla compression
         let w = BufferedWriter::new();
-        let start_time = self.points[0].timestamp_ms as u64;
+        let start_time = points[0].timestamp_ms as u64;
         let mut encoder = StdEncoder::new(start_time, w);
 
-        for point in &self.points {
+        for point in &points {
             let dp = DataPoint::new(point.timestamp_ms as u64, point.value);
             encoder.encode(dp);
         }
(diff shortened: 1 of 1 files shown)


apurvam (approved):
Thanks @SAY-5 . This LGTM!

PR #462, opened 2026-05-27
[buffer] add consumer read-ahead

apurvam:
## Summary

Adds parallel-IO + batched-ack primitives to `buffer::Consumer` so a consumer can
fetch and decode many batches concurrently while keeping manifest mutation
serialized and batched. This is the buffer-side foundation for high-throughput
pipelined consumers; the existing serial `next_batch` / `ack` API is preserved
unchanged.

## What's included 
- New public API (additive; documented in `buffer/rfcs/0003-consumer-read-ahead.md`):
  - `BatchDescriptor` — a lightweight, fetch-without-re-reading pointer to a manifest entry.
  - `ConsumerFetchHandle` (`Clone + Send + Sync + 'static`) — stateless `fetch(&self, …)`, cloneable into N fetch workers.
  - `Consumer::next_descriptors(max)` — one manifest read → a contiguous run of descriptors.
  - `Consumer::fetch_handle()` / `fetch_descriptor()` — the handle, and the serial `&mut self` wrapper that the existing `next_batch` is implemented upon.
  - `Consumer::ack_through(seq)` — advance the durable frontier in a single dequeue CAS (dequeue-first; storage/fence errors leave state untouched).
- `next_batch` becomes a wrapper over `next_descriptors(1)` + `fetch_descriptor`; cancellation- and fetch-failure-safe (cursor advances only after a successful fetch).
- New `buffer.descriptors_handed_out` counter; no metric renames.
- RFC 0003 doc.

## Compatibility

Additive — `opendata-buffer` 0.2.0 → **0.3.0** (minor). `next_batch`'s signature is
unchanged; in-workspace consumers (`timeseries`, `vector`) build without changes.

## Testing

Several unit tests added to the consumer to test the new behavior + compatibility with the old behavior. this implementation was tested with a pipelined clickhouse consumer in opendata-contrib (with a separate PR and RFC) for both correctness and performance. It can sustain 1Gbps throughput on a single ingestor node with 4 core CPU and 8GB of ram. Even then the main bottleneck is clickhouse. 

image

## Reviewer notes

The RFC is intended to be the main review artifacts. The public interfaces and implementation pseudo code is included in the RFC. 

- [x] Tests added/updated

### DIFF
--- a/buffer/rfcs/0003-consumer-read-ahead.md
+++ b/buffer/rfcs/0003-consumer-read-ahead.md
@@ -8,25 +8,23 @@
 
 ## Summary
 
-Add a fetch-handle type and three methods to `buffer::Consumer` so
-high-throughput consumers can fetch and decode many data batches
-concurrently while keeping manifest mutation serialized:
+Add a fetch-handle type and three methods to `buffer::Consumer` so a consumer
+can fetch and decode many data batches **concurrently** while keeping manifest
+mutation **serialized and batched**:
 
 ```rust
 impl Consumer {
     pub async fn next_descriptors(&mut self, max: usize)
         -> Result<Vec<BatchDescriptor>>;
 
-    /// Convenience wrapper for serial callers. The high-throughput
-    /// runtime uses [`fetch_handle`] instead.
+    /// Convenience wrapper for serial callers; constructs a local
+    /// fetch handle and calls `fetch`.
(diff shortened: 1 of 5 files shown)


apurvam:
I'm going to go ahead and merge this. we can fix the apis forward if there is feedback.

PR #472, opened 2026-06-02
Log: key cardinality benchmark rfc

hachikuji:
In principle, LogDb ought to be able to handle high cardinalities of keys due to its LSM representation. Benchmarks like OMB are more suited to a relatively small number of consumers (say hundreds) which are actively polling as fast as they can. We need a different approach which allows a distribution over the polling frequency of each log and can scale to millions. This RFCS sketches out a basic approach for this benchmark.

### DIFF
--- a/log/rfcs/0006-cardinality-benchmark.md
+++ b/log/rfcs/0006-cardinality-benchmark.md
@@ -1,0 +1,320 @@
+# RFC 0006: Log Cardinality Benchmark
+
+**Status**: Draft
+
+**Authors**:
+- [Jason Gustafson](https://github.com/hachikuji)
+
+## Summary
+
+This RFC defines a benchmark for LogDb built around the use cases it is designed to
+serve: large populations of independent, key-addressable logs, each *followed* by its
+own consumer at its own pace. The model is a mailbox — messages arrive throughout the
+day, and each user polls now and then to catch up on what is new. The benchmark gives
+every key a consumer that tracks a cursor and polls for records appended since it last
+read; a distribution over polling frequency, plus occasional offline lag, produces a
+realistic mix of caught-up tail reads and deeper catch-up reads.
+
+The workload runs in real time under an open-loop scheduler, so its results are
+directly interpretable: poll latency, sustainable throughput, and object-store GET
+requests per poll — reported as a function of how far behind each consumer is. The two
+primary knobs are the **number of keys** and the **distribution over polling
+frequency**.
+
+## Motivation
+
+LogDb is designed for use cases that maintain many independent logs, each addressed by
+a key and followed at its own pace — the dominant pattern is "show me what's new in
+this one log since I last looked." These use cases recur across domains:
+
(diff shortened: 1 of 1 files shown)


agavra (commented):
This makes sense to me with two general points of feedback:

1. it would be nice to compare results to Kafka (which I assume will just fall over pretty quickly) and things like postgres queues/redis queues or any other system users might use
2. I wonder if we can ground the benchmark in a "story" that makes it more concrete than the abstract concept. that's more for the narrative, so we can do that later if need be

hachikuji:
@agavra Agreed on both points. I'll tweak the design to add an abstraction for the backend so that we can plug in alternatives. A compelling story will probably take a few iterations. If we're agreed with the emphasis on high cardinality, then we probably don't need to block here.

agavra (approved):
Yup!

apurvam:
This looks good to me. The comparison with Kafka will be interesting, and really depends on the story we tell. If indeed you have millions of independent consumers (one per key), Kafka is going to fall over very fast. Our story needs to highlight the situations where that is common.

log/rfcs/0006-cardinality-benchmark.md
```diff
@@ -0,0 +1,361 @@
...
+
+- **Arrivals.** A writer `append`s records to keys at a target arrival rate. The
+  aggregate write rate is `cardinality × arrival_rate_per_key` and must stay within the
+  system's ingest capacity — at high cardinality this forces a low per-key rate (e.g. a
+  message every few minutes), which is realistic for the target use cases. Arrival is
+  uniform across keys in the first pass; per-key skew is a possible later knob.
+- **Polls.** Each key has a consumer holding a `cursor`. A consumer polls on its own
+  schedule, drawn from the polling-frequency distribution, calling
+  `poll(key, cursor, page_size)` and storing the returned cursor; a deeply lagged
+  consumer drains over several polls rather than one unbounded read. An offline event
```
rodesai, line 179: If this happens does the consumer try to catch up fully, or does it stick to the poll interval yielded by the poll distribution? Also is the poll interval fixed per consumer (so each consumer polls regularly, but different consumers use different intervals), or does it vary according to the distribution?

log/rfcs/0006-cardinality-benchmark.md
```diff
@@ -0,0 +1,361 @@
...
+directly interpretable: poll latency, sustainable throughput, and object-store GET
+requests per poll — reported as a function of how far behind each consumer is. The two
+primary knobs are the **number of keys** and the **distribution over polling
+frequency**.
+
+## Motivation
+
+LogDb is designed for use cases that maintain many independent logs, each addressed by
+a key and followed at its own pace — the dominant pattern is "show me what's new in
+this one log since I last looked." These use cases recur across domains:
```
rodesai, line 28: Durable execution seems like a good use case for this as well (e.g. each durable function/workflow has its own log)

apurvam: That's a good call.

PR #475, opened 2026-06-03
[vector] query engine: split bm25/ann search into chain of operators

rodesai:
## Summary

Splits bm25/ann search into chain of operators like: project -> lookup -> ann/bm25.
This allows us to reuse the same projection/lookup operators or both ann and bm25, and
split up planning from query execution.

The search operators (`ANNOperator`, `BM25Operator`) each share a similar structure. The
operators traverse indexes and score vectors, and stream the scored vectors into an impl of
`Collector`, which does the actual TopK aggregation. The final top k are returned to the parent
operator which looks up vectors in the forward index.

## Test Plan

- unit tests for each operator
- e2e benchmarks/functional tests show no regression

## Checklist

- [x] Tests added/updated

### DIFF
--- a/vector/src/serde/term_postings.rs
+++ b/vector/src/serde/term_postings.rs
@@ -2,8 +2,8 @@
 //!
 //! A term's posting list is a sequence of typed blocks written one after
 //! another to the end of the value buffer. Each `Postings` block holds up to
-//! [`TARGET_POSTINGS_PER_BLOCK`] `(doc_id, freq, norm)` triples in strictly
-//! descending `doc_id` order. Each `Postings` block is preceded by a `Skip`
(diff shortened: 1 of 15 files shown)


vector/src/serde/term_postings.rs
```diff
@@ -2,8 +2,8 @@
 //!
 //! A term's posting list is a sequence of typed blocks written one after
 //! another to the end of the value buffer. Each `Postings` block holds up to
-//! [`TARGET_POSTINGS_PER_BLOCK`] `(doc_id, freq, norm)` triples in strictly
```
rodesai, line 5: the changes in this file just switch the naming over from "document" to "vector" to keep things consistent with the rest of the db

PR #478, opened 2026-06-10
Add log inspection api

hachikuji:
The `count` API walks the LSM tree in order to count the number of records for a log within a sequence range. It's useful to surface details from the walk as well in order to understand the structure and layout of the data in the log. This can help for example with tuning GC. To that end, this patch adds an `inspect` API which exposes the segments that were traversed and the l0s/srs consulted within each segment. Finally, we modify `count` to delegate to it.

### DIFF
--- a/common/src/storage/sst_blocks.rs
+++ b/common/src/storage/sst_blocks.rs
@@ -82,41 +82,80 @@ impl BlockOpCounts {
     }
 }
 
-/// LSM traversal statistics for a single [`count_in_range`] walk.
-///
-/// Captures how much of the tree the walk had to touch to produce the
-/// count — the raw material for read-amplification analysis. `l0_ssts` and
-/// `sorted_runs` are summed over the trees the walk visited: the
-/// unsegmented default tree plus every configured segment whose prefix
-/// interval overlaps `query` (see [`count_in_range`]). For a query confined
-/// to one segment's prefix — the LogDb case — they describe that one
-/// segment's shape. The remaining fields are scoped to the query range.
-///
-/// L0 SSTs are not range-partitioned within a tree, so the walk opens
-/// *every* L0 SST of a visited tree regardless of overlap — `ssts_opened`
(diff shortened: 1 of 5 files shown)


common/src/storage/sst_blocks.rs
```diff
@@ -81,6 +82,83 @@ impl BlockOpCounts {
...
+/// SSTs; writes not yet flushed are accounted separately by the caller.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct WalkStats {
+    /// The L0 tier's contribution; see [`L0Stats`].
+    pub l0: L0Stats,
+    /// The sorted-run tier's contribution; see [`SortedRunStats`].
+    pub sorted_runs: SortedRunStats,
+}
+
+/// Private per-tier accumulator the walk threads through `count_view`. The
```
apurvam: What's a tier in this context?

log/src/log.rs
```diff
@@ -499,11 +499,15 @@ impl LogRead for LogDb {
         Ok(view.scan_with_options(key, seq_range, &options))
     }
 
-    async fn count(&self, key: Bytes, seq_range: impl RangeBounds<Sequence> + Send) -> Result<u64> {
+    async fn inspect(
```
apurvam, line 503: So we are dropping `count` from the Log API, and only making it available in log reader?

apurvam (commented):
This looks good to me Jason. i have one question about the change to the log API that would be good to get to the bottom of before merging.

common/src/storage/sst_blocks.rs
hachikuji: I think tier refers to l0s or sorted runs. Let me rewrite the comments.

log/src/log.rs
hachikuji: Not dropping it. Just now we have a default implementation which delegates to `inspect`.

apurvam (approved):
LGTM, thanks!

PR #482, opened 2026-06-11
[log] Fix `build_active_l0_spec` to return the oldest L0s

apurvam:
The manifest's L0 list is newest-first; taking the first max_l0 entries compacted the newest L0s, which moved slatedb's compaction watermark past every older uncompacted L0 and the writer/compactor manifest merge then silently dropped them — destroying their data. Fires whenever an active segment's L0 count exceeds max_l0_per_compaction at compaction time (masked at default settings where the slatedb flush gate caps L0 count at the same value). Select the contiguous oldest suffix instead, preserving newest-first order since slatedb reads the watermark from the spec's first source.

(cherry picked from commit 5b0fadef374f3bc400ea6b3c5cf38d3afa7e0500)

- [x] Tests added/updated

### DIFF
--- a/log/src/compaction.rs
+++ b/log/src/compaction.rs
@@ -316,12 +316,22 @@ fn is_single_sr_steady(segment: &SegmentSnapshot) -> bool {
 }
 
 /// Builds an L0-relief compaction spec for an active segment.
+///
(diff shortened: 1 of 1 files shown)


hachikuji (approved):
LGTM!

PR #486, opened 2026-06-12
Logdb: support for prefix bloom filters

hachikuji:
Posting this as a draft. Once we have the upstream range-based api, then we can update this. Using `scan_prefix` will also open the door to custom filters, which will address one of the major gaps with scoping sequence range-based queries.


hachikuji:
I'm going to optimistically merge this since we know the subranged prefix scan is on the way.

PR #487, opened 2026-06-12
Logdb: efficient tail scan rfc and preliminary work

hachikuji:
My early read benchmarks showed that polling at the tail of the log was expensive even when there was no data to return. This is especially evident in high cardinality key tests with sparser data. The global sequence provides one path to improve efficiency. The basic intuition is that we can use the advancement of the global sequence as a way to avoid repeated work on the reader. At the moment, the reader only advances its sequence when there is new data to return from the log key it is reading. Instead, we can advance the reader's position along with the global sequence number.  In principle, this lets us search a smaller portion of the tree. If the reader knows that the global sequence number is at N and a search for a key at  M < N returned nil, then it's safe to begin the next search at N and avoid the work in the range [M, N). 

However, this only has a benefit if we can use the advanced sequence N to filter parts of the tree that couldn't be filtered with the sequence M. This is definitely possible when N lands the reader in a new segment. If M were contained in an older segment, then the key constructed for N will exclude the older segment entirely. It is also possible to exclude SSTs within the active segment, but achieving that requires a custom filter policy which stores the global sequence range on each SST. We can then use our query range to intersect the global range of each table. The combined effect of this is that we should be able to keep tail queries on the most recent l0s, even for keys which do not have recent data.

The patch here implements the first pieces of this. We document the approach in new RFC, and we wire up some of the initial bookkeeping and APIs. We will also need https://github.com/opendata-oss/opendata/pull/486, which wires up the log prefix bloom filter, and a subsequent patch which implements the global sequence filter defined in the RFC.s


hachikuji:
I converted this to a draft while I sort through a few issues in the implementation.

PR #489, opened 2026-06-15
Logdb configurable block size

hachikuji:
We do get some performance benefit from larger blocks in some benchmarks. It is a useful knob to configure.


hachikuji:
I'll just merge this since it is a small configuration change.

PR #495, opened 2026-06-22
Logdb: Size-based segment seal

hachikuji:
This is a minor addition which I had in one of my benchmark branches. We currently only support time-based segment sealing, but this policy can lead to large differences between segment sizes, which ultimately affects both compaction and query performance. Size-base sealing is a nice alternative which gives more predictable behavior.


apurvam (approved):
LGTM, thanks!

PR #497, opened 2026-06-24
Add preview to publish workflow

hachikuji:
This patch adds a preview for releases which must be approved by the "committers" team (newly created).


hachikuji:
Here is what the preview looks like: https://github.com/opendata-oss/opendata/actions/runs/28124996185.

hachikuji:
Assuming this is not too controversial, I'll go ahead and merge.

PR #499, opened 2026-06-24
LogDb: read-only http gateway

hachikuji:
We have an HTTP service today, but it only uses the embedded reader/writer. Easy win to give it a read-only option. No cross-az charges when the writer is not in the same zone. Also did some minor api visibility fixes: `LogService` remains public, but `AppState` is now crate-visible.


apurvam (approved):
LGTM, thnaks!

PR #500, opened 2026-06-25
Upgrade to slatedb 0.14

hachikuji:
I think the only notable change is in log. Since `scan_prefix` now accepts a sub-range, we can pass through the explicit sequence range.

### DIFF
--- a/log/src/log.rs
+++ b/log/src/log.rs
@@ -567,28 +567,33 @@ impl LogDbBuilder {
     pub async fn build(self) -> Result<LogDb> {
         self.config.validate_retention()?;
         self.config.validate_compaction()?;
-        let sb = StorageBuilder::new(&self.config.storage)
-            .await
-            .map_err(|e| Error::Storage(e.to_string()))?
-            .with_semantics(
-                StorageSemantics::new()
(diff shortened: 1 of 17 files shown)


log/src/log.rs
```diff
@@ -567,28 +567,33 @@ impl LogDbBuilder {
...
+                    //
+                    // The filter policies are an orthogonal slatedb axis (see RFC
+                    // 0007): the `LogKeyPrefixExtractor` bloom hashes the logical
+                    // `(segment, user_key)` prefix into per-SST bloom filters to
+                    // prune SSTs that cannot contain the key
+                    // (`whole_key_filtering=false` — the extracted prefix
+                    // discriminates everything we probe), and the sequence-range
+                    // filter prunes SSTs whose records for the key all sit below a
+                    // scan's resume cursor. Writer, compactor, and reader must share
+                    // the same set.
```
apurvam: The writer, compactor, and reader must share the same set of what? The same set of segment extractor and filter policies, or something else. Not clear from this comment.

apurvam (approved):
LGTM, with a minor comment inline.

PR #513, opened 2026-09-02
LogDb: Fast Forking RFC

TheEnthralled:
## Summary

This is an RFC for adding a metadata-only fork feature to OpenData Log. New keys can be created on-demand by forking.  Child keys inherit the parent key entries below the creation sequence stamp with no data copies. At the same time, forked keys diverge from the parents and can be appended to like regular key streams. 

## Related Issues

N/A

## Fixes

N/A

## Test Plan

N/A

## Checklist

- [ ] Tests added/updated


log/rfcs/0009-metadata-forking.md
```diff
@@ -0,0 +1,532 @@
...
+- Fork sequence: The exclusive inheritance boundary.
+- Logical history: The applicable ancestor intervals followed by the stream’s own entries.
+- Fork family: A root and all its descendants.
+
+### Forking
+
+A fork is a log stream whose prefix of user entries (i.e., entries whose sequence numbers are less than the sequence
+number at which `/fork` was called) reference ancestor streams.
+
+Forks can use arbitrary keys just like regular logs as long as the key name is not in use by any other live stream.
```
apurvam, line 77: Not sure what you mean by this? Today, the sequence number is global and a given log can contain any number of keys. If you fork it, the new log can also have any number of keys right? Or are we using different notions of a 'log'.

apurvam (commented):
Thanks @TheEnthralled . I left a specific comment. But more generally, I'm curious if you have a specific use case in mind for forking? If so, we probably should start from there and design backward.

log/rfcs/0009-metadata-forking.md
TheEnthralled: Yes, the new log can have any number of keys. What I mean here is that if we have for example called 

```
curl -X POST http://localhost:8081/api/v1/log/append \
  -H "Content-Type: application/json" \
  -d '{
    "records": [
      {
        "key": "",
        "value": "eyJpdGVtIjoid2lkZ2V0IiwicXR5IjozLCJwcmljZSI6OS45OX0="
      }
    ]
  }'
```

and then we call 

```
curl -X POST http://localhost:8081/api/v1/log/fork \
  -H "Content-Type: application/json" \
  -d '{
       "parentKey": {"value": "cGFyZW50"},
       "childKey": {"value": "b3JkZXJz"},  
       "awaitDurable": true
  }'
```

The fork call will return a 400 as the key "b3JkZXJz" already has records written to it.

TheEnthralled:
I'm thinking this feature would be useful for parallel dev/test branches on a stream. Example: multiple engineers are each testing a different fix to the same consumer against last week's orders stream. Each needs to (a) replay the shared prefix, (b) append their own synthetic edge cases after it, and (c) write outputs without touching prod or each other.

[Kafka replay](https://cwiki.apache.org/confluence/spaces/KAFKA/pages/66852643/Kafka+Streams+Data+Re+Processing+Scenarios) with MirrorMaker can cover this use case. OpenData Log already supports part (a) through `scan` but if multiple developers want to create multiple dev branches of a stream for testing multiple solutions in parallel without copying the entire stream prefix there currently isn't straightforward  way to do so.

